PR #151, opened 2014-02-18
Copying all udev rules from inst-sys to installed system

kobliha:
- There might be several udev rules depending on the HW
- bnc#860089


kobliha:
BTW, I just found out that `70-persistent-net.rules is obsolete so its copying should be dropped` is not that black and white, bug https://bugzilla.suse.com/show_bug.cgi?id=944348 says it's actually needed.

PR #247, opened 2014-12-03
Proposal cleaning

jreidinger:
not yet ready for merge, need real testing and also automatic tests. More for discussion of design.


test/proposal_runner_test.rb
```diff
@@ -0,0 +1,37 @@
...
+      Yast.import "GetInstArgs"
+      allow(Yast::ProductControl).to receive(:GetDisabledProposals).and_return(["initial"])
+      allow(Yast::GetInstArgs).to receive(:proposal).and_return("initial")
+
+      expect(subject.run).to eq :auto
+    end
+
+    it "runs" do
+      # TODO: just catch exceptions
+      expect(subject.run).to eq :abort
```
mvidner: Huh, why should a plain `run` return `:abort`? I would expect `:next`.

What do you mean by "just catch exceptions"? Is it "the code does whatever, but does not raise an exception?".

jreidinger: yes, it is trivial test that there is no syntax or nil error. :abort because some parts are not set and it quit quickly.

ancorgs: So "runs" is obviously not a enlightening description.

jreidinger: ok, maybe not crash is better. Runner is quite tricky and not easy to test, so I just add basic test that there is no trivial error.

mvidner: Yes, it is perfectly fine to have such a test, only it should be phrased slightly differently IMHO:
```ruby
it "passes a smoke test" do
  expect { subject.run }.to_not raise_error
end
```

A "smoke test" is a term from electronics: you solder a device, power it on, and nothing happens. No smoke comes out :)

jreidinger: @mvidner good idea, I will change it this way.

jreidinger:
@mvidner your suggestion implemented

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+# this program; if not, contact Novell, Inc.
+#
+# To contact Novell about this file by physical or electronic mail, you may find
+# current contact information at www.novell.com.
+# ------------------------------------------------------------------------------
+
+require "yast"
+
+module Installation
+  # Stores various proposals clients and provides metadata about them
```
mvidner: I don't understand this description.
Reading through the class, I find it hard to understand its purpose(s) too. That is understandable, given it was pulled out of spaghetti code, but we should improve on that.
Can we at least list and describe the actual responsibilities of this class? If we manage to do that, we can even split the class so that the resulting classes have a single responsibility.

jreidinger: Well, goal of this class is Store for proposal clients. It provides metadata about them like its name, presentation order, evaluation order. It also provide actions on such clients like generate proposal, generate rich text string for proposal screen or process link from such rich text.

mvidner:
Since `ProposalStore` is an API for the properties of the proposal clients, which are defined in the control file, it should refer to the documentation. We have
- https://github.com/yast/yast-installation/blob/master/doc/control-file.md#proposals (which is somewhat outdated)
- https://github.com/yast/yast-installation-control/blob/master/control/control.rnc seems up to date

I'll go through that and check what can be used for documenting the code.

jreidinger:
@mvidner I convert control-file.md to markdown, but do not verify it too deep, so it can contain really outdated stuff, but I hope in markdown it should be easy edit it.

src/lib/installation/proposal_runner.rb
```diff
@@ -0,0 +1,806 @@
...
+      build_dialog
+      Yast::ProductControl.RetranslateWizardSteps
+      Yast::Wizard.RetranslateButtons
+      get_submod_descriptions_and_build_menu
+
+      nil
+    end
+
+    def load_matching_submodules_list
+      @locked_modules = Yast::ProductControl.getLockedProposals(
```
mvidner: It was added in commit 5423af8197f033a5b48f713b47dc4581b4b18a4b (no bug or feature mentioned). It had an entry on the old schema (DTD) but no entry in the RelaxNG schema, which is even validated at build time.
So I think the code can be dropped.

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+        Yast::Stage.stage,
+        Yast::Mode.mode,
+        @proposal_mode
+      )
+
+      @proposal_names.map!(&:first) # first element is name of client
+
+      # FIXME: is it still used?
+      # in normal mode we don't want to switch between installation and update
+      @proposal_names.delete("mode_proposal") if Yast::Mode.normal
```
mvidner: Not used in any existing skelcd-control-*.
But if we want to keep it, move it one layer up to the runner.

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+      return @tab_labels if @tab_labels
+
+      raise "Invalid call to tab_labels for proposal without tabs" unless has_tabs?
+
+
+      tabs = properties["proposal_tabs"]
+      @tab_labels = tabs.map { |m| m["label"] || "Tab" }
+    end
+
+    def proposal_names
```
mvidner: \# @return [Array\<String>] proposal names in execution order, including the "_proposal" suffix

src/lib/installation/proposal_store.rb
mvidner: This?

> 1. Provides access to metadata of proposal parts (clients), as defined in the control file elements
/productDefines/proposals/proposal:
https://github.com/yast/yast-installation-control/blob/master/control/control.rnc
2. Handles all calls to the parts (clients).

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+      Yast.import "ProductControl"
+      Yast.import "Stage"
+
+      textdomain "installation"
+
+      @proposal_mode = proposal_mode
+
+    end
+
+    def headline
```
mvidner: \# @return [String] translated headline

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+        Yast::Builtins.dgettext(
+          Yast::ProductControl.getProposalTextDomain,
+          properties["label"]
+        )
+      else
+        _("Installation Overview")
+      end
+    end
+
+    def icon
```
mvidner: \# @return [String] like "yast-foo"

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+      when "network"
+        "yast-network"
+      when "hardware"
+        "yast-controller"
+      else
+        properties["icon"] || "yast-software"
+      end
+    end
+
+    def help_text(current_tab: nil, locked_modules: false)
```
mvidner: Remove the locked_modules part.

\# @return [String] Richtext, the complete help text: a common intro + all individual parts.

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+    end
+
+    def tab_labels
+      return @tab_labels if @tab_labels
+
+      raise "Invalid call to tab_labels for proposal without tabs" unless has_tabs?
+
+
+      tabs = properties["proposal_tabs"]
+      @tab_labels = tabs.map { |m| m["label"] || "Tab" }
```
mvidner: That element is compulsory, remove ` || "Tab"`

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+      end
+
+      @can_skip
+    end
+
+    def has_tabs?
+      properties.key?("proposal_tabs")
+    end
+
+    def tab_labels
```
mvidner: \# @return [Array\<String>] translated tab labels

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+      # FIXME: is it still used?
+      # in normal mode we don't want to switch between installation and update
+      @proposal_names.delete("mode_proposal") if Yast::Mode.normal
+
+      # FIXME add filter to only installed clients
+      @proposal_names
+    end
+
+    # returns single list of modules presentation order or list of tabs with list of modules
+    def presentation_order
```
mvidner: The return value is so hard to describe that I conclude this method should not exist :)

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+      return @modules_order if @modules_order
+
+      has_tabs? ? order_with_tabs : order_without_tabs
+    end
+
+    # Makes proposal for all proposal clients.
```
mvidner: A general note: "clients" is an implementation detail. But I struggle to come up with a better name for it. "parts", "items"?

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+    def make_proposals(force_reset: false, language_changed: false, callback: Proc.new)
```
mvidner: \# @param callback Called after each client/part, to report progress. Gets part name and part result as arguments.

mvidner: \# @return [void]

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+        if proposal_map["language_changed"]
+          @descriptions = nil # invalid descriptions cache
+          return make_proposals(force_reset: force_reset, language_changed: true)
+        end
+
+        break if proposal_map["warning_level"] == :fatal
+      end
+    end
+
+    def descriptions
```
mvidner: > Calls all clients/parts to retrieve the description
@return [Hash{String => Hash}] map client/part names to hashes with keys "id", "menu_title" "rich_text_title"
http://www.rubydoc.info/github/yast/yast-yast2/Installation/ProposalClient:description

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+          missing_no += 1
+        end
+
+        @id_mapping[description["id"]] = client
+
+        res[client] = description
+      end
+    end
+
+    def id_for(client)
```
mvidner: @return [String] an id provided by the description API

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+      descriptions[client]["id"]
+    end
+
+    def title_for(client)
+      descriptions[client]["rich_text_title"] ||
+        descriptions[client]["rich_text_raw_title"] ||
+        client
+    end
+
+    def handle_link(link)
```
mvidner: > Calls `ask_user`, to change a setting interactively (if *link* is the heading for the part) or noninteractively (if it is a "shortcut")

src/lib/installation/proposal_store.rb
```diff
@@ -0,0 +1,357 @@
...
+          _(
+            "<p>\n" \
+              "To use the settings as displayed, press <b>Next</b>.\n" \
+              "</p>\n"
+          )
+        end
+      end
+    end
+
+    def order_with_tabs
```
mvidner: So this is not AS complex as I thought. Could we simply make this more consistent by saying that we always return "an array of arrays of strings that end with _proposal" ? If we have no tabs, it is a singleton array.

mvidner:
Overall, this is a partial clean-up of a big piece of code. Therefore, it is important to know what is in scope. For me, it was describing the shorter one of the new classes, `ProposalStore`, regarding its responsibilities and the interface of its public methods.

src/lib/installation/proposal_store.rb
jreidinger: I also do not know, so I do not want to invent new name that can confuse. Maybe "steps"

src/lib/installation/proposal_store.rb
jreidinger: callback is actually used mostly to get report when one client is processed so it can be reported and shown partial progress. I am not sure if yield is good idea in this case.

src/lib/installation/proposal_store.rb
jreidinger: well, this will need changes in runner. In fact, this way it is really dependent if proposal have tabs or not. I worry it won't make runner code cleaner as there is big difference in tabs and without tabs scenario.

jreidinger:
@mvidner I fixed most of issues, except ones I comment. Maybe next time regarding documentation it will be faster if you directly commit to branch

mvidner:
Thank you!
Before merging, don't forget a .changes entry. I suggest
> Refactored inst_proposal into Installation::ProposalRunner and Installation::ProposalStore.

I see you have made a nice conflict with your own Rubocop changes... hope most can be autocorrected.

mvidner:
LGTM now :)

jreidinger:
@mvidner thanks for very exhausting review :+1:

PR #252, opened 2015-01-07
Fix log copy

jreidinger:


src/lib/installation/copy_logs_finish.rb
```diff
@@ -0,0 +1,107 @@
...
+# To contact Novell about this file by physical or electronic mail, you may find
+# current contact information at www.novell.com.
+# ------------------------------------------------------------------------------
+
+require "yast"
+
+require "installation/finish_client"
+
+module Installation
+  class CopyLogsFinish < ::Installation::FinishClient
```
ancorgs: Just asking. Do we have any convention on naming all these clients living under ```lib/installation```? Like always using ```WhateverFinish``` for clients inheriting from ```FinishClient```?

src/lib/installation/copy_logs_finish.rb
```diff
@@ -0,0 +1,107 @@
...
+
+          target_basename = "y2log-#{target_no}"
+          copy_log_to_target(file, target_basename)
+
+          target_path = ::File.join(
+            Yast::Installation.destdir,
+            Yast::Directory.logdir,
+            target_basename
+          )
+          compress_cmd = "gzip -f #{target_path}"
```
ancorgs: Maybe a comment here to justify the ```-f``` (with a reference to the bug report)?

src/lib/installation/copy_logs_finish.rb
jreidinger: actually no convention...in bootloader it is simple `Bootloader/FinishClient` but here it can create collisions so I name it this way. `CopyLogs` without finish can confuse reader as it can be seen as common class to copy logs from one place to another.

src/lib/installation/copy_logs_finish.rb
jreidinger: good idea :+1:

test/copy_logs_finish_test.rb
```diff
@@ -0,0 +1,53 @@
...
+require "installation/copy_logs_finish"
+
+describe ::Installation::CopyLogsFinish do
+  describe "#write" do
+    before do
+      allow(Yast::WFM).to receive(:Execute)
+    end
+
+    it "copies logs from instalation to target system" do
+      allow(Yast::WFM).to receive(:Read).and_return(["y2start.log"])
```
ancorgs: Without reading the implementation code, the purpose of this ```allow``` (returning a list of log files) is not obvious (no ```with``` part). Moreover, it's repeated in every ```it```.

What about using ```let``` or some specific helper method to label it?

test/copy_logs_finish_test.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      expect(Yast::WFM).to receive(:Execute).with(anything(), /cp/).at_least(:once)
```
ancorgs: The y2start.log returned by the previous call is not used in the expectation. Feels strange.

test/copy_logs_finish_test.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      subject.write
+    end
+
+    it "rotate y2log" do
```
ancorgs: rotates :-)

test/copy_logs_finish_test.rb
jreidinger: I am also considering it and if you get same idea I will do it.

test/copy_logs_finish_test.rb
jreidinger: fair comment. I will add it.

test/copy_logs_finish_test.rb
```diff
@@ -0,0 +1,53 @@
...
+      allow(Yast::WFM).to receive(:Read).and_return(["y2log-1.gz"])
+
+      expect(Yast::WFM).to receive(:Execute).with(anything(), /cp .*y2log-1.gz .*y2log-2.gz/)
+
+      subject.write
+    end
+
+    it "compress y2log if not already done" do
```
ancorgs: "compresses"... and "does" (in the next one) and "rotates" (last one).

test/copy_logs_finish_test.rb
```diff
@@ -0,0 +1,53 @@
...
+      allow(Yast::WFM).to receive(:Read).and_return(["y2log-1"])
+
+      expect(Yast::WFM).to receive(:Execute).with(anything(), /gzip .*y2log-2/) #-2 due to rotation
```
ancorgs: ```expect(Yast::WFM).to receive(:Execute).with(anything(), ...)```  is also quite long and repeated several times, so it maybe deserves a helper method.

jreidinger: yep, maybe time to finally find time to have generic rspec helper for SCR in yast2 or ruby bindings

ancorgs:
LGTM.

BTW, if you start working on the SCR helpers, as suggested by you comment, I'm not unhappy with my last attempt. https://github.com/ancorgs/yast-security/blob/pre907907_cleanup/test/SCRStub.rb

jreidinger:
@ancorgs yep, I plan to reuse your stub, but to be generic enough, there is some problems like .target.bash and .target.bash_output and also .local.bash for WFM and not SCR. I worry it won't be extra nice and probably some parts will be repetitive

ancorgs:
That's why I removed things like ```expect_to_execute``` in the last incarnation and just implemented a ```exec_bash``` matcher for ".target.bash". My idea was to have a generic ```exec``` matcher that checks again every possible form and them some specific matchers.

jreidinger:
@ancorgs unfortunatelly this won't work as 
```
def exec_bash(command)
receive(:Execute).with(path(".target.bash"), command)
end
```
work only for SCR, because WFM use `.local.bash` as System agent is attached to path `.local` and not `.target`. It is quite stupid and hard to handle. It also do not solve issue with `.target.bash_output`

ancorgs:
@jreidinger ahm. I see.

PR #253, opened 2015-01-12
Rubocop

jreidinger:


src/clients/inst_disk_for_image.rb
```diff
@@ -90,8 +89,8 @@ def disk_for_image_dialog
     end
 
     def disk_for_image_help_text
-      _("Select the disk, which the image will be deployed to. " +
-        "All data on the disk will be lost and the disk will be " +
+      _("Select the disk, which the image will be deployed to. " \
+        "All data on the disk will be lost and the disk will be " \
```
lslezak: NP: according to our style guide [no method should be called on multi-line block result](https://github.com/SUSE/style-guides/blob/master/Ruby.md#blocks).

src/clients/inst_disk_for_image.rb
```diff
@@ -62,20 +62,19 @@ def main
...
       Builtins.y2milestone("TM: %1", target_map)
       # FIXME: move blacklist to Storage
-      used_by_blacklist = [ :CT_DMRAID, :CT_DMMULTIPATH, :CT_MDPART ]
-      target_map.select { | key, value |
-        Storage.IsDiskType(value["type"]) && (! used_by_blacklist.include? value["used_by"])
-      }.keys
+      used_by_blacklist = [:CT_DMRAID, :CT_DMMULTIPATH, :CT_MDPART]
+      target_map.select do | _key, value |
+        Storage.IsDiskType(value["type"]) && (!used_by_blacklist.include? value["used_by"])
+      end.keys
```
lslezak: NP: according to our style guide [no method should be called on multi-line block result](https://github.com/SUSE/style-guides/blob/master/Ruby.md#blocks).

src/clients/inst_download_release_notes.rb
```diff
@@ -63,9 +62,7 @@ def download_release_notes
 
         if Ops.get_boolean(proxy_ret, ["HTTP", "tested"], true) == true &&
             Ops.get_integer(proxy_ret, ["HTTP", "exit"], 1) == 0
-          user_pass = Proxy.user != "" ?
-            Ops.add(Ops.add(Proxy.user, ":"), Proxy.pass) :
-            ""
+          user_pass = Proxy.user != "" ? "#{Proxy.user}:#{Proxy.pass}" : ""
           proxy = "--proxy #{Proxy.http}"
           proxy << " --proxy-user '#{user_pass}'" unless user_pass.empty?
```
lslezak: (Not changed, but maybe here should be used `String.Quote()` or `Shellwords.escape()`...)

mvidner:
The target branch is `master`, but we should target `SLE-12-GA` otherwise merging later will be very painful.

jreidinger:
@mvidner well, it is not without changes to code, so I am not sure if it is good idea to merge to SLE-12-GA. What is benefit? I still hope that SP1 will be based on master

src/clients/inst_upgrade_urls.rb
```diff
@@ -402,12 +402,12 @@ def FillUpURLs
...
 
       # If some new (since 10.3 Alpha?) URLs found, use only them
-      if @system_urls != nil && @system_urls != []
+      if !@system_urls.nil? && @system_urls != []
         Builtins.foreach(@system_urls) do |one_url_map|
           # If the user has gone back and forward, some of the repositories of
           # the previous system could be already registered for this
           # installation
-          if repo = equivalent_repo_for(one_url_map)
+          if repo == equivalent_repo_for(one_url_map)
```
lslezak: Um, maybe this assignment was intentional, could you double check?

ancorgs: Yes, it was.

src/clients/inst_disk_for_image.rb
jreidinger: thanks, fixed

src/clients/inst_download_release_notes.rb
jreidinger: well, then for more complex passwords it won't be backward compatible, but I am not sure how big problem it is as old behavior looks like buggy one.

src/clients/inst_upgrade_urls.rb
jreidinger: really? I double check it before change and for me it looks like easy check if repo is already created, at least comment above check lead me to such idea.

lslezak: Anyway, I'd avoid assignment in condition, it is confusing...

lslezak:
Please, add `rubocop` call to `.travis.yml` so we keep the good style in the future :wink: 
(BTW lock the rubocop version, it might change in the future and break Travis builds suddenly...)

lslezak:
I agree with @mvidner, we should target to `SLE-12-GA`, to avoid possible conflicts.

Yes, there are some code changes, but 99% of the changes are in white space or style fixes without any side effect. The rest has been reviewed manually :wink: ...

lslezak:
LGTM

Thanks, nice improvement! :+1:

PR #257, opened 2015-01-23
release the last two changes

mvidner:
fixing a failure in openqa, see https://github.com/yast/yast-yast2/pull/325#issuecomment-71190413


lslezak:
LGTM

PR #260, opened 2015-01-27
ensure the langugages value is set and Solver is run in

jsuchome:
case of live installation (bnc#904103)


jsuchome:
Mode.live_installation might be strange here, but during normal installation it works, so I assume there's a different order of Pkg initialization and solving during live install

lslezak:
Please, add a comment there with summary of your previous comment, I think that it should be included in the code.

lslezak:
LGTM, thanks for the fix!

PR #263, opened 2015-01-28
updated SLES SP4 control files (bnc#914288)

lslezak:
- display the generic network settings also on PPC and S390
  (patch by Dinar)

- 2.17.110

See https://bugzilla.suse.com/show_bug.cgi?id=914288#c7


jreidinger:
LGTM

PR #265, opened 2015-02-02
Plymouth bug 903682

schubi2:
AutoYaST: If the system starts in multi-user mode plymouth will
be quit while installation in order to ensure that installation
will be finished on console 1 and the login prompt will be
shown.


aschnell:
A bit hackish but the bugs reports seem to indicate that a clean solution does not exist ATM.

LGTM

PR #266, opened 2015-02-03
Delete or copy install.inf as applicable (bnc#897066)

cwh42:


src/include/installation/misc.rb
```diff
@@ -365,5 +365,37 @@ def WriteSecondStageRequired(scst_required)
 
       nil
     end
+
+    def SecondStageRequired?
```
lslezak: For new code we should use the usual Ruby naming conventions, i.e. use `second_stage_required?` here.

src/include/installation/misc.rb
```diff
@@ -365,5 +365,37 @@ def WriteSecondStageRequired(scst_required)
...
+      if Stage.initial
+        # the current one is 'initial'
+        run_second_stage = true
```
lslezak: You can remove this assignment, `run_second_stage` is set in both `if` branches.

lslezak:
NTH: it would be nice to add a testcase (testing `copy_files_finish.rb` client would be difficult, but a test for the new `second_stage_required?` method would be great).

src/include/installation/misc.rb
lslezak: BTW use a guard clause here:

```ruby
return false unless Stage.initial
.... <the rest goes here>
```

mvidner:
(Travis failed beacuse of Rubocop checks, see line 656 of the [Travis log](https://travis-ci.org/yast/yast-installation/builds/50935264))

test/installation_misc_test.rb
```diff
@@ -0,0 +1,99 @@
...
+def stub_secondstage(bool)
+  allow(Yast::AutoinstConfig).to receive(:second_stage).and_return(bool)
+end
+
+describe Yast::InstallationMiscInclude do
+  subject { InstallationMiscIncludeTest.new }
+  describe "#second_stage_required?" do
+    before { allow(Yast::ProductControl).to receive(:RunRequired).and_return(true) }
+
+    it "when not in initial stage" do
```
lslezak: The `it` description should describe the behavior and it should look like an English sentence, e.g.
```ruby
it "returns false when not running in the first stage" do
```

test/installation_misc_test.rb
```diff
@@ -0,0 +1,99 @@
...
+      stub_initialstage(false)
+      expect(subject.second_stage_required?).to be_false
+    end
+
+    context "when in AutoinstConfig.second_stage" do
```
lslezak: Prefer human readable texts in the descriptions, if you run `rspec --format doc` you should be able to simply read the output as an English text.

RSpec is BDD (behavior driven development) so the RSpec output should describe the behavior as a readable document.

lslezak: I think that the grouping by `AutoinstConfig.second_stage` is wrong, it makes sense only in `autoinst` or `autoupgrade` mode, it does not make sense in other modes. It does not split the tests into logically separated groups as the "other" modes case is actually the same in both contexts.

In this case you can leave it without contexts (the number of tests is still OK for a single group) or you could divide them into these groups (contexts):

- `installation` mode
- `upgrade` mode
- `autoinst` mode
- `autoupgrade` mode

Testing the other modes does not make sense, the `installation/inst_inc_first.rb`, as the name suggests, is used only in the first stage.

test/installation_misc_test.rb
```diff
@@ -0,0 +1,99 @@
...
+end
+
+describe Yast::InstallationMiscInclude do
+  subject { InstallationMiscIncludeTest.new }
+  describe "#second_stage_required?" do
+    before { allow(Yast::ProductControl).to receive(:RunRequired).and_return(true) }
+
+    it "when not in initial stage" do
+      stub_initialstage(false)
+      expect(subject.second_stage_required?).to be_false
```
jreidinger: `be_false` is rspec2 specific. Please use `.to eq false` instead. it works in rspec 2 and also rspec 3

test/installation_misc_test.rb
```diff
@@ -0,0 +1,99 @@
...
+      before do
+        stub_secondstage(true)
+        stub_initialstage(true)
+      end
+
+      it "when in autoinst mode" do
+        stub_autoinst(true)
+        stub_autoupgrade(false)
+
+        expect(subject.second_stage_required?).to be_true
```
jreidinger: same here with `be_true`

cwh42:
Next try :)

jreidinger:
for me it looks good

PR #267, opened 2015-02-04
Remote display - allow setting custom display number (bnc#913888)

lslezak:
- 2.17.112

Notes:

- The regexp is little bit tricky, the problem is that IPv6 contains `:` delimiter which is also used in remote display specification. To make the check easier IPv6 address needs to contain full display specification including the screen number.

- I'll adapt https://en.opensuse.org/SDB:Linuxrc#p_displayip documentation accordingly


wfeldt:
lokks ok

PR #268, opened 2015-02-05
Support custom display number in "display_ip" boot option (bnc#913888)

lslezak:
- 3.1.130

Port #267 to `master`


jreidinger:
LGTM

PR #270, opened 2015-02-11
Removed redundant initialization label (bnc#878538)

lslezak:
- 3.1.132


lslezak:
@cwh42 Actually the dialog already contains "Initializing..." header, this just a duplicated label in the working area on the right side. See [this screenshot](https://bugzilla.suse.com/attachment.cgi?id=590746).

PR #272, opened 2015-02-17
- get list of mounts from /proc/mounts in umount_finish (for fate#318392)

aschnell:


src/clients/umount_finish.rb
```diff
@@ -98,16 +91,19 @@ def main
...
-          if mountpoint != "swap"
-            @umountList = Builtins.add(@umountList, mountpoint)
+        umount_list = []
+        SCR.Read(path(".proc.mounts")).each do |entry|
+          mountpoint = entry["file"]
+          if mountpoint.start_with?(Installation.destdir)
+            umount_list << mountpoint[Installation.destdir.length, mountpoint.length]
           end
         end
+        umount_list = umount_list.sort
```
ancorgs: NP: ```umount_list.sort!```

src/clients/umount_finish.rb
```diff
@@ -82,13 +82,6 @@ def main
...
         Pkg.TargetFinish
 
-        # loop over all filesystems
-        @mountPoints = Convert.convert(
-          Storage.GetMountPoints,
-          from: "map",
-          to:   "map <string, list>"
-        )
-        Builtins.y2milestone("Known mount points: %1", @mountPoints)
         Builtins.y2milestone(
```
ancorgs: NP: these calls to Builtins are just 2 lines away from the code you are rewriting. Nice chance to remove them :wink:

src/clients/umount_finish.rb
```diff
@@ -187,33 +183,33 @@ def main
...
         # file is not on the root fs
-        Builtins.y2milestone("umount list %1", @umountList)
+        Builtins.y2milestone("umount list %1", umount_list)
         Builtins.foreach(
           Ops.get_list(@targetMap, ["/dev/loop", "partitions"], [])
         ) do |e|
           if Ops.greater_than(Builtins.size(Ops.get_string(e, "mount", "")), 0)
             Storage.Umount(Ops.get_string(e, "device", ""), true)
-            @umountList = Builtins.filter(@umountList) do |m|
+            umount_list = Builtins.filter(umount_list) do |m|
```
ancorgs: What about changing this ```Builtins.filter``` for a
```
umount_list.select! {|m| m != e["mount"] }
```
And the ```y2milestones``` that you are already changing by ```log.info```

ancorgs: Or even more expressive using ```reject!``` and a positive comparison.

src/clients/umount_finish.rb
```diff
@@ -187,33 +183,33 @@ def main
...
             )
           end
         end
 
-        # *** umountList is lexically ordered !
+        # *** umount_list is lexically ordered !
         # now umount in reverse order (guarantees "/" as last umount)
 
-        @umountLength = Builtins.size(@umountList)
+        @umountLength = Builtins.size(umount_list)
```
ancorgs: umount_list.size

ancorgs:
@aschnell I know you are not aiming for a code cleanup or nothing similar, but I think we should rubify at least the lines we touch.

src/clients/umount_finish.rb
```diff
@@ -98,16 +91,19 @@ def main
...
+        # e.g. snapper, mounted filesystems in the target
 
-        Builtins.foreach(@mountPoints) do |mountpoint, _mountval|
-          if mountpoint != "swap"
-            @umountList = Builtins.add(@umountList, mountpoint)
+        umount_list = []
+        SCR.Read(path(".proc.mounts")).each do |entry|
+          mountpoint = entry["file"]
+          if mountpoint.start_with?(Installation.destdir)
+            umount_list << mountpoint[Installation.destdir.length, mountpoint.length]
```
ancorgs: NP: ```mountpoint[Installation.destdir.length, -1]```

aschnell:
The problem with refactoring the things you touch anyway often results in an explosion. See, I already changed the name from @umountList to umount_list and that caused several more modification. I don't want to go further now.

src/clients/umount_finish.rb
aschnell: I saw the -1 way in the documentation but don't like it.

src/clients/umount_finish.rb
```diff
@@ -187,33 +183,33 @@ def main
...
               "loop umount %1 new list %2",
               Ops.get_string(e, "mount", ""),
-              @umountList
+              umount_list
             )
           end
         end
 
-        # *** umountList is lexically ordered !
+        # *** umount_list is lexically ordered !
```
ancorgs: It looks like they are actually lexically ordered in my /proc/mounts, but since you are the expert: Is still true despite the change in the method for getting the list? Just making sure that you read the comment and didn't just changed the name of the variable as part of a search&replace :-)

aschnell: The list is alphabetical sorted and AFAIS that's identical to lexically ordered.

ancorgs:
I still think that when we are targeting master we should be more "aggressive" with the code cleanup than when targeting maintenance branches. Otherwise, we will never get rid of the cruft. Even though, I can live with the current shape, so as soon as "mergeability" is fixed it LGETM (looks good enough to me). :-)

PR #273, opened 2015-03-16
Use a consistent look for openqa tests even via SSH

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=920635


startup/YaST2.call
```diff
@@ -341,7 +341,12 @@ function prepare_for_ssh () {
 # prepare SSH installation
 # ---
 #
-	:
+        # bsc#920635 consistent look for openqa tests even via ssh
+        if [ "$Y2FULLSCREEN" = 1 ]; then
+                log "\tFull screen mode requested"
+                Y2_UI_ARGS="--fullscreen --noborder --auto-fonts"
+        fi
+        set_inst_qt_env
```
jreidinger: why this change is needed? why not call it only in if statement?

jreidinger: respective it is missing in changes, that it backport enablement of yast theme  in ssh

mvidner: Right, I forgot to mention it. How about this:
```
- Use a consistent look for openqa tests even via SSH (bsc#920635)
  - Enable the installation theme.
  - Let Y2FULLSCREEN=1 enable the full screen mode.
```

jreidinger: yep, looks good. LGTM if you mention it

PR #276, opened 2015-04-01
Fix runner

jreidinger:
fix crashing proposal_runner in offline update ( happen only in if condition for normal stage and update mode ). So it affects only new yast2-migration module.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Apr  1 18:54:34 UTC 2015 - jreidinger@suse.com
+
+- fix method missing error in proposal_runner
+- 3.1.137
+
+-------------------------------------------------------------------
 Wed Apr  1 11:39:32 UTC 2015 - jreidinger@suse.com
 
 - fix dependencies in proposal_store
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,19 +1,20 @@
 -------------------------------------------------------------------
 Wed Apr  1 18:54:34 UTC 2015 - jreidinger@suse.com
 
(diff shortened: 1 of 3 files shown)


package/yast2-installation.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Apr  1 18:54:34 UTC 2015 - jreidinger@suse.com
+
+- fix method missing error in proposal_runner
```
kobliha: PLS add FATE number, it actually IS for wagon, right?

kobliha:
It seems that you will have to release an online update for SLE12GA, which means you should have merged this to SLE12GA first and then to master.

package/yast2-installation.changes
jreidinger: yes, but it is error in code and I think it do not provide additional info in fate entry, right?

jreidinger:
well, I worry that there more changes from master needed for migration module, so we probably need to release master in SLE12GA

jreidinger:
question is how big problem it is as yast2-installation is used only in inst-sys now

package/yast2-installation.changes
kobliha: But it's an explanation for maintenance team why we need to release it - because of a feature, otherwise we would not find this error. Look at it from the different point of view: maintenance update without bugreport? And how will QA test it? For that reason, I believe FATE fits better.

jreidinger: OK, make sense

kobliha:
LGTM

Side note: We have to solve these tests for private methods in general

jreidinger:
LGTM

side note: I think testing it as part of public method is OK, we just need to find good way how to test UI.

PR #279, opened 2015-04-15
[Review] Request from 'schubi2' @ 'yast/yast-installation/review_150415_fixed_proposal_tab'

schubi2:
  * 19782af packaging
  * 5afb46c fixed different bugs in proposal with multiple tabs


package/YaST2-Second-Stage.service
```diff
@@ -18,7 +18,7 @@ Environment=SYSTEMCTL_OPTIONS=--ignore-dependencies TERM=linux PX_MODULE_PATH=""
 # If the system starts in multi-user mode plymouth will be quit while
 # installation in order to ensure that installation will be finished on
 # console 1 and the login prompt. (bnc#903682,889757,897956)
-ExecStartPre=/bin/sh -c 'if [ `/usr/bin/systemctl get-default` = "multi-user.target" ];then /usr/bin/plymouth quit; else /usr/bin/plymouth deactivate --hide-splash; fi'
+ExecStartPre=/bin/sh -c 'if [ -x /usr/bin/plymouth ] ; then if [ `/usr/bin/systemctl get-default` = "multi-user.target" ];then /usr/bin/plymouth quit; else /usr/bin/plymouth deactivate --hide-splash; fi; fi'
```
lslezak: I'm not a shell expert, but I think the conditions should be merged via AND (-a test option) to avoid nested conditions.

package/yast2-installation.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Apr 15 13:11:40 CEST 2015 - schubi@suse.de
+
+- Fixed differnt bugs in proposal overview with multiple tabs.
```
lslezak: Any bug numbers?

src/lib/installation/proposal_runner.rb
```diff
@@ -683,19 +683,23 @@ def build_dialog
       )
 
       if @store.tabs?
+        t = -1
         tab_labels = @store.tab_labels
         if Yast::UI.HasSpecialWidget(:DumbTab)
           panes = tab_labels.map do |label|
+            t += 1
```
lslezak: I'd prefer using "map with index" here (http://stackoverflow.com/questions/4697557/how-to-map-with-index-in-ruby), that's better than a local counter explicitly increased. Moreover initializing it to -1 looks weird...

(In general, using a local index iterator can be usually avoided in Ruby.)

package/yast2-installation.changes
schubi2: There is no one. I have found it in master while trying out integration tests with current master

package/YaST2-Second-Stage.service
schubi2: Good point.

schubi2: Hm, how it should look like, because there is an "else" between ? 
-X means that the file is executeable too.

src/lib/installation/proposal_store.rb
```diff
@@ -348,13 +348,18 @@ def order_without_tabs
...
       modules_order = presentation_order
-      modules_order = modules_order[current_tab] if tabs?
+      if tabs? && current_tab
+        modules_order = modules_order[current_tab]
 
-      modules_order.each_with_object("") do |client, text|
-        next unless descriptions[client]["help"]
-        next if descriptions[client]["help"].empty?
+        modules_order.each_with_object("") do |client, text|
+          next unless descriptions[client]
```
lslezak: Um, the guard block does not look nice to me esp. with mixed `if` and `unless` it's rather complicated.

What about 
```ruby
if descriptions[client] && !descriptions[client]["help"].to_s.empty?
  text << descriptions[client]["help"]
end
```

instead? (Take advantage of `nil.to_s => ""`...)

package/YaST2-Second-Stage.service
lslezak: Ouch, I overlooked that `else`... I hate single line bash scripts... :grimacing: 

OK, seems that it cannot be improved.

src/lib/installation/proposal_runner.rb
schubi2: Yes, that's really nice. I have changed.

src/lib/installation/proposal_store.rb
schubi2: Yes, looks much more better --> changed

lslezak:
Great, your `map.with_index` looks even better than my proposed solution!

LGTM!

PR #280, opened 2015-04-15
Updated the README file with some more info according to our template

kobliha:
- also includes a little cleanup of "temporary" solution ~ already 7 years old and not in use anymore


lslezak:
LGTM

PR #281, opened 2015-04-16
Always enable startup scripts for Second Stage and Firstboot

kobliha:
- The approach with %service_add_post works only for new installations but doesn't work for upgrades
- bsc#924278

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Apr 16 13:02:23 CEST 2015 - locilka@suse.com
+
+- Always enable systemd startup services for Second Stage and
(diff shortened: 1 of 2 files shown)


jreidinger:
Can we somehow detect that user want to have it disabled? Otherwise it will be quite annoying if customer try to reduce reboot time by disabled not used services and we will enable it again.

kobliha:
1. I don't know how
2. The script checks for existence of a file and exits if file doesn't exist - cheap
3. We already ENABLE the script, always, even after online update, this only handles upgrade from SLE 11 where scripts had different name (and were not handled by systemd)

jreidinger:
ok, so LGTM maybe add it as comment why it is fine to have it always enabled.

kobliha:
OK, comment could help

PR #285, opened 2015-05-20
Set Xvnc server resolution to 96 dpi (bsc#919456)

lslezak:
...to fix broken layout in VNC installations (defaults to 75 dpi) ([bsc#919456](https://bugzilla.suse.com/show_bug.cgi?id=919456)).

The installation stylesheet uses units like `em` or `pt` [*] which
depend on the display resolution (dpi).

- 2.17.114

[*] See http://www.w3.org/Style/Examples/007/units.en.html for more details about the topic.


mvidner:
Good catch!
LGTM.

PR #286, opened 2015-05-21
Set Xvnc server resolution to 96 dpi (bsc#919456)

lslezak:
...to fix broken layout in VNC installations (defaults to 75 dpi).

The installation stylesheet uses units like 'em' or 'pt' [*] which
depend on the display resolution (dpi).

- 3.1.142

[*] see http://www.w3.org/Style/Examples/007/units.en.html


mvidner:
LGTM

PR #292, opened 2015-06-17
Implemented "triggers" for calling installation proposals again when needed

kobliha:
## Use Case
In some cases, proposals depend on each other. For instance, kdump sets required packages for software proposal, but software proposal influences space available on disk and kdump needs to evaluate this space whether dump fits there (FATE#317488). There have been several other cases in the past. That's why I've created a generic handling using triggers.

## Implementation details
Each proposal client can define so called `trigger` with keys `expect` and `value`. All these triggers are evaluated at the end of the proposal and if output of the `expect` does not match `value`, proposal will be called again. All such proposals are called after all triggers are evaluated and then triggers are evaluated again. There is a counter that breaks from an infinite loop if needed.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Wed Jun 17 09:29:09 CEST 2015 - locilka@suse.com
+
+- Implemented 'trigger's for installation proposal (FATE #317488).
+  Any *_proposal client can define 'trigger' in 'MakeProposal'
+  that defines in which circumstances it should be called again
+  after all proposals have been called, e.g., if partitioning or
+  software selection changes.
+- 3.1.145
+
+-------------------------------------------------------------------
 Tue Jun  2 08:41:03 UTC 2015 - jreidinger@suse.com
 
 - fix crash in Upgrade when creating post upgrade snapshot
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Wed Jun 17 09:29:09 CEST 2015 - locilka@suse.com
 
-- Implemented 'trigger's for installation proposal (FATE #317488).
+- Implemented triggers for installation proposal (FATE #317488).
   Any *_proposal client can define 'trigger' in 'MakeProposal'
   that defines in which circumstances it should be called again
   after all proposals have been called, e.g., if partitioning or
--- a/src/lib/installation/proposal_store.rb
+++ b/src/lib/installation/proposal_store.rb
@@ -131,8 +131,13 @@ module Installation
 
       @proposal_names.map!(&:first) # first element is name of client
 
-      # FIXME: add filter to only installed clients
-      @proposal_names
+      missing_proposals = @proposal_names.reject { |proposal| Yast::WFM::ClientExists(proposal) }
+      unless missing_proposals.empty?
+        log.warn "These proposals are missing on system: #{missing_proposals}"
+      end
+
+      # Filter missing proposals out
+      @proposal_names -= missing_proposals
     end
 
     # returns single list of modules presentation order or list of tabs with list of modules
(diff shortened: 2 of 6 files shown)


package/yast2-installation.changes
```diff
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Wed Jun 17 09:29:09 CEST 2015 - locilka@suse.com
+
+- Implemented 'trigger's for installation proposal (FATE #317488).
```
imobach: I guess `'triggers'` instead of `'trigger's`

kobliha: That's for purpose: one `trigger` for each proposal :)

src/lib/installation/proposal_store.rb
```diff
@@ -145,73 +155,100 @@ def presentation_order
...
-      @link2submod = {}
+    def make_proposals(force_reset: false, language_changed: false, callback: Proc.new {})
+      clear_triggers
+      clear_proposals_counter
 
-      proposal_names.each do |submod|
-        proposal_map = make_proposal(submod, force_reset:      force_reset,
-                                             language_changed: language_changed)
+      # At first run, all clients will be called
+      call_proposals = Yast.deep_copy(proposal_names)
```
imobach: Is `Yast.deep_copy` really needed?

src/lib/installation/proposal_store.rb
```diff
@@ -145,73 +155,100 @@ def presentation_order
     # Makes proposal for all proposal clients.
     # @param callback Called after each client/part, to report progress. Gets
     #   part name and part result as arguments
-    def make_proposals(force_reset: false, language_changed: false, callback: Proc.new)
-      @link2submod = {}
+    def make_proposals(force_reset: false, language_changed: false, callback: Proc.new {})
```
imobach: I'll try to split this method a little bit because it seems too long for me. Maybe you could just do initializations and let that the body of the `while` iteration  is performed by another method (something like `process_proposal` or something like that). I think it make the code to look cleaner.

src/lib/installation/proposal_store.rb
```diff
@@ -145,73 +155,100 @@ def presentation_order
...
+    # @see http://www.rubydoc.info/github/yast/yast-yast2/Installation/ProposalClient:description
+    def description_for(client)
+      @descriptions ||= {}
+      return @descriptions[client] if @descriptions.has_key?(client)
 
-        @id_mapping[description["id"]] = client
+      description = Yast::WFM.CallFunction(client, ["Description", {}])
 
-        res[client] = description
+      if !description.has_key?("id")
```
imobach: Maybe I'd use `unless` to be consistent with other uses (line 193, for example). If you want to use `if`, I guess it's no problem in doing `description["id"]`.

src/lib/installation/proposal_store.rb
```diff
@@ -145,73 +155,100 @@ def presentation_order
...
+
+      @descriptions[client] = description
+    end
+
+    # Returns all currently cached client descriptions
+    #
+    # @return [Hash] with descriptions
+    def descriptions
+      @descriptions ||= {}
+      @descriptions
```
imobach: This line is not needed (as the previous one returns `@descriptions`).

src/lib/installation/proposal_store.rb
```diff
@@ -221,8 +258,94 @@ def handle_link(link)
...
+    # returns nil if link is unknown.
+    # Link can be either the client ID or a shortcut link.
+    #
+    # @param [String] link
+    # @return [String] client name
+    def client_for_link(link)
+      raise "There are no client descriptions known, call 'client(Description)' first" if @descriptions.nil?
+
+      matching_client = @descriptions.find do |client, description|
+        description.fetch("id", nil) == link ||
```
imobach: `description.fetch("id", nil)` is basically the same that `description["id"]`.

src/lib/installation/proposal_store.rb
```diff
@@ -221,8 +258,94 @@ def handle_link(link)
...
+      matching_client.first
+    end
+
   private
 
+    def clear_proposals_counter
+      @proposals_run_counter = {}
+    end
+
+    def should_run_proposals_again?(proposals)
```
imobach: This method, as a side effect, updates the `@proposal_run_counter`. You can split this part in an independent method or, if you're not willing to do that, please add a note about that.

src/lib/installation/proposal_store.rb
```diff
@@ -221,8 +258,94 @@ def handle_link(link)
...
+    # Returns whether given client should be called again during 'this'
+    # proposal run according to triggers
+    #
+    # @param [String] client name
+    # @return [Boolean] whether it should be called
+    def should_be_called_again?(client)
+      @triggers ||= {}
+      return false unless @triggers.has_key?(client)
+
+      unless @triggers[client].has_key?("expect") &&
```
imobach: np: I'd probably take that check out of here (to an auxiliary method). Not so important anyway as this method is quite readable.

package/yast2-installation.changes
imobach: Ok. I failed to 'parse' it :)

src/lib/installation/proposal_store.rb
```diff
@@ -352,9 +478,8 @@ def modules_help(current_tab)
         modules_order = modules_order[current_tab]
 
         modules_order.each_with_object("") do |client, text|
-          if descriptions[client] && !descriptions[client]["help"].to_s.empty?
-            text << descriptions[client]["help"]
-          end
+          description = description_for(client)
+          text << description["help"] unless !description["help"].to_s.empty?
```
imobach: Are you sure do you want to use an `unless`? Anyway, it looks confusing to me.

imobach:
In general, it looks nice. But replacing evals for fun_ref, as proposed by @jreidinger, will be great.

src/lib/installation/proposal_store.rb
kobliha: Hmm, probably leftover from previous implementation

src/lib/installation/proposal_store.rb
kobliha: Me too :)

src/lib/installation/proposal_store.rb
kobliha: I actually think that splitting into submethod will make it harder to understand as the code will be not visible anymore (it's gonna be a private method), but that's life. I don't have strong opinion on that.

imobach:
LGTM

kobliha:
Thanks @imobach 
I will change the implementation to use `fun_ref` first, and then ask for reLGTM again :)
Additionally, RuboCop still doesn't like begin/rescue/while but I want it exactly that way, so I have to adapt RC rules first before merging.

imobach:
Ok, no problem :) I thought that you'll be implementing the fun_ref in another PR.

src/lib/installation/proposal_store.rb
```diff
@@ -145,73 +155,85 @@ def presentation_order
...
-          return make_proposals(force_reset: force_reset, language_changed: true)
-        end
+        # Second and next runs: only triggered clients will be called
+        call_proposals = proposal_names.select { |client| should_be_called_again?(client) }
 
-        break if proposal_map["warning_level"] == :fatal
-      end
+        unless call_proposals.empty?
+          log.info "These proposals want to be called again: #{call_proposals}"
+          raise StopIteration, "too_many_loops" unless should_run_proposals_again?(call_proposals)
```
jreidinger: I do not get why you overuse exception here. For me it can loop like as whole:

loop do
  call_proposals.each ...

  # we are done, no proposals need to be redone
  break if call_proposal.empty?

  log.info "These proposals want to be called again: #{call_proposals}"
  
  if !should_run_proposals_again?(call_proposals)
    log.info "Iteration over proposals stopped: too many loops"
    break
  end
end

kobliha: Because there are more possible exit points...

imobach:
LGTM

kobliha:
I'll manually retest and merge when all is OK, thanks @imobach

mvidner:
A manual integration test revealed a bug, https://bugzilla.opensuse.org/show_bug.cgi?id=935385

imobach:
LGTM

kobliha:
@imobach thanks :)

PR #294, opened 2015-06-29
Hide export

jreidinger:


src/lib/installation/proposal_runner.rb
```diff
@@ -74,6 +74,9 @@ def run
         return :auto
       end
 
+      args = (Yast::WFM.Args || []).first || {}
+      @hide_export = args["hide_export"]
```
lslezak: NP: I'd probably named it `hide_export_button` to be clear what the export means...

lslezak:
But still LGTM

src/lib/installation/proposal_runner.rb
jreidinger: well, I mean it more generic like "hide export autoyast profile ability" it is not just about button ( even in fact it is only not adding button ).

lslezak: OK, then full LGTM :wink:

PR #296, opened 2015-07-01
Fixed handling user request to change a proposal

kobliha:
- Handling values returned from Description and MakeProposal were messed up, especially in test-cases
- bsc#936448


jreidinger:
LGTM

PR #297, opened 2015-07-07
Huha snapshot cleanup

shundhammer:
https://trello.com/c/YDDWzV5X


jreidinger:
except similar stuff with update/upgrade LGTM

PR #298, opened 2015-07-20
Enabled coverage reporting to Coveralls

mvidner:
https://coveralls.io/github/yast/yast-installation

Note that ALL clients are being measured (and the coverage is pretty low because of that)


jreidinger:
to be honest I do not like it. There is bunch of proposal and finish clients, that can be easily changed to new client library. This way it is just moving mess, which is not nice.

mvidner:
Yes, it is moving the mess *cheaply* and we get a coverage report. I am not making things worse, am I? https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good

jreidinger:
well, you add one redirection when someone debugging ( see client, then check lib and then see mess).
Another part is that when you want to make it testable, you still need to adapt it to be more object like and less script like.

mvidner:
You are right, but the purpose if this is NOT improving the code. It is improving the awareness, by enabling the reporting of the (bad) coverage.

The improvements you mention are all good, but IMHO they should be prioritized and should not block us from improving test coverage in, say [modules/ImageInstallation.rb](https://coveralls.io/builds/3102258/source?filename=src%2Fmodules%2FImageInstallation.rb), if we need to touch it soon.

If I had to rewrite all clients to the new class, I would never enable coveralls for this package.

ancorgs:
I agree with @mvidner. It's a pity not to fix things in a definitive way, but I don't think this is making the situation any worse.

jreidinger:
ok, if @ancorgs is not against, then proceed with it

ancorgs:
I'm not generally against, but I have not checked for typos or something like that (quite likely in massive changes). @jreidinger did you do it or should I?

jreidinger:
I hope mvidner use his magic tool, so no typos there :)

mvidner:
Yes, I've used a script to do the mass changes, so either there are no typos or we will find out really soon :)

PR #300, opened 2015-07-21
Do install the new files; mention the move in .changes.

mvidner:
Fixing #298 which omitted the new files, causing https://openqa.suse.de/tests/102837


mvidner:
@jreidinger said LGTM in person

PR #303, opened 2015-08-05
store cio_ignore settings before installing bootloader (bsc#933177)

jsrain:


jsrain:
Fix tested by Ihno

jsrain:
The fix is S/390-specific and requires specific set-up, therefore hard to be tested automatically.

kobliha:
LGTM

PR #306, opened 2015-08-13
[Review] Request from 'schubi2' @ 'yast/yast-installation/review_150813_cio_ignore'

schubi2:
  * 8719626 packaging
  * 39c2aac added new flag for cio_ignore in autoyast

### DIFF
--- a/src/lib/installation/cio_ignore.rb
+++ b/src/lib/installation/cio_ignore.rb
@@ -1,8 +1,12 @@
 require "yast"
 
 module Installation
+
   class CIOIgnore
     include Singleton
+    Yast.import "Mode"
+    Yast.import "AutoinstConfig"
+
     attr_accessor :enabled
 
     def initialize
@@ -10,8 +14,12 @@ module Installation
     end
 
     def reset
-      # default value requested in FATE#315586
-      @enabled = true
+      if Yast::Mode.autoinst
+        @enabled = Yast::AutoinstConfig.cio_ignore
+      else
+        # default value requested in FATE#315586
+        @enabled = true
+      end
     end
   end
 
--- a/src/lib/installation/cio_ignore.rb
+++ b/src/lib/installation/cio_ignore.rb
@@ -16,9 +16,7 @@ module Installation
       if Yast::Mode.autoinst
         @enabled = Yast::AutoinstConfig.cio_ignore
       else
-        if File.exist?("/proc/sysinfo") &&
-            (File.readlines("/proc/sysinfo").grep(/Control Program: KVM\/Linux/).any? ||
-              File.readlines("/proc/sysinfo").grep(/Control Program: z\/VM/).any?)
+        if is_kvm? || is_zvm?
           # cio_ignore does not make sense for KVM or z/VM (fate#317861)
           @enabled = false
         else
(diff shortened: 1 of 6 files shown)


src/lib/installation/cio_ignore.rb
```diff
@@ -10,8 +13,19 @@ def initialize
...
     def reset
-      # default value requested in FATE#315586
-      @enabled = true
+      if Yast::Mode.autoinst
+        @enabled = Yast::AutoinstConfig.cio_ignore
+      else
+        if File.exist?("/proc/sysinfo") &&
+            (File.readlines("/proc/sysinfo").grep(/Control Program: KVM\/Linux/).any? ||
+              File.readlines("/proc/sysinfo").grep(/Control Program: z\/VM/).any?)
+          # cio_ignore does not make sense for KVM or z/VM (fate#317861)
```
ancorgs: NTH: I would encapsulate this huge condition into one or two methods with meaningful names, so we can have something like
```ruby
if is_kvm? || is_zvm?
  # cio_ignore does not make sense here
  @enabled = false
else
```

test/cio_ignore_test.rb
```diff
@@ -4,6 +4,25 @@
...
 require "installation/cio_ignore"
 
+describe ::Installation::CIOIgnore do
+  describe "enable/disable" do
+
+    it "take AutoYaST cio_ignore setting if it is enabled" do
+      allow(Yast::Mode).to receive(:autoinst).and_return(true)
+      allow(Yast::AutoinstConfig).to receive(:cio_ignore).and_return(false)
+      ::Installation::CIOIgnore.instance.reset
+      expect(::Installation::CIOIgnore.instance.enabled).to eq(false)
```
ancorgs: This test is ok, but maybe it can be the case that it's returning false for a different reason. What I usually do in those cases to ensure the value is coming from there is using a completely unexpected value, like.

```
allow(Yast::AutoinstConfig).to receive(:cio_ignore).and_return("autoinst_value")
# and then...
expect(::Installation::CIOIgnore.instance.enabled).to eq("autoinst_value")
```

Some people could consider using those crazy values as a bad practice, though. I think the gain is bigger than the lost.

test/cio_ignore_test.rb
```diff
@@ -4,6 +4,25 @@
...
+      allow(Yast::Mode).to receive(:autoinst).and_return(false)
+      expect(Yast::AutoinstConfig).not_to receive(:cio_ignore)
+      ::Installation::CIOIgnore.instance.reset
+      expect(::Installation::CIOIgnore.instance.enabled).to eq(true)
+    end
+  end
+end
+
 describe ::Installation::CIOIgnoreProposal do
   subject { ::Installation::CIOIgnoreProposal.new }
```
ancorgs: btw, this is already the default value for `subject` if you use `describe ::Installation::CIOIgnoreProposal`

lslezak: See https://www.relishapp.com/rspec/rspec-core/v/3-3/docs/subject/implicitly-defined-subject for more details.

schubi2: But then I am getting: 
noMethodError:
       private method `new' called for Installation::CIOIgnore:Class

CIOIgnore is a singleton

ancorgs: I was commenting on `::Installation::CIOIgnoreProposal` (yes, I know it was already there, I simply saw it), not on `Installation::CIOIgnore`

ancorgs:
LGTM

PR #309, opened 2015-08-25
[Review] Request from 'schubi2' @ 'yast/yast-installation/review_150825_fixed_testcases'

schubi2:
  * 18767bf packaging
  * 4a58c34 fixed cio_ignore testcase


lslezak:
Just one small annoying thing: there is a missing `bsc` number, the OBS check would complain and the SR would be rejected...

Otherwise LGTM.

lslezak:
LGTM

PR #311, opened 2015-09-02
Move second_stage_required?

imobach:
Move second_stage_required? method so it can be used by other YaST modules.


imobach:
@aschnell To be used from Ruby, it's not required. But maybe it makes sense to add it so maybe we can use it from Perl. Let's ask @mvidner.

mvidner:
So far we haven't needed it from Perl. If we did, we would have to alias the method because I doubt Perl can handle the trailing question mark.

imobach:
@mvidner @aschnell Thanks. I've published the method and also aliased it.

mvidner:
LGTM

imobach:
Thanks to both of you.

PR #313, opened 2015-09-03
Fix languages during upgrade

kobliha:
- bsc#943746, bsc#944035
- P1!
- Setting language/languages was using wrong data type string vs array


imobach:
LGTM

PR #315, opened 2015-09-29
- fixed typo (variable name and log message)

aschnell:


jreidinger:
I worry that without some bug or fate ID it do not pass to SP1. Otherwise LGTM

mvidner:
I have merged it, with actually removing the .changes message; it made more sense that way.

PR #316, opened 2015-10-01
Remove systemd dependency on xinetd

ancorgs:
It was introduced to avoid a very remote possibility of race condition. Now it's double checked in the startup script.

Tested manually with a successful vnc+autoyast installation and also simulating the race condition.

@imobach is running it through the entire autoyast test suite.

### DIFF
--- a/package/YaST2-Second-Stage.service
+++ b/package/YaST2-Second-Stage.service
@@ -2,7 +2,7 @@
 Description=YaST2 Second Stage
 # If xinetd is enabled, make sure it's already running so we can stop it during
 # initialization of the VNC server
-After=apparmor.service local-fs.target plymouth-start.service xinetd.service
+After=apparmor.service local-fs.target plymouth-start.service
 Before=getty@tty1.service display-manager.service network.service NetworkManager.service SuSEfirewall2_init.service SuSEfirewall2.service
 ConditionPathExists=/var/lib/YaST2/runme_at_boot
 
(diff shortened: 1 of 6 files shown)


imobach:
Basics tests confirm that it does not break anything. However, trying with sp1-beta1 (as autoyast-integration-test could not run rc1 at this time), I get: `/usr/bin/Xvnc: error while loading shared libraries: libgnutls.so.28: cannot open shared object file: No such file or directory`. Manual testing, anyway, looks good.

ancorgs:
@imobach And testing sp1-beta1 without this patch produces the same error? That's what I would expect, since this change has NOTHING to do with how Xvnc is compiled, but.. you know... just in case.

imobach:
I was wrong: I was using RC1. I'll try now without that patch, but I expect the same result.

ancorgs:
@imobach And do you expect beta without the patch to work properly? And beta with the patch?

imobach:
Just to summarize (using RC1):

* Using VNC, with and without the patch, it fails in the same way.
* Not using VNC it works (as it does not need to launch `Xvnc`).

ancorgs:
So everything looks the same with and without the patch in AY tests. And all my manual tests worked. And already LGTMed. So there we go.

kobliha:
@ancorgs What happens if xinetd was running and we shut it down? Shouldn't we start it again when Yast finishes? IMO looks like an inconsistent state...

ancorgs:
@kobliha We actually start it right after Xvnc has taken the ports (see the `restore_xinetd` line). So we only shut xinetd down during one second or so.

PR #319, opened 2015-10-26
Poc recover values

jreidinger:
not to merge yet ( planned for SP2 ) and now only RFC way how to handle reusable data.

### DIFF
--- a/src/clients/inst_complex_welcome.rb
+++ b/src/clients/inst_complex_welcome.rb
@@ -75,27 +75,25 @@ module Yast
 
   private
     def event_loop
-      ret = nil
-
       loop do
         ret = UI.UserInput
         Builtins.y2milestone("UserInput() returned %1", ret)
 
-        if ret == :back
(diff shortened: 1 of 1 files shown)


lslezak:
Looks good, I'm just thinking whether it would make sense to share the storing/loading the reusable data in some base class. Is something planned?

Something similar to what we already have for the proposal clients ([proposal_client.rb](https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/installation/proposal_client.rb))...

jreidinger:
@lslezak I also consider it, but there is only few clients need it and in the end it is simple Yaml based storing.

kobliha:
Good point @lslezak I'd like to see the data at one place using one API. It might be YAML, but with a shared interface, 0.02 EUR

imobach:
This was implemented in #348 as part of fate#319716, so I'm closing this PR.

Thanks!

PR #320, opened 2015-11-05
Don't let getty kill YaST with SIGHUP on serial console (bsc#935965).

mvidner:
[bsc#935965](https://bugzilla.suse.com/show_bug.cgi?id=935965)

PR #323, opened 2015-11-19
Don't let getty kill YaST on further serial consoles (bsc#935965)

mvidner:
[bsc#935965](https://bugzilla.suse.com/show_bug.cgi?id=935965)
https://trello.com/c/q47excv8
A follow-up to #320.
It turns out that the failure the customer reported was a real problem with our earlier fix, caused by an even more uncommon setup.


mvidner:
Better wait for a validation from the customer before merging :)

mvidner:
Fix confirmed by customer to work.

jreidinger:
LGTM, maybe just add there comment line explaining why we need S1 and S2

PR #326, opened 2015-11-26
Fix non available proposal

jreidinger:


package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Nov 26 09:09:59 UTC 2015 - jreidinger@suse.com
+
+- Do not crash in proposal client if zfcp proposal report itself as
```
imobach: typo: "reports" ?

imobach:
Rubocop found a minor issue.

src/lib/installation/proposal_store.rb
```diff
@@ -194,6 +201,8 @@ def description_for(client)
 
       description = Yast::WFM.CallFunction(client, ["Description", {}])
 
+      return nil unless description
```
imobach: Is ok to return nil? I mean, `description_for` is called also https://github.com/yast/yast-installation/blob/fix_non_available_proposal/src/lib/installation/proposal_store.rb#L227. Maybe it never won't be called in a situation when `description_for` is `nil`, but I just want to be sure.

package/yast2-installation.changes
jreidinger: right, will fix it

test/proposal_store_test.rb
```diff
@@ -94,6 +98,17 @@ def mock_properties(data = {})
...
+        .and_return([
+          ["test1"],
+          ["test2"]
+        ])
+
+      expect(subject.proposal_names).to include("test1")
+      expect(subject.proposal_names).to include("test2")
+    end
+
+    it "filter out from result all clients that do not exist on media" do
```
mvidner: Just English nitpicking:

> filters out clients that do not exist on media

test/proposal_store_test.rb
```diff
@@ -108,6 +123,21 @@ def mock_properties(data = {})
       expect(subject.proposal_names).not_to include("test3")
     end
 
+    it "filter out clients that itself report as unavailable" do
```
mvidner: > filters out clients that report themselves as unavailable

src/lib/installation/proposal_store.rb
mvidner: There are multiple places in this file that access the return value without cheking for nil: `id_for`, `title_for`, ...

jreidinger: all usage depends on proposal_names, which gives list of clients that can be used. So already filtered out clients with nil.

jreidinger: I want to also keep change small, to avoid side effects. At least now it is not worse, as now it will crash if it return nil in description call.

imobach:
LGTM

src/lib/installation/proposal_store.rb
imobach: I only wanted to be sure that it's not worse :)

src/lib/installation/proposal_store.rb
```diff
@@ -143,6 +143,13 @@ def proposal_names
 
       # Filter missing proposals out
       @proposal_names -= missing_proposals
+
+      unavailable_proposals = @proposal_names.select { |name| description_for(name).nil? }
+      unless unavailable_proposals.empty?
+        log.info "These proposals report itself as unavailable: #{unavailable_proposals}"
```
mvidner: I think this is misleading. AFAIK we are handling a case where a proposal has encountered an error condition and returned `nil` because of that. It is not an API for "report as unavailable", it is a "rescue clause", right? So we should log
> These proposals returned an error, disabling them:

Otherwise we should document the `nil` in the client API http://www.rubydoc.info/github/yast/yast-yast2/Installation/ProposalClient:description

jreidinger: no see e.g. https://github.com/yast/yast-s390/blob/master/src/clients/zfcp_proposal.rb#L66 and same for dasd, So we probably need to document it in yast2. but I vote to document it only for master, not SP1

test/proposal_store_test.rb
```diff
@@ -108,6 +123,21 @@ def mock_properties(data = {})
       expect(subject.proposal_names).not_to include("test3")
     end
 
+    it "filters out clients that report themselves as unavailable" do
```
mvidner: ... which means that now I would rather correct myself to
> filters out clients that return an error

src/lib/installation/proposal_store.rb
mvidner: OK. So the code looks good. But [as requested](https://bugzilla.suse.com/show_bug.cgi?id=956745#c17) the previous fix that Lada made should be reverted.

jreidinger: yes, I will handle it directly in SLE-12-SP1 branch

mvidner:
LGTM now, thank you!

PR #329, opened 2015-12-02
Don't let getty kill YaST on further serial consoles (bsc#935965) SP1

mvidner:
Merge #323 (from SLE12-GA) into the SP1 codebase


ancorgs:
LGTM

PR #330, opened 2015-12-02
Don't let getty kill YaST on further serial consoles (bsc#935965) master

mvidner:
Merges the original fix #323.
Actually this is a merge of the whole after-SP1 branch so it includes #326 too.


imobach:
LGTM

PR #331, opened 2015-12-11
don't enforce the disk activation dialog on zKVM (bsc#956736)

jsrain:


jsrain:
Depends on https://github.com/yast/yast-yast2/pull/418

package/yast2-installation.spec
```diff
@@ -38,7 +38,7 @@ BuildRequires:  yast2-devtools >= 3.1.10
 BuildRequires:  rubygem(rspec)
 
 # Base clients for inst clients
-BuildRequires: yast2 >= 3.1.126
+BuildRequires: yast2 >= 3.1.162
```
jreidinger: please update also comment why we need so high version above.

jreidinger:
just NP, otherwise LGTM

coolo:
you even add a comment in the spec file why yast is required and then use a non-existant function in the code? ;(

jsrain:
Missed the is_ prefix of the function name :-(

jsrain:
Fixed in https://github.com/yast/yast-installation/pull/333

PR #334, opened 2016-02-17
Remove autoyast clone button.

cwh42:


src/lib/installation/proposal_runner.rb
```diff
@@ -654,21 +640,14 @@ def skip_buttons
 
     def build_dialog
       headline = @store.headline
+      change_point = Empty()
```
imobach: np: I would move this to the `else` of the following `if` condition. But it's just a matter of taste.

imobach:
LGTM. I guessed you tested it manually, right? Maybe other teammate could think of some forgotten place. I can't :)

lslezak:
After removing the `clone_finish` client you should remove it also from the [list of `*_finish` clients](https://github.com/yast/yast-installation/blob/master/src/lib/installation/clients/inst_finish.rb#L152).

lslezak:
It seems you need to update the `Makefile.am` file as well: https://travis-ci.org/yast/yast-installation/builds/109902915#L1097

cwh42:
@lslezak All green, now.

src/lib/installation/proposal_runner.rb
cwh42: It actually was there before, but my personal taste told me to move it outside the condition.

lslezak:
Thanks, LGTM

PR #340, opened 2016-02-26
System Role (FATE#317481)

mvidner:
https://fate.suse.com/317481 Add 'System Role' capabilities to installer


mvidner:
Needs control file extensions (run the code; define the roles)  https://github.com/yast/skelcd-control-SLES/pull/53
Will need ProductFeatures extensions https://github.com/yast/yast-yast2/pull/446

doc/control-file.md
```diff
@@ -842,6 +842,59 @@ Section *supported\_desktops* contains list of one or more
...
+      </virtualization_host_kvm>
+      <virtualization_host_kvm_description>
+        <label>Will install the appropriate packages.
+Will use LVM disk layout.</label>
+      </virtualization_host_kvm_description>
+    </texts>
+</productDefines>
+```
+
+Each role has a short label and a few lines of description in the *texts*
```
jreidinger: what is advantage of this solution over having label and description in each role?

src/lib/installation/select_system_role.rb
```diff
@@ -0,0 +1,121 @@
...
+module Installation
+  class SelectSystemRole < ::UI::InstallationDialog
+    include Yast
+
+    def initialize
+      super
+
+      textdomain "installation"
+      Yast.import "ProductControl"
+      Yast.import "ProductFeatures"
```
jreidinger: yast import is like require, so better to have it on top, unless you want lazy loading.

src/lib/installation/select_system_role.rb
```diff
@@ -0,0 +1,121 @@
...
+
+#  To contact SUSE about this file by physical or electronic mail,
+#  you may find current contact information at www.suse.com
+
+require "yast"
+require "ui/installation_dialog"
+
+module Installation
+  class SelectSystemRole < ::UI::InstallationDialog
+    include Yast
```
jreidinger: why include whole yast namespace? usually in lib we include only modules we need,

src/lib/installation/select_system_role.rb
```diff
@@ -0,0 +1,121 @@
...
+    def run
+      if raw_roles.empty?
+        log.info "No roles defined, skipping their dialog"
+        return :auto            # skip forward or backward
+      end
+
+      super
+    end
+
+    public                      # called by parent class
```
jreidinger: ugh, by default it is public, I do not get why needed?

src/lib/installation/select_system_role.rb
```diff
@@ -0,0 +1,121 @@
...
+        VBox(
+          * ui_roles.map do |r|
+              VBox(
+                Left(RadioButton(Id(r[:id]), r[:label])),
+                HBox(
+                  HSpacing(4),
+                  Left(Label(r[:description]))
+                ),
+                VSpacing(2)
+              )
```
jreidinger: quite deep nesting. What helps is something like roles_ui = ui_roles.map and then VBox(*ui_roles)

BTW what is purpose of VBox containing list of VBoxes?

src/lib/installation/select_system_role.rb
```diff
@@ -0,0 +1,121 @@
...
+    # [
+    #  { "id" => "foo", "partitioning" => ... },
+    #  { "id" => "bar", "partitioning" => ... , "software" => ...},
+    # ]
+    # @return [Array<Hash{String => Object}>]
+    def raw_roles
+      ProductControl.productControl.fetch("system_roles", [])
+    end
+
+    def ui_roles
```
jreidinger: quite confusing as ui_roles is in fact role attributes and not UI. UI will be for me if it returns UI term.

test/select_system_role_test.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  describe "#run" do
+    context "when no roles are defined" do
+      before do
+        allow(Yast::ProductControl).to receive(:productControl)
+          .and_return("system_roles" => [])
+      end
+
+      it "does not display dialog" do
```
jreidinger: it should also specify that it returns auto as it is important part, which hide it when going forward and also backward

test/select_system_role_test.rb
```diff
@@ -0,0 +1,71 @@
...
+        expect(Yast::Wizard).to_not receive(:SetContents)
+        subject.run
+      end
+    end
+
+    context "when some roles are defined" do
+      include Yast::UIShortcuts
```
jreidinger: it should not be needed, ruby bindings rspec helper should already contain it

test/select_system_role_test.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      let(:control_file_roles) do
+        [
+          { "id" => "foo", "partitioning" => { "format" => true } },
+          { "id" => "bar", "software" => { "desktop" => "knome" } }
```
jreidinger: knome :laughing:

doc/control-file.md
mvidner: I copied the style used for the system scenarios. I thought otherwise translating the texts would not be possible but now I am not so sure.

src/lib/installation/select_system_role.rb
mvidner: ProductControl has a constructor loading the control file so I think I need lazy loading.

src/lib/installation/select_system_role.rb
mvidner: Er, I had `protected` originally, to delimit the methods implementing the InstallationDialog API, but that did not work (because EventHandler asks `respond_to?`).

src/lib/installation/select_system_role.rb
mvidner: An alternative is to return a [Left, Hbox, Vspacing] from the map block. Then you end up with a nested array that you cannot just `*` splat, can you?

src/lib/installation/select_system_role.rb
mvidner: Well, yes, you are right, but I do not have a better idea. `role_attributes`?

test/select_system_role_test.rb
mvidner: right! I forgot that

test/select_system_role_test.rb
mvidner: yes. actually I have an old system so I needed it locally.

test/select_system_role_test.rb
mvidner: You cannot stop Prokress!

src/lib/installation/select_system_role.rb
jreidinger: no, it will require to use reduce or each with object and fill VBox in it

src/lib/installation/select_system_role.rb
mvidner: OK, I'll try without it.

src/lib/installation/select_system_role.rb
jreidinger: so can be removed now? btw it can be change to respond_to with true as second arg

src/lib/installation/select_system_role.rb
jreidinger: ok, make sense, maybe document it?

src/lib/installation/select_system_role.rb
jreidinger: or really return term and only another usage is getting first radio_button id which is easy with nested find search functionality in term

src/clients/testme.rb
```diff
@@ -0,0 +1,14 @@
+# a quick test of dialog layout without running the whole installation business
```
jreidinger: hmm, it is distributed as client in rpm. I think it is not so nice. what about moving it somewhere like manual_test directory?

src/lib/installation/select_system_role.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    def dialog_content
+      ui_roles = role_attributes.each_with_object(VBox()) do |r, vbox|
+        vbox << Left(RadioButton(Id(r[:id]), r[:label]))
+        vbox << HBox(
+          HSpacing(4),
+          Left(Label(r[:description]))
+        )
+        vbox << VSpacing(2)
+      end
```
jreidinger: :+1:

test/select_system_role_test.rb
jreidinger: well, unless we plan to backport it to legacy release I think it make sense to benefit from new features that makes live easier

jreidinger:
LGTM

mvidner:
Screenshots of a testing build of SLES12-SP2:
In the installation workflow, after the Registration and optional Add-ons, we come to selecting the System Role:
system-role-mar14
Then in the next step we can see the effect. It is either the Default System:
system-role-mar14-default-red
or the KVM Virtualization Host:
system-role-mar14-kvm-red

PR #341, opened 2016-02-29
Fix macro expansion in %post comment

Vogtinator:
rpm macros get expanded in comments as well.
This one leads to weird errors during installation,
like "Failed to execute operation: No such file or directory".


mvidner:
Thank you!
I couldn't reproduce the error message, but the macro getting expanded is obviously wrong.

PR #346, opened 2016-03-07
update to the latest yast rubocop style

jreidinger:
Interested fact here is that due to disablement of check there is 7 outdent and 4 indent of access modifier, so this unify it.


lslezak:
LGTM

jreidinger:
travis failed due to not yet published new devtools, will rebuild when published

mvidner:
Now that https://github.com/yast/yast-installation/pull/340 is merged (thanks!) there is one more `private` to reindent.

PR #347, opened 2016-03-11
Self update

imobach:
This PR is a work in progress, although is almost ready (some final testing is due).

* Add a client to perform the installer update
* Add classes UpdatesManager and DriverUpdate to handle the update process.


src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,185 @@
...
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+
+require "installation/updates_manager"
+require "uri"
+
+module Yast
+  class InstUpdateInstaller < Client
```
jreidinger: actually for newer clients I prefer to not use this base class. As it is useful mainly for old clients. For new ones use includes which you need.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,185 @@
...
+    def self_update_url_from_linuxrc
+      url = Linuxrc.InstallInf("SelfUpdate") || ""
+      valid_url?(url) ? URI(url) : nil
+    end
+
+    # Return the self-update URL according to product's control file
+    #
+    def self_update_url_from_control
+      url = ProductFeatures.GetStringFeature("globals", "self_update_url")
+      valid_url?(url) ? URI(url) : nil
```
jreidinger: looks like duplicite code :)

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,185 @@
...
+    # Check if installer was updated
+    #
+    # It checks if a file UPDATED_FLAG_FILENAME exists in Directory.vardir
+    #
+    # @return [Boolean] true if it exists; false otherwise.
+    def installer_updated?
+      File.exist?(update_flag_file)
+    end
+
+    # Returns the name of the "update flag file"
```
jreidinger: this say it returns name, but below is specified path, which is correct

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,185 @@
...
+    #
+    # It also shows feedback to the user.
+    #
+    # @return [Boolean] true if installer was updated; false otherwise.
+    def update_installer
+      if fetch_update
+        apply_update
+      else
+        false
+      end
```
jreidinger: what about
```
return false unless fetch update

apply_update
```

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,185 @@
...
+
+    # Fetch updates from self_update_url
+    #
+    # @return [Boolean] true if update was fetched successfully; false otherwise.
+    def fetch_update
+      ret = nil
+      Popup.Feedback(_("YaST2 update"), _("Searching for installer updates")) do
+        ret = updates_manager.add_update(self_update_url)
+      end
+      Popup.Error(_("Update could not be found")) unless ret || using_default_url?
```
jreidinger: Better to use here Report.Error otherwise it stuck autoyast installation if you wan to use same client there.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,185 @@
...
+    # @return [Boolean] true if the update was applied; false otherwise
+    def apply_update
+      if updates_manager.all_signed? || insecure_mode? || ask_insecure?
+        Popup.Feedback(_("YaST2 update"), _("Applying installer updates")) do
+          updates_manager.apply_all
+        end
+      else
+        false
+      end
+    end
```
jreidinger: a bit complicated workflow, what about
```
return false if denied_insecure?

Popup.Feedback(_("YaST2 update"), _("Applying installer updates")) do
    updates_manager.apply_all
 end
```

and denied_insecure contain those three conditions?

src/lib/installation/clients/inst_update_installer.rb
jreidinger: and second note, I found quite hard to decode meaning of unless with more then one condition. Here it means do not report if ret is true or default_url is used.
I am also not sure if this popup is good idea as some company can have own update url, which is same as SUSE is filled when it is needed, so it is expected to be empty from beginning ( like running installation over PXE with default params passed ).

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+require "tempfile"
+require "pathname"
+require "transfer/file_from_url"
+
+module Installation
+  # Represents a driver update disk (DUD)
+  #
+  # The DUD will be fetched from a given URL.
+  class DriverUpdate
+    include Yast::I18n # missing in yast2-update
```
jreidinger: I18n used, but textdomain missing
Also do not get what comment mean?

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+    include Yast::Transfer::FileFromUrl # get_file_from_url
+
+    class NotFound < StandardError; end
+
+    # Command to extract the content of the DUD
+    EXTRACT_CMD = "gzip -dc %<source>s | cpio --quiet --sparse -dimu --no-absolute-filenames"
```
jreidinger: why adding `s` here? btw do you consider using new Yast::Execute library?

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+    SIG_EXT = ".asc"
+
+    attr_reader :uri, :local_path, :keyring, :gpg_homedir
+
+    # Constructor
+    #
+    # @param uri         [URI]      Driver Update URI
+    # @param keyring     [Pathname] Path to keyring to check signatures against
+    # @param gpg_homedir [Pathname] Path to GPG home dir
+    def initialize(uri, keyring: keyring, gpg_homedir: Pathname.new("/root/.gnupg"))
```
jreidinger: ugh what? `keyring: keyring` is same as `keyring: nil` just obscured, unless I overlook something.

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+    # @return [Boolean] True if it's correctly signed; false otherwise.
+    def signed?
+      @signed
+    end
+
+    # Fetch the DUD and extract it in the given directory
+    #
+    # @param target [Pathname] Directory to extract the DUD to.
+    def fetch(target)
+      @local_path = target
```
jreidinger: this indicates quite nasty side effects. Why not simple passing it where needed?

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+      Dir.mktmpdir do |dir|
+        Dir.chdir(dir) do
+          temp_file = Pathname.pwd.join(TEMP_FILENAME)
```
jreidinger: it fact easier then pwd is to use `File.join(dir, TEMP_FILENAME)`

jreidinger: so in return it looks like chdir is not needed, or am I wrong?

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+      end
+    end
+
+    # Determine if gpg command was successful
+    #
+    # The signature was successfully signed if command error code was 0 and
+    # no warning was shown.
+    #
+    # @return [Boolean] True if signature check was successful; false otherwise.
+    def check_gpg_output(out)
```
jreidinger: why is this public method? I do not feel it is right place to have public gpg output checker here

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+      out["exit"].zero? && !out["stderr"].include?("WARNING")
+    end
+
+    # Apply the DUD to inst-sys
+    #
+    # @see #adddir
+    # @see #run_update_pre
+    def apply
+      raise "Driver updated not fetched yet!" if local_path.nil?
```
jreidinger: regarding API, I think that unless there is really strong, then such dependent call should not be used. Why not having simple apply, that do fetch inside and have return value depending if system is updated or not?

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+
+    # Extract the DUD at 'source' to 'target'
+    #
+    # @param source [Pathname]
+    #
+    # @see EXTRACT_CMD
+    def extract(source, target)
+      cmd = format(EXTRACT_CMD, source: source)
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), cmd)
+      raise "Could not extract DUD" unless out["exit"].zero?
```
jreidinger: for free with cheetah alias Yast::Execute :)

src/lib/installation/clients/inst_update_installer.rb
imobach: Yes, it is :) You know, I've not done my own review :)

src/lib/installation/clients/inst_update_installer.rb
imobach: I think `fetch_update ? apply_update : false` would be even better.

src/lib/installation/clients/inst_update_installer.rb
imobach: The idea if that if you're using a custom url (adding `SelfUpdate=my-own-url` in Linuxrc), you should be warned that the update wasn't found. If you're using the one in control.xml, it means that there's no update available (which is fine).

src/lib/installation/clients/inst_update_installer.rb
imobach: +1

src/lib/installation/driver_update.rb
imobach: Transfer::FileFromUrl does not use Yast::I18n but it calls to `textdomain`. I saw Martin doing this workaround, but maybe I should fix the offender.

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+    end
+
+    # Determine if gpg command was successful
+    #
+    # The signature was successfully signed if command error code was 0 and
+    # no warning was shown.
+    #
+    # @return [Boolean] True if signature check was successful; false otherwise.
+    def check_gpg_output(out)
+      out["exit"].zero? && !out["stderr"].include?("WARNING")
```
jreidinger: so warnings are treated as errors? looks a bit permissive to me

src/lib/installation/driver_update.rb
imobach: I didn't remember it :) I'll check it.

src/lib/installation/driver_update.rb
imobach: A typo :( It should be `keyring: nil`.

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+    def setup_target(dir)
+      ::FileUtils.rm_r(dir) if dir.exist?
+      ::FileUtils.mkdir_p(dir.dirname) unless dir.dirname.exist?
+    end
+
+    # Download the DUD to a file
+    #
+    # If the file is not downloaded, DriverUpdate::NotFound exception is risen.
+    #
+    # @return [True] true if download was successful
```
jreidinger: useless as it cannot return anything else

src/lib/installation/driver_update.rb
imobach: I consider it object's state. If it's not extracted, @local_path is nil.

Well, the idea is that you 1) fetch a DUD and you can 2) extract it to some place. And I would like that information to be stored in the object. I could fetch+extract in the constructor, but I was not comfortable with that idea.

src/lib/installation/driver_update.rb
imobach: I tend to use `Pathname` as much as possible. About the chdir, I think I can get most of the code out of it.

src/lib/installation/driver_update.rb
imobach: Yes, it was intended to be private.

src/lib/installation/driver_update.rb
imobach: Well, it can throw and exception.

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+      path.relative_path_from(Pathname.new("/"))
+    end
+
+    # Add files/directories to the inst-sys
+    #
+    # @see APPLY_CMD
+    def adddir
+      cmd = format(APPLY_CMD, source: local_path)
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), cmd)
+      out["exit"].zero?
```
jreidinger: so here no exception just return boolean? looks a bit inconsistent to me.

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,223 @@
...
+    end
+
+    # Run update.pre script
+    #
+    # @return [Boolean,NilClass] true if execution was successful; false if
+    #                            it failed; nil if script does not exist or
+    #                            was not executable.
```
jreidinger: ternary logic is a bit pain

jreidinger: btw it is used in apply as return value, but not documented there at all, so result is that it is very easy to check for false above. Reason why I prefer exceptions over boolean returns.

src/lib/installation/driver_update.rb
imobach: To be honest, I think that adddir does not return a different error code if it fails. I have to check it.

src/lib/installation/updates_manager.rb
```diff
@@ -0,0 +1,92 @@
...
+  #   manager.add_update(URI("http://update.opensuse.org/sles12.dud"))
+  #   manager.add_update(URI("http://example.net/example.dud"))
+  #   manager.fetch_all
+  #   manager.apply_all
+  #
+  # @example Applying multiple driver updates
+  #   manager = UpdatesManager.new
+  #   manager.add_update(URI("http://update.opensuse.org/sles12.dud"))
+  #   manager.fetch_all
+  #   manager.apply_all
```
jreidinger: good to have doc with examples :+1:

jreidinger: but it also shows that it have fetch_all and apply_all together. What is reason to have it separated?

src/lib/installation/updates_manager.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    # Add an update to the updates pool
+    #
+    # @param uri [URI]                               URI where the update (DUD) lives
+    # @return    [Array<Installation::DriverUpdate>] List of updates
+    #
+    # @see Installation::DriverUpdate
+    def add_update(uri)
+      new_update = Installation::DriverUpdate.new(uri)
+      dir = target.join(format("%03d", next_update))
```
jreidinger: why this format is needed?

src/lib/installation/updates_manager.rb
```diff
@@ -0,0 +1,92 @@
...
+      new_update.fetch(dir)
+      @updates << new_update
+    rescue Installation::DriverUpdate::NotFound
+      false
+    end
+
+    # Fetches all updates in the pool
+    def fetch_all
+      shift = next_update
```
jreidinger: why it check only for one number and not all of it? what if there is a gap?

src/lib/installation/driver_update.rb
jreidinger: still return value should be void and exception is documented with `@raise`. As usage is 
```
begin
  do_something
rescue
   handle_problem
end
```
if you document return value it indicate you should keep it with something like ` value = do_something` but it is useless as it can return only true.

src/lib/installation/updates_manager.rb
imobach: Yep, documentation is outdated as `fetch_all` is not used anymore. Only the `add_update` and `apply_all` methods are needed.

src/lib/installation/updates_manager.rb
imobach: Because Linuxrc uses `001`, `002`, etc.

src/lib/installation/updates_manager.rb
imobach: No problem if there's a gap. Although we shouldn't fill those gaps, as it would affect `update.pre` scripts execution order.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,185 @@
...
+      Linuxrc.InstallInf("Insecure") == "1" # Insecure mode is enabled
+    end
+
+    # Ask the user if she/he wants to apply the update although it's not properly signed
+    #
+    # @return [Boolean] true if user answered 'Yes'; false otherwise.
+    def ask_insecure?
+      Popup.AnyQuestion(
+        Label::WarningMsg(),
+        _("Installer update is not signed or signature is invalid. Do you want to apply this update?"),
```
lslezak: I'd add some more details for beginner users that using usigned/invalid signature is a security risk.

See e.g. [this](https://github.com/yast/yast-registration/blob/master/src/lib/registration/ui/import_certificate_dialog.rb#L140), [this](https://github.com/yast/yast-yast2/blob/master/library/packages/src/modules/SignatureCheckDialogs.rb#L464) or [this](https://github.com/yast/yast-yast2/blob/master/library/packages/src/modules/SignatureCheckDialogs.rb#L552) example.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,185 @@
...
+        _("Yes, apply and continue"),
```
lslezak: Did you ask Ken about this? Personally I'd prefer for a clear `Do you want to ...?`  question simple `Yes`/`No` buttons.

lslezak:
BTW is there any documentation how to build and sign that .dud files?

I think that the documentation will be a very important part of this feature...

src/lib/installation/clients/inst_update_installer.rb
imobach: No, texts are not definitive in any way. I've to check them with Ken.

imobach:
@lslezak I'll search if there's some official issue (as support engineers create them). On the other hand, I agree documentation is really important in this feature.

src/lib/installation/driver_update.rb
imobach: Well, I've checked those cases:

* Good signature and PGP key known. Exit code: 0 no warnings.
* Good signature and PGP key not trusted. Exit code: 0 and 1 warning.
* PGP key not known. Exit code: 2 and 1 warning.
* Bad signature. Exit code: 2 and 1 warning.

Maybe it's too strict, but I would only allow the 1st scenario.

imobach:
More logging is missing... I'll work on that.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,195 @@
...
+
+    # Return the self-update URL according to product's control file
+    #
+    def self_update_url_from_control
+      get_url_from(ProductFeatures.GetStringFeature("globals", "self_update_url"))
+    end
+
+    # Converts the string into an URI if it's valid
+    #
+    # @return [Boolean] True if it's valid; false otherwise.
```
lslezak: Actually it is `[URI,nil]`.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,195 @@
...
+    # @return [Boolean] true if user answered 'Yes'; false otherwise.
+    def ask_insecure?
+      Popup.AnyQuestion(
+        Label::WarningMsg(),
+        _("Installer update is not signed or the signature is invalid.\n" \
+          "Using this update may put the integrity of your system at risk.\n" \
+          "Use it anyway?"),
+        _("Yes"),
+        _("No"),
+        :focus_yes
```
lslezak: To be on the safe side we should rather set `No` as the default. Users really need to think about this and actively confirm the insecure update, blindly pressing `Next`, `Next`... can be dangerous in cases like this.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,195 @@
...
+    # Check whether the update is allowed to be applied
+    #
+    # It's allowed to be applied when one of these requirements is met:
+    #
+    # * All updates are signed.
+    # * We're running in insecure mode (so we don't need them to be signed).
+    # * The user requests to install it although is not signed.
+    #
+    # @report [Boolean] true if it should be applied; false otherwise.
+    def allowed_to_be_applied?
```
jreidinger: what about shorted name `apply_it?`

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,195 @@
...
+    # @see #self_update_url_from_control
+    def self_update_url
+      self_update_url_from_linuxrc || self_update_url_from_control
+    end
+
+    # Return the self-update URL according to Linuxrc
+    #
+    # @return [URI,nil] self-update URL. nil if no URL was set in Linuxrc.
+    def self_update_url_from_linuxrc
+      get_url_from(Linuxrc.InstallInf("SelfUpdate") || "")
```
lslezak: Is the `|| ""` part actually needed? The `get_url_from` method returns `nil` for the `nil` argument and for `""` as well.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,195 @@
...
+    def update_installer
+      fetch_update ? apply_update : false
+    end
+
+    # Fetch updates from self_update_url
+    #
+    # @return [Boolean] true if update was fetched successfully; false otherwise.
+    def fetch_update
+      ret = nil
+      Popup.Feedback(_("YaST2 update"), _("Searching for installer updates")) do
```
lslezak: NP: I'd use `YaST Update` (title style capitalization for the dialog title, `2` should not be used), and I'd add ellipsis (`...`) at the end of the message to indicate progress.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,195 @@
...
+      Report.Error(_("Update could not be found")) unless ret || using_default_url?
+      ret
+    end
+
+    # Apply the updates and shows feedback information
+    #
+    # @return [Boolean] true if the update was applied; false otherwise
+    def apply_update
+      return false unless allowed_to_be_applied?
+      Popup.Feedback(_("YaST2 update"), _("Applying installer updates")) do
```
lslezak: Same here.

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,229 @@
...
+    # Determine if gpg command was successful
+    #
+    # The signature was successfully signed if command error code was 0 and
+    # no warning was shown.
+    #
+    # @return [Boolean] True if signature check was successful; false otherwise.
+    #
+    # @see SIGNATURE_OK
+    def check_gpg_output(out)
+      out["exit"].zero? && out["stderr"].include?(SIGNATURE_OK)
```
lslezak: NP: This does not exactly match the comment above: `no warning was shown` vs. `include?("gpg: Good signature")`. The action should be rather described as `error code was 0 and a good signature was reported`.

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,229 @@
...
+    # @see APPLY_CMD
+    def adddir
+      cmd = format(APPLY_CMD, source: local_path)
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), cmd)
+      raise CouldNotBeApplied unless out["exit"].zero?
+    end
+
+    # Run update.pre script
+    #
+    # @return [Boolean,NilClass] true if execution was successful; false if
```
lslezak: Yardoc uses `nil` type for the `nil` value, see the [yardoc literals](http://www.rubydoc.info/gems/yard/file/docs/Tags.md#Literals) documentation.

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,229 @@
...
+    #                            update script didn't exist.
+    # @raise DriverUpdate::CouldNotBeApplied
+    def run_update_pre
+      update_pre_path = local_path.join("install", "update.pre")
+      return false unless update_pre_path.exist? && update_pre_path.executable?
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), update_pre_path.to_s)
+      raise CouldNotBeApplied unless out["exit"].zero?
```
lslezak: I think it should use a different exception here (probably derived from the `CouldNotBeApplied`). With a specific exception we could report more details about the failure to the user.

This will be especially important if you build the update by ourselves, then you need to know that if failed at the *pre* script, not because of broken DUD file etc.

Also some error logging will be useful for debugging later.

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,229 @@
...
+
+    # Check a detached signature for a DUD
+    #
+    # The signature will be taken from the same URL than the DUD but adding
+    # the suffix '.asc'.
+    #
+    # @return [Boolean] True if the signature is OK; false otherwise.
+    #
+    # @see #check_gpg_output
+    def check_detached_signature(temp_file)
```
lslezak: NP: What about extracting the GPG functionality into a separate class? The GPG check looks like a pretty self contained functionality to me. It could also make the unit tests simpler / more readable.

imobach: Yes, I was considering that. But also I saw that we have a GPG module (but does not fit our requirements). I'd like to postpone the discussion about this point.

lslezak: Actually the test looks OK so not a big deal.

test/updates_manager_test.rb
```diff
@@ -0,0 +1,105 @@
...
+      end
+
+      it "returns an array containing the updates" do
+        expect(manager.updates).to eq([update0])
+      end
+    end
+  end
+
+  describe "#apply_all" do
+    let(:update0) { double("update0") }
```
lslezak: There already is `update0` defined above (in the global scope). Either change the numbering or reuse the previously defined doubles. This looks a bit confusing and could break when moving parts of the code in the file.

test/updates_manager_test.rb
```diff
@@ -0,0 +1,105 @@
...
+    it "applies all the updates" do
+      allow(manager).to receive(:updates).and_return([update0, update1])
+      expect(update0).to receive(:apply)
+      expect(update1).to receive(:apply)
+      manager.apply_all
+    end
+  end
+
+  describe "#all_signed?" do
+    let(:update0) { double("update0", signed?: true) }
```
lslezak: NP: Maybe use `signed_update` and `unsigned_update` instead of `update0` and `update1` to make it more readable.

src/lib/installation/clients/inst_update_installer.rb
lslezak: `applicable?` ?

src/lib/installation/driver_update.rb
imobach: You're right. I changed the behavior and I forgot to change the description.

test/updates_manager_test.rb
imobach: It was a leftover.

imobach:
@lslezak I've updated this branch with your suggestions and adding two new changes:

* Substitution of $ARCH in the self-update URL.
* Improvement of error messages when the signature checking fails.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,218 @@
...
+    #
+    # It substitutes $ARCH pattern with the architecture of the current system.
+    #
+    # @return [URI,nil] The string converted into a URL; nil if it's
+    #                   not a valid URL.
+    #
+    # @see URI.regexp
+    def get_url_from(url)
+      return nil unless url.is_a?(::String)
+      real_url = url.sub("\$ARCH", Arch.architecture)
```
lslezak: Notes:
- We should use `$arch` variable name (in lower case) to be compatible with the [libzypp expansion](https://github.com/openSUSE/libzypp/blob/master/zypp/repo/RepoVariables.cc#L493).
- `gsub` should be preferred to replace *all* variables (usually only one `$arch` will be used, but we should be on the safe side...)
- Word boundary should be checked, `url.sub("\$ARCH",..)` would also match 'foo/$ARCHITECTURE/bar' => `foo/x86_64ITECTURE/bar`. I.e. use a regexp argument.

=> use `url.gsub(/\$arch\b/,...)`

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,218 @@
...
+    end
+
+    # Ask the user if she/he wants to apply the update although it's not properly signed
+    #
+    # @return [Boolean] true if user answered 'Yes'; false otherwise.
+    def ask_insecure?
+      Popup.AnyQuestion(
+        Label::WarningMsg(),
+        signatures_error_message,
+        _("Yes"),
```
lslezak: Use [`Label.YesButton`](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Label.rb#L62) and [`Label.NoButton`](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Label.rb#L68) here.

Unfortunately we cannot use [`Popup.YesNoHeadline`](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Popup.rb#L580) here because it has `:focus_yes` by default...

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,218 @@
...
+    #
+    # @return [String] Error message
+    def signatures_error_message
+      # UpdatesManager support several updates to be applied. But this client
+      # does not take advantage of that feature yet.
+      update = updates_manager.updates.first
+      reason =
+        case update.signature_status
+        when :error
+          format(_("Installer update at %s can't be verified."), update.uri)
```
lslezak: Add a `TRANSLATORS:` comment and describe what the `%s` mean so the translators know that it will be replaced by an URL and they can use correct context for the translation.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,218 @@
...
+        when :missing
+          format(_("Installer update at %s is not signed."), update.uri)
```
lslezak: The same here.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,218 @@
...
+        end
```
lslezak: I think we should handle also the `else` case in the switch otherwise `reason` could be `nil` and `nil + "Using this update ..."` would cause a crash. Just to be on the safe side...

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -0,0 +1,218 @@
...
+      reason + _("\n\nUsing this update may put the integrity of your system at risk.\n" \
```
lslezak: NP: Using a string interpolation is IMHO a bit more translator friendly as they will know that the `\n\n` at the beginning are used as a separator and they will know the context of the message:

```ruby
# TRANSLATORS: Popup question, %s contains the details about the failed signature verification
_("%s\n\nUsing this update...")
```

src/lib/installation/driver_update.rb
```diff
@@ -0,0 +1,243 @@
...
+    end
+
+    # Run update.pre script
+    #
+    # @return [Boolean,nil] true if execution was successful; false if
+    #                       update script didn't exist.
+    # @raise DriverUpdate::CouldNotBeApplied
+    def run_update_pre
+      update_pre_path = local_path.join("install", "update.pre")
+      return false unless update_pre_path.exist? && update_pre_path.executable?
```
lslezak: Please add more logging, it will be quite difficult to debug it without details like what script was executed, what was the exit value.. etc.

lslezak:
LGTM :tada:

Thanks! :+1:

imobach:
Thanks to you for such a hard and long review :)

PR #348, opened 2016-03-11
Adding recover values behaviour (fate#319716)

teclator:
* I continued work done in #319 by @jreidinger moving code to src/lib and adding tests.
* The behaviour was probed also in yast/yast-registration#249

### DIFF
--- a/src/lib/installation/clients/inst_complex_welcome.rb
+++ b/src/lib/installation/clients/inst_complex_welcome.rb
@@ -19,136 +19,146 @@
 # current contact information at www.novell.com.
 # ------------------------------------------------------------------------------
 
+require "yaml"
+require "fileutils"
+require "yast"
+
 module Yast
   # This client shows main dialog for choosing the language,
   # keyboard and accepting the license.
   class InstComplexWelcomeClient < Client
+    include Yast::Logger
+    extend Yast::I18n
+
+    HEADING_TEXT = N_("Language, Keyboard and License Agreement")
+
     def main
-      Yast.import "UI"
-      Yast.import "Pkg"
+      Yast.import "Mode"
+
+      # bnc#206706
+      return :auto if Mode.autoinst
+
+      import_modules
+
       textdomain "installation"
 
-      Yast.import "Console"
-      Yast.import "GetInstArgs"
-      Yast.import "Keyboard"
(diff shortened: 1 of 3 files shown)


src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -19,136 +19,149 @@
...
-                )
-              )
-            )
-          )
-        ),
-        VWeight(1, VStretch())
-      )
+      UI.ChangeWidget(Id(:language), :Value, @language)
+
+      if Keyboard.user_decision == true
```
lslezak: Simple `if Keyboard.user_decision` is not sufficient?

src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -19,136 +19,149 @@
...
+        UI.ChangeWidget(Id(:keyboard), :Value, Keyboard.current_kbd)
+      else
+        @kbd = Keyboard.GetKeyboardForLanguage(@language, "english-us")
```
lslezak: `@kbd` is not used anywhere out side this block, simple `kbd` is sufficient.

src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -316,106 +226,223 @@ def AllLicensesDialog
...
+      change_language
+      setup_final_choice
+
+      ::FileUtils.rm_rf(DATA_PATH)
+    end
+
+    def text_mode?
+      return @text_mode unless @text_mode.nil?
+
+      @text_mode = Language.GetTextMode
```
lslezak: NP: `Language.GetTextMode` is basically just a [wrapper around `UI.GetDisplayInfo["TextMode"]`](https://github.com/yast/yast-country/blob/master/language/src/modules/Language.rb#L156) which can be simplified as `UI.TextMode`. (Unfortunately it is a boolean value so we cannot use the simple `@text_mode ||= UI.TextMode` trick here.)

I'd use `UI.TextMode` directly here.

lslezak:
Just some minor issues, otherwise LGTM.

lslezak:
LGTM :+1:

PR #349, opened 2016-03-14
Move file from url

imobach:
Moves `Transfer::FileFromUrl` from yast-update to yast-installation, as it's needed by #347. More information in https://github.com/yast/yast-update/pull/53 which moved it from AutoYaST2 to yast2-update.


mvidner:
LGTM

imobach:
Thanks!

PR #352, opened 2016-03-15
Try to update during installation as last step of system analysis (FATE#319716).

teclator:
- The update installer is responsible of not enter into a restart loop and
- This is part of yast/yast-installation#347 and yast/yast-installation#348


ancorgs:
lgtm

lslezak:
LGTM

imobach:
I've updated the PR with some information to be included in the blog (some screenshot is missing yet). Here is the old description:

- The update installer is responsible of not enter into a restart loop and
- This is part of yast/yast-installation#347 and yast/yast-installation#348

PR #354, opened 2016-03-15
Self update takes place before searching for system files

imobach:
Reviewing the workflow, it looks like the best moment to search for an update is right after USB, FireWire and hard disks are initialized:

* Probe USB devices
* Probe FireWire devices
* Probe floppy disk devices
* Probe hard disks controllers
* Load kernel modules for hard disk controllers
* Probe hard disks
* **Update Installer** <- USB, floppy and hard disks are initialized
* Search for system files
* Initialize software manager

There's no point in searching for system files or initializing the software manager if we need to update the installer later. So I would prefer to update the system before those tasks are executed.

### DIFF
--- a/src/lib/installation/clients/inst_system_analysis.rb
+++ b/src/lib/installation/clients/inst_system_analysis.rb
@@ -124,6 +124,11 @@ module Yast
         WFM.CallFunction("inst_features", [])
       end
 
+      # FATE #319716: Installer self-update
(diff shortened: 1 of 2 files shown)


imobach:
We're not merging this PR because we'll add a feature to replace the product version in the self-update URL and we'll need thet software manager to be initialized.

PR #355, opened 2016-03-16
Remove an extra popup during installer update

imobach:
Part of https://trello.com/c/LoAaDF97/468-8-sle-12-sp2-p1a-m-feature-319716-automatic-update-of-yast-at-beginning-of-the-installation

As it's shown as another step in the system analysis, I guess we should remove this popup (it's a leftover from the previous UI).


kwwii:
Yes, I think it should be removed.

mvidner:
LGTM.
I guess this does not need a .changes entry.

imobach:
@mvidner Yes, you're right, it does not deserve a .changes entry. Thanks!

PR #356, opened 2016-03-22
RPM-based self-update

imobach:
https://trello.com/c/T49hdvDd

This is a work in progress.


doc/SELF_UPDATE.md
```diff
@@ -14,32 +14,22 @@ Self-update is enabled by default. However, it can be disabled setting
...
-bandwidth and memory usage, all updates should be contained in a single DUD.
-Moreover, to make maintenance of this feature simpler, the only allowed archive
-format to use is `cpio` (default format for
-[mkdud](https://github.com/openSUSE/mkdud)).
-
-You can find more information about DUDs in [The Update Media
-HOWTO](http://ftp.suse.com/pub/people/hvogel/Update-Media-HOWTO/index.html) and
-in the [Linuxrc page](http://en.opensuse.org/SDB:Linuxrc#p_driverupdate]) in
-the openSUSE wiki.
+YaST will use RPM packages stored in a rpm-md repository.
```
mvidner: This sounds like a normal repo handled with zypp. But IIUC we will handle it in a simplified way and that should be documented here(?). I hope I got the idea right:

- all rpms in the repo are considered (no "patch" metadata, or "only those already installed")
- rpms are not "installed" with RPM and scripts. Just the file contents extracted
- no dependency checks are performed. RPMs are processed in ascii order of their filename

imobach: Yes, you got the idea right. I'll update the repo with your notes. Thanks!

src/lib/installation/update_repository.rb
```diff
@@ -0,0 +1,255 @@
...
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+require "yast"
+require "tempfile"
+require "pathname"
+
+module Installation
+  # Represents a update repository
```
mvidner: This repo has updates specific to the installation, right? The same thing as described in SELF_UPDATE.md: Update Format. We should say so otherwise someone will confuse this with a regular Zypp update.

src/lib/installation/update_repository.rb
```diff
@@ -0,0 +1,255 @@
...
+  #   repo.apply
+  #
+  # @example Fetching and applying to non-standard places
+  #   repo = UpdateRepository.new(URI("http://update.opensuse.org/42.1"))
+  #   repo.fetch(Pathname("/downloading"))
+  #   repo.apply(Pathname("/updates"))
+  class UpdateRepository
+    include Yast::Logger
+
+    attr_reader :uri, :repo_id, :paths, :instsys_parts_path
```
mvidner: I prefer explicit documentation of the attributes. Is this correct?
```rb
# @return [URI] URI of the repo
attr_reader :uri
# @return [Fixnum] Zypp ID of the repo
attr_reader :repo_id
# @return [Array<Pathname>] local paths of packages fetched from the repo
attr_reader :paths
# @return [Pathname]
attr_reader :instsys_parts_path
```
OK, `paths` may go away as we figure out the callback approach.

I don't know what `instsys.parts` is. It looks like an internal implementation detail. Maybe it should be a private attr?

src/lib/installation/update_repository.rb
```diff
@@ -0,0 +1,255 @@
...
+
+    class NotFound < StandardError; end
+    class CouldNotBeApplied < StandardError; end
+    class UpdatesNotFetched < StandardError; end
+    class CouldNotExtractUpdate < StandardError; end
+    class CouldNotMountUpdate < StandardError; end
+    class CouldNotSquashUpdate < StandardError; end
+
+    # Command to extract an RPM which is part of an update
+    EXTRACT_CMD = "rpm2cpio %<source>s | cpio --quiet --sparse -dimu --no-absolute-filenames"
```
mvidner: These strings are not part of the API so I think they should be moved close to the code that uses them.

src/lib/installation/update_repository.rb
```diff
@@ -0,0 +1,255 @@
...
+    # @return [Pathname] File name
+    def next_name(basedir, prefix: "yast_", length: 3)
+      files = Pathname.glob(basedir.join("*")).map(&:basename)
+      dirs = files.map(&:to_s).grep(/\A#{prefix}\d+\Z/)
+      dirs = dirs.map { |d| d.sub(prefix, "") } unless prefix.empty?
+      number = dirs.empty? ? 0 : dirs.map(&:to_i).max + 1
+      basedir.join(format("#{prefix}%0#{length}d", number))
+    end
+
+    # Register a mounted filesystem in instsys.parts file
```
mvidner: Please document why it is needed. IIUC yast does not use this file, linuxrc does. What would happen if we omitted this step?

mvidner:
Repo cleanup is here: https://github.com/mvidner/yast-installation/commit/dea5c94effcb7b8fbc6260125f3b0740986e58a5

src/lib/installation/update_repository.rb
```diff
@@ -0,0 +1,329 @@
...
+  # @example Fetching and applying to non-standard places
+  #   repo = UpdateRepository.new(URI("http://update.opensuse.org/42.1"))
+  #   repo.fetch(Pathname("/downloading"))
+  #   repo.apply(Pathname("/updates"))
+  class UpdateRepository
+    include Yast::Logger
+
+    # @return [URI] URI of the repository
+    attr_reader :uri
+    # @return [Fixnum] libzypp ID of the repository
```
lslezak: NP: just a technical detail: the repository IDs (Fixnum) are actually created by pkg-bindings, libzypp uses alias (String) as the identifier. (This is legacy from the pre-libzypp era.)

mvidner:
About the exceptions: we introduce quite a number of them, but AFAIK we only catch a subset in the new code. Is it intended that the rest fly away until they presumably pop up as an Internal Error?

A general note, I haven't looked in detail how it applies for this PR: I think it was http://exceptionalruby.com/ where I read that it is recommended to group exceptions according to how the user can react:

1. user input was invalid -> provide better input and retry
2. system transient failure (eg. a network repo did not respond) -> wait a bit and retry
3. system permanent failure -> too bad, report a bug (this can be further divided into: report a bug against this program / against some other library that we use)

src/lib/installation/update_repository.rb
```diff
@@ -0,0 +1,329 @@
...
+      if package_path.nil?
+        log.error("Package #{package} could not be retrieved.")
+        raise PackageNotFound
+      end
+      extract(Pathname(package_path), workdir)
+      squashed_path = next_name(dir, length: 3)
+      build_squashfs(workdir, squashed_path)
+      squashed_path
+    ensure
+      FileUtils.remove_entry(workdir)
```
lslezak: Use the block variant here, it automatically removes the created temp dir at the end. See http://ruby-doc.org/stdlib-2.2.0/libdoc/tmpdir/rdoc/Dir.html#method-c-tmpdir.

imobach:
@mvidner Yes, you're right and most of them occurs during packages fetch. I plan to improve the `fetch` method so, when `PackageNotFound`, `CouldNotExtractPackage` or `CouldNotSquashPackage` happens, the update will be rolled-back.

About `CouldNotMountUpdate` or `CouldNotBeApplied`, I think it's ok to let them go up as they look like a real bug. Moreover, if you have "applied" some update already, your system can be in a "unrecoverable" state (as `adddir` is not reversible).

The rest of the exceptions should be handled gracefully.

src/lib/installation/update_repository.rb
```diff
@@ -0,0 +1,329 @@
...
+    # Extract a RPM content to a given directory
+    #
+    # @param package_path [String]   RPM local path
+    # @param dir          [Pathname] Directory to extract the RPM contents
+    #
+    # @raise CouldNotExtractPackage
+    def extract(package_path, dir)
+      Dir.chdir(dir) do
+        cmd = format(EXTRACT_CMD, source: package_path)
+        out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), cmd)
```
lslezak: NP: Maybe we could simply use `` (backticks) here? Using SCR looks too heavyweight to me (we always extract to the inst-sys anyway thus do not need the chrooting feature....). But maybe it's just my personal taste...

src/lib/installation/update_repository.rb
```diff
@@ -0,0 +1,329 @@
...
+
+    # Mount the squashed filesystem containing updates
+    #
+    # @param path [Pathname] Mountpoint
+    #
+    # @raise CouldNotMountUpdate
+    #
+    # @see MOUNT_CMD
+    def mount_squashfs(file, mountpoint)
+      FileUtils.mkdir(mountpoint) unless mountpoint.exist?
```
lslezak: I'd use `FileUtils.mkdir_p` here to not fail when the parent dir is missing. Just to be safe...

src/lib/installation/updates_manager.rb
```diff
@@ -15,79 +15,70 @@
...
-    rescue Installation::DriverUpdate::NotFound
-      log.error("Driver updated at #{uri} could not be found")
-      false
+    # @see Installation::UpdateRepository
+    def add_repository(uri)
+      new_repository = Installation::UpdateRepository.new(uri)
+      new_repository.fetch
+      @repositories << new_repository
+      :ok
+    rescue Installation::UpdateRepository::ValidRepoNotFound
```
lslezak: I'm not sure about this construct. Here you convert the exceptions to symbols, then you [convert the symbol to boolean](https://github.com/yast/yast-installation/pull/356/files#diff-8aa15970732082407af8655c128f0913R159) at the caller.

I'd probably move the exception handling directly to the caller (`fetch_update`) and return the boolean value in the exception handler (without the need for the intermediate symbol).

Or did I overlook something?

src/lib/installation/update_repository.rb
```diff
@@ -0,0 +1,329 @@
...
+    #
+    # @param basedir [Pathname] Directory
+    # @param prefix  [String]   Name prefix
+    # @param length  [Integer]  Name's length
+    # @return [Pathname] File name
+    def next_name(basedir, prefix: "yast_", length: 3)
+      files = Pathname.glob(basedir.join("*")).map(&:basename)
+      dirs = files.map(&:to_s).grep(/\A#{prefix}\d+\Z/)
+      dirs = dirs.map { |d| d.sub(prefix, "") } unless prefix.empty?
+      number = dirs.empty? ? 0 : dirs.map(&:to_i).max + 1
```
lslezak: Um, looks too much complicated IMHO. I think we could use simple `files.size + 1` approach to find the next name. The files will be created only by YaST and we can guarantee proper ascending sequence without gaps or other issues.

But if this works for you and you have unit tests I'm fine with it :wink:

src/lib/installation/driver_update.rb
```diff
@@ -24,176 +22,58 @@ module Installation
...
+      # Find driver updates in a given directory
+      #
+      # A directory with a `dud.config` file will be considered a driver
+      # update.
+      #
+      # @param dir [Pathname] Directory to search for driver updates
+      # @return [Array<DriverUpdate>] Found driver updates
+      def find(dir)
+        log.info("Searching for Driver Updates at #{dir}")
+        Pathname.glob("#{dir}/*/dud.config").map do |path|
```
lslezak: Um, where this `dud.config` file comes from? What's the purpose for it?

imobach: It comes from the driver update file itself. When it's uncompressed, the file is there.

src/lib/installation/update_repository.rb
imobach: Yes, that's a fair point. In the past (well, during the last sprint ;) ) we share the space with other driver updates. But after moving to `yast_` "namespace" it's true that it could be a lot easier :)

src/lib/installation/updates_manager.rb
imobach: Well, it also looked weird to me after a while. I'm changing the way in which exceptions are handled.

src/lib/installation/update_repository.rb
imobach: Yes, it looks overkill, but I was doing in that way for consistence reasons. But it could change it.

src/lib/installation/update_repository.rb
imobach: It also was a little bit confusing for me when reading yast-pkg-bindings and libzypp code :)

src/lib/installation/driver_update.rb
lslezak: OK, but here we are switching to RPMs. IIUC the RPMs will be the usual YaST RPM packages, the same ones released as regular online updates.

In theory you could use YaST:Head repository from OBS to update the installer completely to the latest bleeding edge...

imobach: Yes, but that code belongs to `DriverUpdate` class, which is intended to find local driver updates and apply them of top of repository updates. See https://trello.com/c/cIfOtGM8. So I guess that some extra documentation is needed :)

lslezak: I see... Yes, document that please :sweat_smile:

src/lib/installation/update_repository.rb
lslezak: You have already implemented it this way and you can leave it, I'm fine with that.

Personally I would just prefer clean Ruby code (as external Rubyists would write), possibly without any YaST legacy. Even though it might look inconsistent in some cases.

teclator: Should be nice to know the better way to execute system commands, as use Ruby standard library or use Yast calls.

And maybe update the guide about it:

http://yastgithubio.readthedocs.org/en/latest/architecture/#system-configuration-repository-scr

So in general, to execute any command without agents being involved is preferable to use the ruby systems calls  or Cheetah.  Or could we consider other exceptions?

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -22,9 +22,7 @@ class InstUpdateInstaller
     include Yast::I18n
 
     UPDATED_FLAG_FILENAME = "installer_updated"
-    UPDATES_PATH = Pathname.new("/update")
-    KEYRING_PATH = Pathname.new("/installkey.gpg")
-    GPG_HOMEDIR  = Pathname.new("/root/.gnupg")
+    REMOTE_SCHEMES = ["http", "https", "ftp", "nfs", "cifs", "smb"]
```
lslezak: Libzypp also supports `nfs4` schema, see the *Supported URI formats* section in `man zypper`.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -205,47 +212,11 @@ def try_to_update?
...
-    # * All updates are signed.
-    # * We're running in insecure mode (so we don't need them to be signed).
-    # * The user requests to install it although is not signed.
-    #
-    # @return [Boolean] true if it should be applied; false otherwise.
-    def applicable?
-      updates_manager.all_signed? || insecure_mode? || ask_insecure?
-    end
-
     # Determines whether the given URL is equals to the default one
```
lslezak: NP: Typo,  `is equal to`.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -138,34 +139,60 @@ def update_flag_file
...
-      log.info("Applying installer updates")
-      updates_manager.apply_all
-      true
+    # Ask the user about checking network configuration. If she/he accepts,
+    # the `inst_lan` client will be launched.
+    #
+    # @return [Boolean] true if the network configuration client was launched;
+    #                   false if the network is not configured.
+    def configure_network?
+      if Popup.YesNo(_("There was a problem reading the updates repository. "\
```
lslezak: NTH: I'd add the repo URL to the message so the users knows what could not be have been read. That's especially important for custom URLs, maybe there is typo there...

src/lib/installation/update_repository.rb
```diff
@@ -24,45 +24,66 @@ module Installation
   #   repo = UpdateRepository.new(URI("http://update.opensuse.org/42.1"))
   #   repo.fetch
   #   repo.apply
+  #   repo.cleanup
   #
   # @example Fetching and applying to non-standard places
   #   repo = UpdateRepository.new(URI("http://update.opensuse.org/42.1"))
   #   repo.fetch(Pathname("/downloading"))
   #   repo.apply(Pathname("/updates"))
+  #   repo.cleanup
```
lslezak: NP: The `apply` call can raise an exception, the proper approach would be to put the `cleanup` call into an `ensure` block I guess...

src/lib/installation/update_repository.rb
```diff
@@ -0,0 +1,366 @@
...
+  #   repo = UpdateRepository.new(URI("http://update.opensuse.org/42.1"))
+  #   repo.fetch
+  #   repo.apply
+  #   repo.cleanup
+  #
+  # @example Fetching and applying to non-standard places
+  #   repo = UpdateRepository.new(URI("http://update.opensuse.org/42.1"))
+  #   repo.fetch(Pathname("/downloading"))
+  #   repo.apply(Pathname("/updates"))
+  #   repo.cleanup
```
lslezak: NP: The `apply` call can raise an exception, the proper approach would be to put the `cleanup` call into an `ensure` block I guess...

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -139,53 +137,62 @@ def update_flag_file
...
     def update_installer
-      fetch_update ? apply_update : false
+      log.info("Adding update from #{self_update_url}")
+      updates_manager.add_repository(self_update_url)
+      log.info("Applying installer updates")
+      updates_manager.apply_all
+      true
+
+    rescue ::Installation::UpdatesManager::ValidRepoNotFound
+      Report.Error(_("A valid update could not be found at #{self_update_url}")) unless using_default_url?
```
imobach: The call to `_()` is not translator friendly :P Need to fix it.

src/lib/installation/clients/inst_update_installer.rb
imobach: That's true. And also `obs`.

src/lib/installation/update_repository.rb
lslezak: My opinion: For simple cases I'd prefer Ruby system calls, for more complex Yast::Execute.

But this sounds more like a topic for discussion at yast-devel@opensuse.org...

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -22,9 +22,7 @@ class InstUpdateInstaller
     include Yast::I18n
 
     UPDATED_FLAG_FILENAME = "installer_updated"
-    UPDATES_PATH = Pathname.new("/update")
-    KEYRING_PATH = Pathname.new("/installkey.gpg")
-    GPG_HOMEDIR  = Pathname.new("/root/.gnupg")
+    REMOTE_SCHEMES = ["http", "https", "ftp", "nfs", "nfs4", "cifs", "smb", "obs"]
```
lslezak: Oh, thanks for checking the man page :wink:, but it's a bit tricky. The `obs` schema is a meta schema supported only by zypper (not by libzypp). If you add an `obs://` repository `zypper` will convert it to the usual `http://` URL and use it for the repository.

See the [remote schema list defined by libzypp](https://github.com/openSUSE/libzypp/blob/a7a038aeda1ad6d9e441e7d3755612aa83320dce/zypp/Url.cc#L464).

```
# zypper ar -f obs://zypp:Head/ zypp
# zypper lr zypp | grep URI
URI            : http://download.opensuse.org/repositories/zypp:/Head/openSUSE_Leap_42.1
```

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -186,8 +188,9 @@ def remote_self_update_url?
     # @return [Boolean] true if the network configuration client was launched;
     #                   false if the network is not configured.
     def configure_network?
-      if Popup.YesNo(_("There was a problem reading the updates repository. "\
-                       "Would you like to check your network configuration?"))
+      if Popup.YesNo(
+        format(_("There was a problem reading the updates repository at %s.\n\n"\
```
lslezak: The URL is potentially long, the line might not fit in ncurses. Just put a new line there:

`... the updates repository at\n%s.\n\n`

And add a `TRANSLATORS:` comment describing what the `%s` means (that it's an URL).

src/lib/installation/clients/inst_update_installer.rb
imobach: Oh, well, then `sftp` and `tftp` are missing :) thanks for the explanation. The code is the _source of truth_ :)

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -139,53 +137,65 @@ def update_flag_file
...
-      updates_manager.apply_all
-      true
+    # Ask the user about checking network configuration. If she/he accepts,
+    # the `inst_lan` client will be launched.
+    #
+    # @return [Boolean] true if the network configuration client was launched;
+    #                   false if the network is not configured.
+    def configure_network?
+      if Popup.YesNo(
+        format(_("There was a problem reading the updates repository at %s.\n\n"\
```
lslezak: The URL is potentially long, the line might not fit in ncurses. Just put a new line there:

`... the updates repository at\n%s.\n\n`

And add a `TRANSLATORS:` comment describing what the `%s` means (that it's an URL).

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -22,7 +22,7 @@ class InstUpdateInstaller
     include Yast::I18n
 
     UPDATED_FLAG_FILENAME = "installer_updated"
-    REMOTE_SCHEMES = ["http", "https", "ftp", "nfs", "nfs4", "cifs", "smb", "obs"]
+    REMOTE_SCHEMES = ["http", "https", "ftp", "ftp", "sftp", "nfs", "nfs4", "cifs", "smb"]
```
lslezak: If I can see correctly `ftp` is there twice ...

imobach: Oh, it was `tftp`... fixing.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -189,7 +191,9 @@ def remote_self_update_url?
     #                   false if the network is not configured.
     def configure_network?
       if Popup.YesNo(
-        format(_("There was a problem reading the updates repository at %s.\n\n"\
+        # TRANSLATORS: %s is an URL
+        format(_("Your network is not working properly so installer updates "\
```
lslezak: NP: I'm not sure about the wording, what about this?

```ruby
"Downloading updates from\n%s\nfailed.\n\n" \
"Would you like to check your network configuration?"
```

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -22,9 +22,7 @@ class InstUpdateInstaller
     include Yast::I18n
 
     UPDATED_FLAG_FILENAME = "installer_updated"
-    UPDATES_PATH = Pathname.new("/update")
-    KEYRING_PATH = Pathname.new("/installkey.gpg")
-    GPG_HOMEDIR  = Pathname.new("/root/.gnupg")
+    REMOTE_SCHEMES = ["http", "https", "ftp", "ftp", "sftp", "nfs", "nfs4", "cifs", "smb"]
```
lslezak: If I can see correctly `ftp` is there twice ...

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -139,53 +137,69 @@ def update_flag_file
...
-      true
+    # Ask the user about checking network configuration. If she/he accepts,
+    # the `inst_lan` client will be launched.
+    #
+    # @return [Boolean] true if the network configuration client was launched;
+    #                   false if the network is not configured.
+    def configure_network?
+      if Popup.YesNo(
+        # TRANSLATORS: %s is an URL
+        format(_("Your network is not working properly so installer updates "\
```
lslezak: NP: I'm not sure about the wording, what about this?

```ruby
"Downloading updates from\n%s\nfailed.\n\n" \
"Would you like to check your network configuration?"
```

lslezak:
:exclamation: LGTM :exclamation: :+1:

lslezak:
LGTM, thanks!

PR #357, opened 2016-03-23
Move proc_modules.scr to yast-yast2 (bsc#972310)

cwh42:
This is to avoid yast2-sound to depend on yast2-installation.
Related PR: https://github.com/yast/yast-yast2/pull/453


shundhammer:
LGTM

PR #358, opened 2016-03-24
Fixed spec file

lslezak:
This is a fix up for the previous PR #357, the `proc_modules.scr` file has been moved to yast2 but the spec file was not updated.

Unfortunately not found by Travis as we do not build the package there, found by Jenkins later...


jreidinger:
LGTM

PR #359, opened 2016-04-06
Unifying network button position in activation of disks client. (FATE#319716)

teclator:
I had unified the distribution of `[Network Configuration]` button.

**Previous**
diskactivation

**Now**
diskactivationnet

### DIFF
--- a/src/lib/installation/clients/inst_disks_activate.rb
+++ b/src/lib/installation/clients/inst_disks_activate.rb
@@ -170,7 +170,7 @@ module Yast
   private
 
     def network_button
-      Right(PushButton(Id(:network), _("Network Configuration...")))
+      Right(PushButton(Id(:network), _("Net&work Configuration...")))
     end
 
     def show_base_dialog
--- a/src/lib/installation/clients/inst_disks_activate.rb
+++ b/src/lib/installation/clients/inst_disks_activate.rb
@@ -100,19 +100,20 @@ module Yast
                     missing_part
                   end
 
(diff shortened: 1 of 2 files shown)


kobliha:
It would be nice to use a shared functionality from, e.g. yast2.rpm or something like that ~ because that would make it better for changes in the future, but it does not make the situation worse now. And it has a test case! Thanks!

LGTM (not tested, hope it works :) )

imobach:
Better now. LGTM.

dzedro:
fine for openQA if keyboard shortcuts don't disappear or change :+1: 
no keyboard shortcuts is bad, test will fail :sob: 
different keyboard shortcuts can be changed in openQA too :construction_worker:

imobach:
@dzedro Thanks for pointing it out. @teclator According to screenshots, it looks like the shortcut for network configuration has changed. But, to be honest, I'm not sure if this scenario is tested in openQA.

PS: anyway, it's good to have it present for the future.

dzedro:
e.g. iSCSI test is using [disk activation](https://openqa.suse.de/tests/333427/modules/iscsi_configuration/steps/1)
openQA is matching specific screen areas and this change will not affect existing needles

teclator:
@dzedro yeah, i'm fixing it, i just unified labels text as it was in Registration but i removed the previous shortcut with a dynamic one, so i will unify buttons and fix it. Thnx for your help and time.

imobach:
LGTM

PR #362, opened 2016-04-14
Support for self update in AY mode

lslezak:
Extends #352, #356 also to AutoYaST mode.

See [FATE#319716](https://fate.suse.com/319716).

The AY profile schema is updated in https://github.com/yast/yast-autoinstallation/pull/206.

### DIFF
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -34,6 +34,7 @@ module Yast
     Yast.import "Popup"
     Yast.import "Report"
     Yast.import "NetworkService"
+    Yast.import "Mode"
 
     def main
       textdomain "installation"
@@ -82,8 +83,10 @@ module Yast
     #
     # @see #self_update_url_from_linuxrc
     # @see #self_update_url_from_control
+    # @see #self_update_url_from_profile
     def self_update_url
-      url = self_update_url_from_linuxrc || self_update_url_from_control
+      url = self_update_url_from_linuxrc || self_update_url_from_profile ||
+        self_update_url_from_control
       log.info("self-update URL is #{url}")
       url
     end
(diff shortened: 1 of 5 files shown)


src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -83,7 +84,8 @@ def self_update_enabled?
     # @see #self_update_url_from_linuxrc
     # @see #self_update_url_from_control
```
imobach: np: @see #self_update_url_from_profile

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -101,6 +103,19 @@ def self_update_url_from_control
       get_url_from(ProductFeatures.GetStringFeature("globals", "self_update_url"))
     end
 
+    # Return the self-update URL from the AutoYaST profile
+    # @return [URI,nil] the self-update URL, nil if not running in AutoYaST mode
+    #   or when the URL is not defined in the profile
+    def self_update_url_from_profile
+      return nil unless Mode.auto
+
+      Yast.import "Profile"
```
imobach: Why not move this to `main`?

lslezak: The reason is that the `Mode` is always needed (to know whether we are running in AY or not), but `Profile` is needed only in AY mode, in normal installation it's not needed.

(Maybe the `Profile` is loaded by other code anyway and this small optimization is useless, but at least I tried... :wink:)

test/test_helper.rb
```diff
@@ -11,10 +11,20 @@ module Yast
   class AutoinstConfigClass
     # we need at least one non-default methods, otherwise ruby-bindings thinks
     # it is just namespace
-    def fake_method
+    def cio_ignore
```
imobach: Uh, I didn't know about this.

imobach:
LGTM

PR #361, opened 2016-04-13
Improved network interfaces listing

mchf:
bnc#956473

to be merged once approved by UI/UX expert

for screenshot see https://trello-attachments.s3.amazonaws.com/56573284af24cf7c26f71d78/640x480/2935c69eb6c43fd4d308877b1d40e208/sle12_ssh_inst_2nd_stage_details.png

PR #363, opened 2016-04-15
Improved network interfaces listing (SLE-12-GA)

mchf:
https://github.com/yast/yast-installation/pull/361

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Apr 13 07:14:09 UTC 2016 - mfilka@suse.com
+
+- bsc#956473
+  - improved formatting of network interfaces listing
+- 3.1.116.10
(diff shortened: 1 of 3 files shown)


kobliha:
If it is the same as in master, then it's definitely fine. I've checked that you use correct versioning.

so LGTM

mchf:
well no ... this is "bottom" MU .. others should get into corresponding branches as merges of this one (SLE-12-GA merges into SLE-12-SP1 and it merges into master)

mchf:
I've created https://github.com/yast/yast-installation/pull/361 by accident for SP1 ... it will be replaced when this PR is merged.

kobliha:
Yes, OK, but the code is the same everywhere. The only difference will be in changes file (and version)

So LGTM still applies

mchf:
I just wanted to make it clear for record. Thanks for review.

PR #364, opened 2016-04-15
Improved network interfaces listing (SLE-12-SP2)

mchf:
See https://github.com/yast/yast-installation/pull/363 and https://github.com/yast/yast-installation/pull/361


kobliha:
When Travis is fine, then LGTM
Thanks!

mchf:
Thanks for patience and reviews.

PR #365, opened 2016-04-18
Document the update repo URL evaluation during self-update

lslezak:
- includes some minor formatting/wording fixes

### DIFF
--- a/doc/SELF_UPDATE.md
+++ b/doc/SELF_UPDATE.md
@@ -1,18 +1,18 @@
-# Installer self-update
+# Installer Self-update
 
 Starting on version 3.1.175, yast2-install is able to update itself during
 system installation. This feature will help to solve problems with the
 installation even after the media has been released. Check
 [FATE#319716](https://fate.suse.com/319716) for a more in-depth rationale.
 
-## Disabling updates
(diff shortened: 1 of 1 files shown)


doc/SELF_UPDATE.md
```diff
@@ -1,18 +1,18 @@
-# Installer self-update
+# Installer Self-update
```
imobach: mental note: I have to get use to upper-case letters in titles (in spanish is just wrong). :)

doc/SELF_UPDATE.md
```diff
@@ -31,31 +31,43 @@ handled in a different way:
...
+   ```
+3. Hard-coded in the `control.xml` file on the installation medium (thus it
+   depends on the base product):
+   ```xml
+   <globals>
+     <self_update_url>http://updates.suse.com/sle12/$arch</self_update_url>
+   </globals>
+   ```
+
+The first found option is used. If the base product does not define the URL
```
imobach: It can be confusing as it is. If no URL is found (1, 2 o 3), the self update is skipped (not only if the base product does not define one).

imobach:
LGTM

PR #367, opened 2016-04-20
Skipped disk activation step in case of installer updated (bsc#974409)

teclator:
During the tests i realized that the license agreement was not remembered so i also fix it.

### DIFF
--- a/src/lib/installation/clients/inst_disks_activate.rb
+++ b/src/lib/installation/clients/inst_disks_activate.rb
@@ -26,6 +26,9 @@
 #
 # $Id$
 #
+
+require "installation/clients/inst_update_installer"
+
 module Yast
   class InstDisksActivateClient < Client
     def main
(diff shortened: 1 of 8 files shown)


src/lib/installation/clients/inst_disks_activate.rb
```diff
@@ -55,6 +58,10 @@ def main
       @have_zfcp = false
       @want_fcoe = false
 
+      @installer_update ||= InstUpdateInstaller.new
```
ancorgs: Weird use of a client. Is strange to have a client (something that is usually only meant to be executed) with the responsibility of knowing if the installer is updated.

ancorgs: A better solution discussed offline

teclator: Nice discussion ;)

teclator:
This PR's are related to this one

yast/yast-yast2#463
yast/yast-packager#155

src/lib/installation/clients/inst_disks_activate.rb
```diff
@@ -26,6 +26,9 @@
 #
 # $Id$
 #
+
+require "installation/clients/inst_update_installer"
```
ancorgs: Is this needed?

ancorgs:
Besides the surplus require, looks good.

PR #372, opened 2016-05-06
get more texts for roles dialog from control file, allow a general label (bsc#974625)

jsrain:
Do not merge yet, needs to be done in sync with https://github.com/yast/skelcd-control-SLES/pull/58

### DIFF
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -875,6 +875,24 @@ Example:
     </system_roles>
 
     <texts>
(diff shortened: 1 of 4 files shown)


mvidner:
Please document the additional translatable strings in https://github.com/yast/yast-installation/blob/master/doc/control-file.md#system-roles .

jsrain:
Just added

mvidner:
LGTM

PR #375, opened 2016-05-13
FATE#320262

imobach:
* Implements fate#320262
* Manually tested faking `Arch.is_zkvm` method.

### DIFF
--- a/src/lib/installation/clients/inst_finish.rb
+++ b/src/lib/installation/clients/inst_finish.rb
@@ -516,7 +516,7 @@ module Yast
       # and display it, if necessary
 
       # Do not call any SCR, it's already closed!
-      if Ops.greater_than(Builtins.size(Misc.boot_msg), 0) && !Mode.autoinst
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/inst_finish.rb
```diff
@@ -516,7 +516,7 @@ def main
       # and display it, if necessary
 
       # Do not call any SCR, it's already closed!
-      if Ops.greater_than(Builtins.size(Misc.boot_msg), 0) && !Mode.autoinst
+      if Ops.greater_than(Builtins.size(Misc.boot_msg), 0)
```
imobach: I'm not sure about this change: we must shown the warning about reIPL and it will be skipped in AutoYaST. From my POV, this message should have the same behavior than other messages (if reporting is disabled, it will be skipped).

imobach:
Finally it looks like we'll use a different approach. So I'm closing this PR. Check out [FATE#320262](https://fate.suse.com/320262) for further details.

PR #379, opened 2016-05-23
Ruby debugger integration (FATE#318421)

lslezak:
Optionally start the Ruby debugger at the beginning of installation. The debugger is started using `Y2DEBUGGER` boot option.

*Note: the debugger invocation is implemented here: https://github.com/yast/yast-ruby-bindings/pull/167*

The supported modes are:

- `Y2DEBUGGER=1` - enable the debugger, in Qt UI the debugger frontend is started automatically
- `Y2DEBUGGER=remote` - make the debugger remotely accessible, connect using `byebug -R <host>:3344` command (in the other cases the debugger can be accessed *only localy*)
- `Y2DEBUGGER=manual` - like `1`, but does not start the the debugger frontend automatically

### Screenshots

(Click a screenshot to see the full size image.)

- GUI with `Y2DEBUGGER=1` - the debugger console is opened automatically
  debugger_qt_default

- Text mode with `Y2DEBUGGER=1` - the debugger console must be connected manually (you have to switch to a different virtual terminal and run the debugger frontend there)
  debugger_ncurses_default

- GUI with `Y2DEBUGGER=remote` - it displays the machine IP address in the `byebug` command to easily connect to the machine (the same dialog is displayed in the text mode)
  debugger_qt_remote

- GUI with `Y2DEBUGGER=manual` - behaves like the default in the text mode
  debugger_qt_manual

### DIFF
--- a/src/lib/installation/clients/installation.rb
+++ b/src/lib/installation/clients/installation.rb
@@ -30,6 +30,8 @@
 # $Id$
 module Yast
   class InstallationClient < Client
+    include Yast::Logger
+
     def main
       textdomain "installation"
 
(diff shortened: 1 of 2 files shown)


src/lib/installation/clients/installation.rb
```diff
@@ -30,19 +30,36 @@
...
       Yast.import "Hooks"
+      Yast.import "Linuxrc"
 
       Hooks.search_path.join!("installation")
 
       # Initialize the UI
       UI.SetProductLogo(true)
       Wizard.OpenLeftTitleNextBackDialog
+
+      # start the debugger if required (FATE#318421)
```
imobach: What about moving this piece of code to a separate method? :)

src/lib/installation/clients/installation.rb
```diff
@@ -30,19 +30,36 @@
...
+      if (Linuxrc.InstallInf("Cmdline") || "").match(/\bY2DEBUGGER=(.*)\b/i)
```
imobach: Does it worth adding support to Linuxrc (so you get this argument written to `install.inf`)? Anyway, I'm fine as it's right now.

lslezak: KISS - this is the only place in YaST which uses the value. I'd keep it simple and avoid modifying linuxrc just because of this single place.

If it turns out we need to handle it also some where else we can add the linuxrc support later.

imobach:
LGTM

PR #380, opened 2016-05-25
[Review] Request from 'schubi2' @ 'yast/yast-installation/review_160524_autoyast_support'

schubi2:
  * 24c0278 packaging
  * 7571cdc fixes in exporting settings
  * 0f24f90 setting global variables corretly
  * 8f080a6 typo
  * c5f7936 typo
  * 9ed0d1f fixed makefile pathes
  * 3817ad0 fixed errors
  * af78a0e syntax
  * d46350c autoyast support


src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -0,0 +1,70 @@
...
+      ret
+    end
+
+    def import(data)
+      ::Installation::SshImporter.instance.import(data)
+      true
+    end
+
+    def summary
+      "<UL>" + ::Installation::SshImporter.instance.summaray + "</UL>"
```
imobach: typo: `summaray`

src/lib/installation/dialogs/ssh_import.rb
```diff
@@ -67,7 +67,7 @@ def dialog_content
           CheckBoxFrame(
             Id(:import_ssh_key),
             _("I would like to import SSH keys from a previous installation"),
-            true,
+            (!importer.device.nil? && !importer.device.empty?),
```
imobach: Maybe you would simplify it:

```ruby
importer.device && !importer.device.empty?
```

Or:

```
!(importer.device.nil? || importer.device.empty?)
```

Anyway, I'm fine with the current code.

imobach:
Tests are failing.

autoyast_desktop/ssh_import.desktop
```diff
@@ -0,0 +1,26 @@
...
+X-SuSE-YaST-SortKey=
+X-SuSE-YaST-AutoInstResource=
+X-SuSE-YaST-AutoInstPath=install
+X-SuSE-YaST-AutoInstSchema=ssh_import.rnc
+X-SuSE-YaST-AutoInstClonable=true
+
+Icon=yast-ssh_import
+Exec=
+
+Name=SHH Key Import
```
jreidinger: Typo SHH instead of SSH

src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -0,0 +1,171 @@
...
+require "installation/auto_client"
+require "installation/ssh_importer"
+
+Yast.import "Progress"
+Yast.import "Mode"
+Yast.import "Popup"
+
+module Installation
+  # AutoYaST client for ssh_import
+  class SSHImportAutoClient < ::Installation::AutoClient
```
jreidinger: :+1:

autoyast_desktop/ssh_import.desktop
```diff
@@ -0,0 +1,26 @@
...
+GenericName=Importing SSH keys from a previous installtion
```
lslezak: Typo, should be `installation`.

src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -0,0 +1,171 @@
...
+    # <ssh_import>
+    #   <import config:type="boolean">true</import>
+    #   <config config:type="boolean">true</config>
+    #   <device>/dev/sda4</device>
+    # </ssh_import>
+    def import(data)
+      if data["import"]
+        log.info "Importing AutoYaST data: #{data}"
+        ssh_importer.copy_config = data["config"] == true
+        if data["device"] && !data["device"].empty?
```
jreidinger: so if device is not specified, it do nothing?

src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -0,0 +1,171 @@
...
+            Yast::Report.Warning(
+              # TRANSLATORS: %s is the device name like /dev/sda0
+              _("Device %s not found. Taking default entry.") %
+              data["device"]
+              )
+          end
+        end
+      else
+        ssh_importer.device = nil # do not copy ssh keys into the installed system
+      end
```
jreidinger: In general I think it should be a bit simplified work-flow like
```
if !data["import"]
  ssh_importer.device = nil # do not copy ssh keys into the installed system
  return true
end

# continue with rest of code without indentation
```

src/lib/installation/clients/ssh_import_proposal.rb
```diff
@@ -37,24 +37,9 @@ def importer
...
-          res = _("SSH host keys and configuration will be copied from %s") % partition
-        else
-          # TRANSLATORS: %s is the name of a Linux system found in the hard
-          # disk, like 'openSUSE 13.2'
-          res = _("SSH host keys will be copied from %s") % partition
-        end
-      end
+      res = importer.summary
+      # TRANSLATORS: link to change the proposal
+      res += " " + _("(<a href=%s>change</a>)") % '"ssh_import"'
```
lslezak: Um, IIRC in installation we removed that `change` link, see my comment here: https://github.com/yast/yast-installation/pull/382#issuecomment-221648621

src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -0,0 +1,171 @@
...
+      ret
+    end
+
+    # Writes the SSH keys from the selected device (and also other
+    # configuration files if #copy_config? is true) in the target
+    # filesystem
+    def write
+      if Mode.config
+        Popup.Notify _("It makes no sense to write these settings to system.")
```
jreidinger: does it make sense to show such popup?

schubi2: I think yes, because then we would get complains from the user that nothing has happened. :-)

jreidinger: Mode config means, that user configure its new autoyast profile. How he can complain if it will be part of that autoyast profile?

imobach: Because in config mode, we won't import anything as I'm not mistaken. Maybe we could do it (as it does not make much sense from my POV). We could just log the message.

src/lib/installation/clients/ssh_import_proposal.rb
```diff
@@ -37,24 +37,9 @@ def importer
...
-          # TRANSLATORS: %s is the name of a Linux system found in the hard
-          # disk, like 'openSUSE 13.2'
-          res = _("SSH host keys and configuration will be copied from %s") % partition
-        else
-          # TRANSLATORS: %s is the name of a Linux system found in the hard
-          # disk, like 'openSUSE 13.2'
-          res = _("SSH host keys will be copied from %s") % partition
-        end
-      end
+      res = importer.summary
```
imobach: Importer does not have a `#summary` method. I think we should revert `preformatted_proposal` to the [original version](https://github.com/yast/yast-installation/blob/master/src/lib/installation/clients/ssh_import_proposal.rb#L39).

schubi2: I think that has been the reason for I have put it into src/lib/installation/ssh_importer.rb
in my first attempt:
https://github.com/yast/yast-installation/pull/380/commits/d46350ce6a03a700b6507dc4be2b95b314dfaea6
I would not double the code here. Just a thought :-)

imobach: I agree we should not duplicate the code, you're right. I've added a new class to be responsible for formatting the `SshImporter` summary in #387.

package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jun  2 11:09:33 UTC 2016 - schubi@suse.de
+
+- AutoYaST support for ssh_import module.
```
lslezak: NP: please write more end-user oriented entry, "ssh_import module" is not cryptic, but still the AutoYaST users might not fully understand. I think we should describe the added feature using the end-user language...

src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -0,0 +1,155 @@
...
+    # </ssh_import>
+    #
+    # @param data [Hash] AutoYaST specification.
+    # @option data [Boolean] :import Import SSH keys
+    # @option data [Boolean] :config Import SSH server configuration
+    #   in addition to keys.
+    # @option data [Boolean] :device Device to import the keys/configuration from.
+    def import(data)
+      if !data["import"]
+        log.info("Do not import ssh keys/configuration")
```
lslezak: NP: for logging the convention is to not use the parenthesis, simple `log.info "message"` should be preferred.

src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -0,0 +1,155 @@
...
+
+      log.info "Importing AutoYaST data: #{data}"
+      ssh_importer.copy_config = data["config"] == true
+      if data["device"] && !data["device"].empty?
+        if ssh_importer.configurations.key?(data["device"])
+          ssh_importer.device = data["device"]
+        else
+          Yast::Report.Warning(
+            # TRANSLATORS: %s is the device name like /dev/sda0
+            _("Device %s not found. Taking default entry.") %
```
lslezak: What's the *default entry*? Is it known? Could we print it just like the not found device? For me it's a bit unclear...

src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -0,0 +1,155 @@
...
+      # depends on the installed system we would like to
+      # have this section in the exported AutoYaST file
+      # regardless if the entries have been changed nor not.
+      modified
+
+      begin
+        args = {
+          "enable_back" => true,
+          "enable_next" => false,
+          "going_back"  => true
```
lslezak: *going_back* is used in the workflow to indicate that the user pressed [Back] and is going the opposite direction in the workflow. What's the reason to have `true` here? I'd expect `false`, this looks a bit strange to me...

If it is OK is deserves a comment at least.

src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -0,0 +1,155 @@
...
+
+    # Exporting data to the AutoYaST configuration module.
+    # That's are default entries.
+    def export
+      ret = {}
+      if Mode.config
+        # Taking values from AutoYast configuration module
+        if ssh_importer.device && !ssh_importer.device.empty?
+          ret["import"] = true
+          ret["config"] = ssh_importer.copy_config
```
lslezak: Maybe use `copy_config` key to be more clear what does it mean? And to be compatible with the called method...

src/lib/installation/ssh_config.rb
```diff
@@ -136,8 +137,17 @@ def write_files(root_dir, write_keys: true, write_config_files: true)
...
+        log.info "Writing SSH keys to the target system: "
+        keys.each do |k|
+          log.info "#{k}"
+          k.write_files(dir)
+        end
+      end
+      config_files.each do |f|
+        log.info "Writing SSH configuration to the target system: #{f}"
+        f.write(dir)
+      end if write_config_files
```
lslezak: Do not use trailing `if` with a multi line block, it's harder to read and you can easily overlook it.

src/lib/installation/ssh_importer_presenter.rb
```diff
@@ -0,0 +1,74 @@
...
+#  To contact SUSE about this file by physical or electronic mail,
+#  you may find current contact information at www.suse.com
+
+require "yast"
+
+module Installation
+  # This class is responsible for building a summary for SshImporter
+  # objects. Moving the presentation to a different class, avoid
+  # SshImporter knowing about i18n and HTML.
+  class SshImporterPresenter
```
lslezak: Oh, a presenter! Nice design! :+1:  :+1:

test/ssh_import_auto_test.rb
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env rspec
+
+require_relative "test_helper"
+require "installation/clients/ssh_import_auto"
+
+describe ::Installation::SSHImportAutoClient do
+  let(:importer) { ::Installation::SshImporter.instance }
```
lslezak: Why not to use `subject` here? http://betterspecs.org/#subject

lslezak: Oh, it's a different class, my fault, it's OK.

test/ssh_import_auto_test.rb
```diff
@@ -0,0 +1,170 @@
...
+      it "returns SSH importer summary" do
+        expect(subject.run).to eq(presenter.summary)
+      end
+    end
+
+    context "Import" do
+      let(:func) { "Import" }
+
+      before do
+        importer.add_config(FIXTURES_DIR.join("root1"), "dev")
```
lslezak: DRY:
```ruby
def fixtures_dir(dir)
   FIXTURES_DIR.join(dir)
end

fixtures_dir("root1")
```

test/ssh_import_auto_test.rb
```diff
@@ -0,0 +1,170 @@
...
+        it "sets default device to be used" do
+          subject.run
+          expect(importer.device).to eq("dev")
+        end
+      end
+
+      context "when copying configuration is disabled" do
+        let(:args) { { "import" => true, "config" => false } }
+
+        it "disable copy_config to false" do
```
lslezak: NP: "it disables copy_config" or "it sets copy_config to false"

test/ssh_import_auto_test.rb
```diff
@@ -0,0 +1,170 @@
...
+          subject.run
+          expect(importer.copy_config).to eq(false)
+        end
+      end
+
+      context "when copying configuration is enabled" do
+        let(:args) { { "import" => true, "config" => true } }
+
+        it "disable copy_config to true" do
```
lslezak: same here

test/ssh_import_auto_test.rb
```diff
@@ -0,0 +1,170 @@
...
+      end
+    end
+
+    context "SetModified" do
+      let(:func) { "SetModified" }
+
+      before { described_class.changed = false }
+
+      it "sets the client as 'modified'" do
+        expect { subject.run }.to change { subject.modified? }
```
lslezak: `expect...to change from... to...` Nice! :+1:

test/ssh_importer_presenter_test.rb
```diff
@@ -0,0 +1,84 @@
...
+      allow(Yast::Mode).to receive(:mode).and_return(mode)
+    end
+
+    context "when no previous configurations were found" do
+      let(:config) { "root3" }
+      let(:device) { nil }
+
+      context "and mode is installation" do
+        let(:mode) { "installation" }
+        let(:message) { _("No previous Linux installation found") }
```
lslezak: Um, I cannot see any `textdomain` call, I'm not sure if this would actually work in a non-English locale...

test/ssh_importer_presenter_test.rb
```diff
@@ -0,0 +1,84 @@
...
+
+        it "returns 'No previous Linux...' message" do
+          expect(presenter.summary).to eq("<ul><li>#{message}</li></ul>")
```
lslezak: NP: The test description says `returns 'No previous Linux...' message`, but you test also the HTML tags here. I'd make the test less strict and check only the message, e.g. `expect().to include(message)`.

src/lib/installation/clients/ssh_import_auto.rb
imobach: Let's ask @schubi2 about it.

imobach:
@schubi2 I've updated the PR taking into account @lslezak comments (thanks). I'm not sure about the `going_back` thingie and about changing the `config` AutoYaST element to `copy_config`.

lslezak:
The fixes looks good, thanks! Let's wait for @schubi2 to comment on the remaining issues.

src/lib/installation/clients/ssh_import_auto.rb
schubi2: copy and past error. I have removed all "back" stuff.

lslezak:
LGTM! :tada:  Thanks!

(Do not forget to use the "squash and merge" option when merging the PR, there are lots of commits.)

PR #374, opened 2016-05-10
fate#319624

ancorgs:
## Running this prototype

If you want to play with this, you'll need to:

 - Prepare a DUD with this updated yast2-installation
 - Make [theses changes](https://github.com/ancorgs/skelcd-control-openSUSE/commit/92eaf39163f8f4a3de83fc4cb9e34d70b02a5a7b) to control.xml 
 - Try to install in a system in which there is (at least) a previous Linux installation

## How it works

It presents a new section in the proposal

ssh_import_ncurses1

Clicking in the link in the proposal leads to this dialog

ssh_import_ncurses2

PR #382, opened 2016-05-25
SSH import proposal summary interface improvements (Fate#319624)

teclator:
This pull request is just for some minor changes based on [blog entry](https://lizards.opensuse.org/2016/05/18/highlights-of-yast-development-sprint-19/) feedback. The functionality was added and is explained in #374.

https://trello.com/c/EQzfj9Kf

## Screenshots

In case that there is not a previous installation the proposal will be disable:

ssh-key_proposal_link_no_partition

ssh-key_proposal_link_text_mode_no_partition

In other case:

ssh-key_proposal_link_with_partition
ssh-key_proposal_link_text_mode_with_partition

And the new dialog suggested by @kwwii:

ssh-key_dialog_new

ssh-key_dialog_text_mode_new

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed May 25 15:49:41 UTC 2016 - kanderssen@suse.com
+
+- More visual improvements in the SSH keys importing proposal
+  summary based on blog entry feedback.
+
+-------------------------------------------------------------------
 Wed May 25 13:07:59 UTC 2016 - lslezak@suse.cz
 
 - Start the Ruby debugger at the beginning of installation
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -3,6 +3,7 @@ Wed May 25 15:49:41 UTC 2016 - kanderssen@suse.com
 
 - More visual improvements in the SSH keys importing proposal
   summary based on blog entry feedback. (Fate#319624)
+- 3.1.189
 
 -------------------------------------------------------------------
 Wed May 25 13:07:59 UTC 2016 - lslezak@suse.cz
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -2,7 +2,7 @@
 Wed May 25 15:49:41 UTC 2016 - kanderssen@suse.com
 
 - More visual improvements in the SSH keys importing proposal
-  summary based on blog entry feedback.
+  summary based on blog entry feedback. (Fate#319624)
 
 -------------------------------------------------------------------
 Wed May 25 13:07:59 UTC 2016 - lslezak@suse.cz
--- a/src/lib/installation/clients/ssh_import_proposal.rb
+++ b/src/lib/installation/clients/ssh_import_proposal.rb
@@ -20,7 +20,9 @@ module Yast
         "rich_text_title" => _("Import SSH Host Keys and Configuration"),
         # menubutton entry
         "menu_title"      => _("&Import SSH Host Keys and Configuration"),
-        "id"              => "ssh_import"
+        # empty string is returned in case of not previous installation to
+        # avoid text link.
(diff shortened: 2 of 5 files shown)


src/lib/installation/dialogs/ssh_import.rb
```diff
@@ -61,12 +61,27 @@ def copy_config
...
+                HSpacing(2),
+                partitions_list_widget
+              ),
+              VSpacing(1),
+              HBox(
+                HSpacing(2),
+                Left(copy_config_widget)
+              )
+            )
+          )),
```
lslezak: The frame in the ncurses screenshot is moved to the right side, I think it should be centered.

I think adding a `HStretch` add the end would fix it. Please check.

package/yast2-installation.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed May 25 15:49:41 UTC 2016 - kanderssen@suse.com
+
+- More visual improvements in the SSH keys importing proposal
+  summary based on blog entry feedback.
```
lslezak: Missing BNC/FATE number.

src/lib/installation/dialogs/ssh_import.rb
```diff
@@ -112,7 +125,7 @@ def partition_widget(dev, partition)
     end
 
     def copy_config_widget
-      CheckBox(Id(:copy_config), _("Copy Whole SSH Configuration"), copy_config)
+      CheckBox(Id(:copy_config), _("Insert SSH Configuration"), copy_config)
```
lslezak: Um, I do not know what the original problem was here but for me the old "Copy Whole SSH Configuration" was more clear. It clearly means it will *copy everything* from the old system.

What does *Insert* mean in this context? What configuration it will insert? From the new system or the old one? Will it change somehow? I'm a bit confused here... :confused: 

But maybe it's confusing for me because I'm not a native English speaker?... Maybe we should double check with @kwwii?

teclator: I change it wrong, i realized out that just asking you about the PR, i put Insert instead of Import :-1:

src/lib/installation/dialogs/ssh_import.rb
teclator: Same result, i will try to fix it.

src/lib/installation/clients/ssh_import_proposal.rb
```diff
@@ -20,7 +20,9 @@ def description
         "rich_text_title" => _("Import SSH Host Keys and Configuration"),
         # menubutton entry
         "menu_title"      => _("&Import SSH Host Keys and Configuration"),
-        "id"              => "ssh_import"
+        # empty string is returned in case of not previous installation to
+        # avoid text link.
+        "id"              => importer.configurations.empty? ? "" : "ssh_import"
```
teclator: It doesn't work in text mode. So i will try with DisableProposal.

lslezak:
Some notes to the screenshots:

- In the installation proposal we should use a list item to have the same look. The plain text item it not aligned with the rest, all items are indented, this line stands out.

- I'd add some more vertical space between the radio button list with partitions and the "Insert SSH Configuration" check box. Esp. in the text mode it's not clear that this is not a radio button but a separate check box.

- The "No previous Linux installation found - not importing any SSH Key" looks too long to me, the other installation summary items are quite short. Maybe simple "No previous Linux installation found" should be enough?

lslezak:
Maybe out of scope of this PR but I'd probably change it as well:

```ruby
res += " " + _("(<a href=%s>change</a>)") % '"ssh_import"'
```

The question is if we actually need the link...

Does clicking the "*Import SSH Host Keys and Configuration*" title open the same dialog? If yes then IMHO the link is redundant. See the *Kdump* section above, it does not use any *change* link...

lslezak:
The screens looks much better now! Thanks!

LGTM :tada:

PR #383, opened 2016-05-26
System Role: centered dialog for ncurses.

teclator:
Testing #382 i realized that system role dialog was also not centered.

rolesselectionleft

So, after merge this PR it will look like..

rolesselection

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Thu May 26 13:17:42 UTC 2016 - kanderssen@suse.com
+
+- System Role: centered dialog (ncurses).
+
+-------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ancorgs:
LGTM

package/yast2-installation.changes
```diff
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Thu May 26 13:17:42 UTC 2016 - kanderssen@suse.com
+
+- System Role: centered dialog (ncurses).
```
kobliha: Please add (some) BNC number, otherwise your submit request will not be accepted.
And BTW, without a version change, there will be no SR, so if that is what you want ... ;)

teclator: Yeah, that is the case, it was not exactly part of the other PR, so not part of that Fate request, so i just added the change, but is just a visual change, no behavior or api changed so i didn't incremented the version.

PR #384, opened 2016-05-26
Review 319624 AutoYaST

imobach:
This PR is based on #380:

* Add unit test for `SSHImportAutoClient`.
* Move code from `SshImporter` to `SSHImportAutoClient`. IMHO, the main responsible for handling the import/write/etc. at a "top-level" should be the client itself. SshImporter class shouldn't now anything about AutoYaST, although it's the class that does the final job.

On the other hand, I'm not sure if `SSHImportAutoClient` should be here or in `yast-autoinstallation`.

@schubi2 Of course, I could be wrong :). In that case, you can use at least the unit tests :) Feel free to cherry pick them.

### DIFF
--- a/src/lib/installation/clients/ssh_import_auto.rb
+++ b/src/lib/installation/clients/ssh_import_auto.rb
@@ -22,13 +22,55 @@ module Installation
       ret
     end
 
+    # Importing data from the AutoYaST configuration module
+    # AutoYaST data format:
+    #
+    # <ssh_import>
+    #   <import config:type="boolean">true</import>
+    #   <config config:type="boolean">true</config>
+    #   <device>/dev/sda4</device>
+    # </ssh_import>
     def import(data)
-      ::Installation::SshImporter.instance.import(data)
+      if data["import"]
+        log.info "Importing AutoYaST data: #{data}"
+        ssh_importer.copy_config = data["config"] == true
+        if data["device"] && !data["device"].empty?
+          if ssh_importer.configurations.key?(data["device"])
+            ssh_importer.device = data["device"]
+          else
+            Yast::Report.Warning(
+              # TRANSLATORS: %s is the device name like /dev/sda0
+              _("Device %s not found. Taking default entry.") %
+              data["device"]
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -20,13 +20,55 @@ def run
...
+      end
       true
     end
 
+    # Returns a human readable summary
     def summary
-      "<UL>" + ::Installation::SshImporter.instance.summaray + "</UL>"
+      message =
+        if ssh_importer.configurations.empty?
+          _("No previous Linux installation found - not importing any SSH Key")
```
teclator: " - not importing any SSH Key" was removed in #382

src/lib/installation/clients/ssh_import_auto.rb
```diff
@@ -20,13 +20,55 @@ def run
...
+            # TRANSLATORS: %s is the name of a Linux system found in the hard
+            # disk, like 'openSUSE 13.2'
+            _("SSH host keys and configuration will be copied from %s") % name
+          else
+            # TRANSLATORS: %s is the name of a Linux system found in the hard
+            # disk, like 'openSUSE 13.2'
+            _("SSH host keys will be copied from %s") % name
+          end
+        end
+      "<UL>#{message}</UL>"
```
teclator: If YaST::HTML.List(message) is not used then at least add the li element to generate a correct unordered list entry ```"<ul><li>#{message}</li></ul>"```

imobach: Yes, the list is wrong. And maybe we should use the `Summary` module (I'll check it).

src/lib/installation/clients/ssh_import_auto.rb
imobach: Thanks! I overlooked that.

schubi2:
Thanks for help ! LGTM

imobach:
You're welcome!

PR #385, opened 2016-05-27
Documented the pattern for actual self_update_url values (FATE#319716).

mvidner:
https://trello.com/c/SWxjqjos
https://fate.suse.com/319716


jreidinger:
LGTM

doc/SELF_UPDATE.md
```diff
@@ -60,6 +60,23 @@ The URL can contain a variable `$arch` that will be replaced by the system's
 architecture, such as `x86_64`, `s390x`, etc. You can find more information
 in the [Arch module](http://www.rubydoc.info/github/yast/yast-yast2/Yast/ArchClass).
 
+### Actual URLs
+
+The regular update URLs have the form
```
jreidinger: citation needed, where you get such info? it would be good if someone ask us why we use this form

mvidner: Good point. The source is observation of the file server, not some API specification, of which I don't know.

PR #386, opened 2016-05-30
Move the debugger invocation to ruby-bindings (FATE#318421)

lslezak:
... to use the same implementation at run time.

- 3.1.190

The related ruby-bindings PR is here: https://github.com/yast/yast-ruby-bindings/pull/168

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon May 30 14:35:05 UTC 2016 - lslezak@suse.cz
+
+- Move the debugger invocation code to yast2-ruby-bindings package
+  to use the same implementation at run time (FATE#318421)
+- 3.1.190
+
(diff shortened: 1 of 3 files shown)


jreidinger:
well, it is a bit backward incompatible as new code do not recognize e.g. `Y2Debugger`.

BTW does linuxrc way still works? for me it looks like you have to specify ENV on CLI, not?

lslezak:
Linuxrc sets the environment variables for the passed boot options, if you boot with `Y2DEBUGGER=1` then `ENV["Y2DEBUGGER"]` is set to `"1"`.

Yes, a good point, I'll improve handling for `Y2Debugger`.

lslezak:
Updated at https://github.com/yast/yast-ruby-bindings/pull/168

jreidinger:
ok, so LGTM ( btw we no longer need newer ruby bindings, but it do not harm )

PR #387, opened 2016-05-31
ssh import: updates from code review

imobach:
Updates from code review for #380.

As we need `SshImporterProposal` and `SshImporterAutoClient` to show the same `summary`, instead of moving this logic to `SshImporter#summary`, I preferred to add a presenter. I don't like the idea of `SshImporter` knowing about i18n or HTML.


schubi2:
LGTM

PR #391, opened 2016-06-07
SSH Importer: Width fix to avoid cut of CheckBoxFrame Label

teclator:
By some reason the Label of the CheckBoxFrame is being cut. Now width is forced, but could be nice if some UI expert reviewed it as i'm not sure if could be some problem with long texts **(translations)**.

## Previously
ssh-key_dialog_new

## With MinWidth

ssh-key_dialog_new_length

### DIFF
--- a/src/lib/installation/dialogs/ssh_import.rb
+++ b/src/lib/installation/dialogs/ssh_import.rb
@@ -68,7 +68,8 @@ module Yast
           # For some reason the CheckBoxFrame Label is cut if the label size
           # exceeds the CheckBoxFrame content's width. MinWidth with label length
           # is used to avoid this issue.
-          MinWidth(label.length,
+          MinWidth(
+            label.length,
             CheckBoxFrame(
               Id(:import_ssh_key),
               label,
--- a/src/lib/installation/dialogs/ssh_import.rb
+++ b/src/lib/installation/dialogs/ssh_import.rb
@@ -62,23 +62,26 @@ module Yast
     end
 
     def dialog_content
+      label = _("I would like to import SSH keys from a previous installation")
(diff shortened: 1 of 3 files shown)


src/lib/installation/dialogs/ssh_import.rb
```diff
@@ -64,21 +64,23 @@ def copy_config
...
-              VSpacing(1),
-              HBox(
-                HSpacing(2),
-                Mode.config ? device_name : partitions_list_widget
-              ),
-              VSpacing(3),
-              HBox(
-                HSpacing(2),
-                Left(copy_config_widget)
+          MinWidth(50,
```
ancorgs: Obviously, the right fix is to discover why the cut is happening. But, once we are already in the business of workarounds... why don't make it i18n aware with something like...?

```ruby
label = _("I would like to import SSH keys from a previous installation")
MinWidth(label.length,
...
```

ancorgs:
No version increase?

src/lib/installation/dialogs/ssh_import.rb
```diff
@@ -62,23 +62,26 @@ def copy_config
...
-              VSpacing(1),
-              HBox(
-                HSpacing(2),
-                Mode.config ? device_name : partitions_list_widget
-              ),
-              VSpacing(3),
-              HBox(
-                HSpacing(2),
-                Left(copy_config_widget)
+          MinWidth(label.length,
```
ancorgs: Maybe add a comment explaining why this MinWidth is there

jreidinger:
LGTM

jreidinger:
LGTM

PR #393, opened 2016-06-13
Fix debugging message in UpdateRepository

imobach:


imobach:
Already fixed in #394.

PR #394, opened 2016-06-14
Fix architecture detection

imobach:
Fix architecture detection. It fixes [bsc#984656](https://bugzilla.suse.com/show_bug.cgi?id=984656).


jreidinger:
LGTM

PR #396, opened 2016-06-16
Avoid restarting YaST when self-update repo is empty

imobach:
Fixes [bsc#985113](https://bugzilla.suse.com/show_bug.cgi?id=985113).

### DIFF
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -173,10 +173,14 @@ module Yast
     # @return [Boolean] true if installer was updated; false otherwise.
     def update_installer
       log.info("Adding update from #{self_update_url}")
-      updates_manager.add_repository(self_update_url)
-      log.info("Applying installer updates")
-      updates_manager.apply_all
-      true
+      if updates_manager.add_repository(self_update_url)
+        log.info("Applying installer updates")
+        updates_manager.apply_all
+        true
+      else
+        log.info("No packages were found")
+        false
+      end
 
     rescue ::Installation::UpdatesManager::NotValidRepo
       if !using_default_url?
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -173,14 +173,13 @@ module Yast
     # @return [Boolean] true if installer was updated; false otherwise.
     def update_installer
       log.info("Adding update from #{self_update_url}")
-      if updates_manager.add_repository(self_update_url)
+      updates_manager.add_repository(self_update_url)
+      updated = updates_manager.repositories?
+      if updated
         log.info("Applying installer updates")
         updates_manager.apply_all
-        true
-      else
-        log.info("No packages were found")
-        false
       end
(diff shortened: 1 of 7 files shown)


test/updates_manager_test.rb
```diff
@@ -12,8 +12,8 @@
 
   let(:uri) { URI("http://updates.opensuse.org/sles12") }
 
-  let(:repo0) { double("repo0", apply: true, cleanup: true) }
-  let(:repo1) { double("repo1", apply: true, cleanup: true) }
+  let(:repo0) { double("repo0", apply: true, cleanup: true, :empty? => false) }
+  let(:repo1) { double("repo1", apply: true, cleanup: true, :empty? => false) }
```
lslezak: Use the same Hash items style. e.g. `empty?: false`.

src/lib/installation/updates_manager.rb
```diff
@@ -75,13 +75,19 @@ def initialize(duds_paths = DRIVER_UPDATES_PATHS)
...
       new_repository = Installation::UpdateRepository.new(uri)
       new_repository.fetch
-      @repositories << new_repository
+      if new_repository.empty?
+        log.info("Update repository at #{uri} is empty. Skipping...")
+        false
+      else
+        @repositories << new_repository
+        true
+      end
```
lslezak: NP: Maybe should simply return the `!new_repository.empty?` value instead of the hardcoded booleans... 

```ruby
if new_repository.empty?
  log.info("Update repository at #{uri} is empty. Skipping...")
else
  @repositories << new_repository
end

!new_repository.empty?
```

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -173,10 +173,14 @@ def update_flag_file
     # @return [Boolean] true if installer was updated; false otherwise.
     def update_installer
       log.info("Adding update from #{self_update_url}")
-      updates_manager.add_repository(self_update_url)
-      log.info("Applying installer updates")
-      updates_manager.apply_all
-      true
+      if updates_manager.add_repository(self_update_url)
```
lslezak: I'd break the functionality into two methods, this does not look much intuitive, and we already log that the repo is empty in `add_repository` e.g.:

```ruby
updates_manager.add_repository(self_update_url)

if !updates_manager.empty?
  log.info("Applying installer updates")
  updates_manager.apply_all
end

updates_manager.empty?
```

(The `UpdatesManager#empty?` method needs to be added.)

What do you think about that?

imobach: Initially I wrote:

```ruby
if updates_manager.repositories.empty?
  log.info("Applying installer updates")
  updates_manager.apply_all
#...
```

But I didn't want to break the Law of Demeter :) Anyway, I'll try that approach but changing `empty?` by `has_repositories?` (because UpdatesManager.driver_updates may not be empty). To make it clear, if `has_repositories?` is false, the driver updates won't be re-applied again (it does not make sense).

lslezak:
LGTM

imobach:
Thank you!

PR #397, opened 2016-06-23
More information in the self update error popup (bsc#986091)

lslezak:
The current error message is a bit scary for the customers, we need to tell them that the installer updates are optional and it is OK to continue without them.

See [bug #986091](https://bugzilla.suse.com/show_bug.cgi?id=986091) for more details.

- 3.1.197

#### The Original Message
self_update_error_old

#### The Improved Message

self_update_error


mvidner:
Thanks, LGTM

PR #398, opened 2016-06-27
Optimized bootloader

jreidinger:


package/yast2-installation.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Jun 27 13:00:24 UTC 2016 - jreidinger@suse.com
+
+- Always run kdump before final bootloader installation to allow
+  skip kdump writting of bootloader and speed up installation
```
mvidner: to allow skip*ping* the call of bootloader from kdump and speeding up the installation

mvidner:
The changes file mention kdump but then you also change cio-ignore.

package/yast2-installation.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Jun 27 13:00:24 UTC 2016 - jreidinger@suse.com
+
+- Always run kdump before final bootloader installation to allow
```
mvidner: "Always run kdump" is confusing because it is run conditionally depending on product feature flags.

package/yast2-installation.changes
mvidner: I see no skipping of bootloader from kdump in the code.

package/yast2-installation.changes
mvidner: How about
> Make writing of bootloader settings the last step so that other installation steps (kdump, cio-ignore) do not have to waste time repeating it

jreidinger: yes, make sense

jreidinger:
updated

mvidner:
LGTM

jreidinger:
thanks

PR #399, opened 2016-07-01
Inst finish cleaning

jreidinger:


test/inst_finish_test.rb
```diff
@@ -0,0 +1,175 @@
...
+      expect(subject.main).to eq :abort
+    end
+
+    it "returns :auto if going back in installation" do
+      expect(Yast::GetInstArgs).to receive(:going_back).and_return(true)
+
+      expect(subject.main).to eq :auto
+    end
+
+    it "disabled during its run Back button" do
```
imobachgs: What about *disables Back button while running*?

test/inst_finish_test.rb
```diff
@@ -0,0 +1,175 @@
...
+      expect(Yast::Wizard).to receive(:DisableBackButton)
+
+      subject.main
+    end
+
+    it "disabled during its run Next button" do
```
imobachgs: Same as above.

test/inst_finish_test.rb
```diff
@@ -0,0 +1,175 @@
...
+
+      it "opens dialog unless already opened" do
+        allow(Yast::SlideShow).to receive(:HaveSlideWidget).and_return(false)
+
+        expect(Yast::SlideShow).to receive(:OpenDialog)
+
+        subject.main
+      end
+
+      it "closes dialog if it is opened by method" do
```
imobachgs: np: I think that the adjective is *open* and the verb past form is *opened*.

test/inst_finish_test.rb
```diff
@@ -0,0 +1,175 @@
...
+    end
+
+    context "finish clients" do
+      it "Call info for all specified finish clients" do
+        expect(Yast::WFM).to receive(:CallFunction).at_least(1).and_return({})
+
+        subject.main
+      end
+
+      it "filter out all cliets which 'when' key do not match current mode" do
```
imobachgs: cliets -> clients

test/inst_finish_test.rb
jreidinger: it is actually not while running as it keep disabled. I won't re enable it or restore previous status.

jreidinger: but you are right, I want to write disables..so maybe simple `disables Back button` ?

test/inst_finish_test.rb
jreidinger: I plan to use passive form, not adjective for it.  http://www.englishpage.com/verbpage/activepassive.html so it should be past form here. @mvidner your opinion?

jreidinger: so idea is `method opens dialog` and `dialog is opened by method`

imobachgs: You're right.

imobachgs:
Nice improvement, thanks. Now it looks like Ruby code. LGTM.

package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jul 21 11:52:59 UTC 2016 - jreidinger@suse.com
+
+- simplify and speed up inst_finish client(bnc#986649)
```
imobachgs: Just add an space before the `bnc`.

imobachgs:
LGTM

PR #400, opened 2016-07-04
Get self-update URL from SMT through SUSEconnect

imobachgs:
Add SMT integration in [FATE#319716](https://fate.suse.com/319716).

Only 1 URL is currently supported, but `list_installer_updates` will return an `Array` anyway. If we want to support multiple URLs, we should also answer some additional questions:

* Should we support multiple self-update values in control.xml and AutoYaST profiles?
* Should we add support to add a self-update URL (not replace) in Linuxrc?

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Jul  4 08:05:53 UTC 2016 - igonzalezsosa@suse.com
+
+- Retrieve self-update URL from SMT (FATE#319716)
+- 3.1.199
+
+-------------------------------------------------------------------
 Mon Jun 27 13:00:24 UTC 2016 - jreidinger@suse.com
 
 - Make writing of bootloader settings the last step so that other
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,7 +1,8 @@
 -------------------------------------------------------------------
 Mon Jul  4 08:05:53 UTC 2016 - igonzalezsosa@suse.com
 
-- Retrieve self-update URL from SMT (FATE#319716)
+- Retrieve the self-update URL from the registration
+  server (SCC/SMT) (FATE#319716)
 - 3.1.199
 
 -------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


lslezak:
I'd keep it simple, i.e. allow only one URL in the linuxrc, AY profile and control.xml and allow multiple in the registration server response.

package/yast2-installation.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Jul  4 08:05:53 UTC 2016 - igonzalezsosa@suse.com
+
+- Retrieve self-update URL from SMT (FATE#319716)
```
lslezak: ` Retrieve the self-update URL from the registration server (SCC/SMT)`  to be more precise. It's not SMT only.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -86,11 +86,34 @@ def self_update_enabled?
...
+    # (for instance in openSUSE).
+    #
+    # @return [URI,nil] self-update URL. nil if no URL was found.
+    def self_update_url_from_connect
+      require "registration/sw_mgmt"
+      require "registration/url_helpers"
+      require "suse/connect"
+      base_product = Registration::SwMgmt.base_product_to_register
+      product = Registration::SwMgmt.remote_product(base_product)
+      update = SUSE::Connect::YaST.list_installer_updates(product,
```
lslezak: Please update the required SUSEconnect version, this has been added in version 0.2.37 (see https://github.com/SUSE/connect/pull/268).

lslezak: Oops, we cannot, do not have any dependency here...

lslezak:
LGTM, thanks! :+1:

src/lib/installation/clients/inst_update_installer.rb
jreidinger: you can add conflict with lower version forcing installation of newer one ;)

imobachgs:
PR #401 supersedes this one. Closing.

PR #401, opened 2016-07-07
 Get self-update URL from SMT through SUSEconnect (with multiple URLs support)

imobachgs:
Improved version of #400 to support multiple URLs coming from SCC/SMT.

It depends on https://github.com/yast/yast-registration/pull/263 (merged).

### DIFF
--- a/doc/SELF_UPDATE.md
+++ b/doc/SELF_UPDATE.md
@@ -36,7 +36,7 @@ handled in a different way:
 The URL of the update repository is evaluated in this order:
 
 1. The `SelfUpdate` boot option
-2. The AutoYaST control file - in AutoYaST installation only, use the
+2. The AutoYaST profile - in AutoYaST installation only, use the
    `/general/self_update_url` XML node:
 
    ```xml
@@ -44,6 +44,14 @@ The URL of the update repository is evaluated in this order:
      <self_update_url>http://example.com/updates/$arch</self_update_url>
    </general>
    ```
+3. Registration server (SCC/SMT), not available in openSUSE. The URL of the
+   registration server which should be used is determined via:
+   1. AutoYaST profile ([reg_server element](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#CreateProfile.Register)).
+   2. The `regurl` boot parameter
+   3. SLP lookup:
+      * If one server is found, it will be used automatically.
+      * If more than one server is found, it will ask the user to choose one.
+   4. Default SUSE Customer Center API (`https://scc.suse.com/`).
 3. Hard-coded in the `control.xml` file on the installation medium (thus it
    depends on the base product):
 
@@ -62,7 +70,7 @@ in the [Arch module](http://www.rubydoc.info/github/yast/yast-yast2/Yast/ArchCla
 
 ### Actual URLs
 
-The regular update URLs have the form
+Whe using registration servers, the regular update URLs have the form
 `https://updates.suse.com/SUSE/Updates/$PRODUCT/$VERSION/$ARCH/update` where
 - PRODUCT is like OpenStack-Cloud, SLE-DESKTOP, SLE-SDK, SLE-SERVER,
 - VERSION (for SLE-SERVER) is like 12, 12-SP1,
--- a/doc/SELF_UPDATE.md
+++ b/doc/SELF_UPDATE.md
@@ -48,7 +48,7 @@ The URL of the update repository is evaluated in this order:
    registration server which should be used is determined via:
    1. AutoYaST profile ([reg_server element](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#CreateProfile.Register)).
    2. The `regurl` boot parameter
-   3. SLP lookup:
+   3. SLP lookup (this behavior applies to regular and AutoYaST installations):
       * If one server is found, it will be used automatically.
       * If more than one server is found, it will ask the user to choose one.
    4. Default SUSE Customer Center API (`https://scc.suse.com/`).
@@ -109,3 +109,22 @@ Currently there is no API available for remembering the client states. The easie
 way is to store the configuration into an YAML file and load it when restarting the
 installer. See the [example](https://github.com/yast/yast-installation/pull/367/files#diff-4c91d6424e08c9bef9237f7d959fc0c2R48)
 in the `inst_complex_welcome` client.
+
+## Error handling
+
+Errors during the installer update are handled as described below:
+
+* If network is not available, the installer update will be skipped.
+* If the network is configured but the installer updates repository or the
+  registration server are not reachable:
+  * in a regular installation/upgrade, YaST2 will offer the possibility
+    to check/adjust the network configuration.
+  * in an AutoYaST installation/upgrade, a warning will be shown.
+* If the updates repository is found but it is empty or not valid:
+  * in the case that the URL was specified by the user (using the *SelfUpdate* boot
+    option or through the *self_update_url* element in an AutoYaST profile), an
+    error message will be shown.
+  * if the URL was not specified by the user, the installer will skip the update
+    process (it will assume that no updates are available).
+* If something goes wrong trying to fetch and apply the update, the user will be
+  notified.
--- a/doc/SELF_UPDATE.md
+++ b/doc/SELF_UPDATE.md
@@ -52,7 +52,7 @@ The URL of the update repository is evaluated in this order:
       * If one server is found, it will be used automatically.
       * If more than one server is found, it will ask the user to choose one.
    4. Default SUSE Customer Center API (`https://scc.suse.com/`).
-3. Hard-coded in the `control.xml` file on the installation medium (thus it
+4. Hard-coded in the `control.xml` file on the installation medium (thus it
    depends on the base product):
 
    ```xml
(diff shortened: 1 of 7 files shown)


doc/SELF_UPDATE.md
```diff
@@ -36,14 +36,22 @@ handled in a different way:
...
    ```xml
    <general>
      <self_update_url>http://example.com/updates/$arch</self_update_url>
    </general>
    ```
+3. Registration server (SCC/SMT), not available in openSUSE. The URL of the
+   registration server which should be used is determined via:
+   1. AutoYaST profile ([reg_server element](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#CreateProfile.Register)).
+   2. The `regurl` boot parameter
+   3. SLP lookup:
```
jsrain: How is the SLP lookup implemented when it comes to AutoYaST? Should be documented too...

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -55,6 +61,32 @@ def main
...
+    #
+    # Errors handling:
+    #
+    # * A repository is not found: warn the user if she/he is using
+    #   a custom URL.
+    # * Could not fetch update from repository: report the user about
+    #   this error.
+    # * Repository could not be probed: suggest checking network
+    #   configuration if URL has a REMOTE_SCHEME.
+    #
```
jsrain: Maybe the error handling should also be summarized in the "README" document?

jsrain:
Besides the two comments above, I'm fine with this PR (but someone more rubyish should review the code itself).

imobachgs:
@jsrain Thanks! I'll address those documentation issues and I'll ask for a second review.

jsrain:
The docu looks good to me, thanks!

doc/SELF_UPDATE.md
```diff
@@ -36,14 +36,22 @@ handled in a different way:
...
      <self_update_url>http://example.com/updates/$arch</self_update_url>
    </general>
    ```
+3. Registration server (SCC/SMT), not available in openSUSE. The URL of the
+   registration server which should be used is determined via:
+   1. AutoYaST profile ([reg_server element](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#CreateProfile.Register)).
+   2. The `regurl` boot parameter
+   3. SLP lookup (this behavior applies to regular and AutoYaST installations):
+      * If one server is found, it will be used automatically.
+      * If more than one server is found, it will ask the user to choose one.
```
mvidner: Does this mean AutoYaST will require interaction if 2 servers are found via SLP?

doc/SELF_UPDATE.md
```diff
@@ -36,14 +36,22 @@ handled in a different way:
...
+   4. Default SUSE Customer Center API (`https://scc.suse.com/`).
 3. Hard-coded in the `control.xml` file on the installation medium (thus it
```
mvidner: 3\. -> 4.

doc/SELF_UPDATE.md
```diff
@@ -62,7 +70,7 @@ in the [Arch module](http://www.rubydoc.info/github/yast/yast-yast2/Yast/ArchCla
 
 ### Actual URLs
 
-The regular update URLs have the form
+Whe using registration servers, the regular update URLs have the form
```
mvidner: Whe**n**

package/yast2-installation.spec
```diff
@@ -127,6 +130,8 @@ Recommends:	curl
 Recommends:	yast2-update
 Recommends:	yast2-add-on
 
+Conflicts: yast2-registration < 3.1.177
```
mvidner: This is unnecessary because of `Conflicts:  yast2-registration < 3.1.179` above, right?

doc/SELF_UPDATE.md
imobachgs: Yes, that's it. According to the documentation, when you do a SLP lookup for registration, only 1 server is expected. If more than one is found, the installation will fail. Extracted from [AutoYaST docs](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#CreateProfile.Register):

> Expects to find a single server. If more than one server is found, the installation will fail. In case > there is more than one registration server available, you need to specify one with reg_server. 

On the other hand, maybe we should also honor the `slp_discovery` AutoYaST option (so you can disable SLP at all).

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -268,7 +389,50 @@ def could_not_probe_repo_msg
...
+    # @params [URI] Registration server URL.
+    def store_registration_url(url)
+      data = { "custom_url" => url.to_s }
+      File.write(REGISTRATION_DATA_PATH, data.to_yaml)
+    end
+
+    # Load URL of registration server to be used by inst_scc client
+    #
+    # @return [Boolean] true if data was loaded; false otherwise.
```
mvidner: The only caller does not care about the return value.
I am not sure whether we should remove the boolean altogether, or promote the failure to an Internal Error.

mvidner:
The PR title says "multiple URLs" but the changes entry does not mention that. Also, what does it actually mean? Does it mean that YaST will use a single URL but it looks for it in multiple places? Or that one place can specify multiple URLs? Will all of them be combined? Or just the first one that succeeds? I did not find it in SELF_UPDATE.md.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -75,44 +107,146 @@ def self_update_enabled?
...
+    #
+    # Determined in the following order:
+    #
+    # * via AutoYaST profile
+    # * regurl boot parameter
+    # * SLP look up
+    #   * If there's only 1 SMT server, it will be chosen automatically.
+    #   * If there's more than 1 SMT server, it will ask the user to choose one
+    #
+    # @return [String,Symbol] Registration URL; :scc if SCC server was selected;
```
mvidner: I should know better than suggesting `nil`. Make that `@return [URI,:scc]`

imobachgs:
@mvidner From the user's point of view, we only handle 1 URL. However, SCC/SMT (and only SCC/SMT) will return an array of URLs with only 1 element. Initially, it's not likely to change in the near future, but I wanted to prepare the code for the future so we won't have to refactor in case SCC/SMT returns more than 1 URL.

Having said that, URLs won't be combined in any way.

mvidner:
> Having said that, URLs won't be combined in any way.

IIUC we will add repos for *all* of the provided URLs, and they will all be used. So it seems appropriate to document "If SCC/SMT provides multiple update repos, they will all be used."

https://github.com/imobachgs/yast-installation/blob/7cf7546313a9ae32caf8c4056e3e633399c25eb8/src/lib/installation/clients/inst_update_installer.rb#L78-L86

imobachgs:
@mvidner Yes, I didn't described it correctly. URLs *from different sources* won't be combined. I'll document it.

src/lib/installation/clients/inst_update_installer.rb
imobachgs: You're right that the caller won't care about it. But in this case, the  `false` is not an error at all: it simply tries to load the registration URL from the file (after the YaST2 restart) and it won't be available in openSUSE (as registration stuff is not available). I only wanted to distinguish some scenarios in the API.

mvidner: OK

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -182,23 +182,23 @@ def self_update_url_from_connect
...
       services = ::Registration::UrlHelpers.slp_discovery
-      return nil if services.empty?
+      return :scc if services.empty?
       service =
         if services.size > 1
           registration_service_from_user(services)
         else
           services.first
         end
       return service unless service.respond_to?(:slp_url)
```
mvidner: It seems this statement can return a `Yast::SlpServiceClass::Service`: http://www.rubydoc.info/github/yast/yast-registration/master/Registration/UrlHelpers.slp_discovery

mvidner: Actually slp_discovery always returns  something which does respond to slp_url, (at least in `master`) so it seems this check can be removed.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -75,44 +107,146 @@ def self_update_enabled?
...
+      return URI(url) if url
+      services = ::Registration::UrlHelpers.slp_discovery
+      return :scc if services.empty?
+      service =
+        if services.size > 1
+          registration_service_from_user(services)
+        else
+          services.first
+        end
+      return service unless service.respond_to?(:slp_url)
```
mvidner: This line is strange. It seems unnecessary because slp_discovery always (in master) returns something which can respond to slp_url.

test/inst_update_installer_test.rb
```diff
@@ -20,17 +32,23 @@
...
+  let(:profile) { {} }
+  let(:ay_profile) { double("Yast::Profile", current: profile) }
 
   before do
     allow(Yast::Pkg).to receive(:GetArchitecture).and_return(arch)
     allow(Yast::Mode).to receive(:auto).and_return(false)
     allow(Yast::NetworkService).to receive(:isNetworkRunning).and_return(network_running)
     allow(::Installation::UpdatesManager).to receive(:new).and_return(manager)
-    allow(Yast::Installation).to receive(:restarting?)
+    allow(Yast::Installation).to receive(:restarting?).and_return(restarting)
```
mvidner: Also, to remove a dependency on recent yast2.rpm:
```rb
    allow(Yast::Installation).to receive(:finish_restarting!)
```

src/lib/installation/clients/inst_update_installer.rb
imobachgs: But `registration_service_from_user` could return a symbol (like `:scc`). I could write something like `is_a?(Yast::SlpServiceClass::Service)` but I didn't to make this client aware at all of `SlpService` at all.

mvidner: Ah, I see, OK

mvidner:
LGTM, thank you!

imobachgs:
@mvidner Thanks a lot for your review!

PR #403, opened 2016-07-13
Making the installer run with storage-ng: installation part

ancorgs:


jsrain:
LGMT if pending checks pass

PR #404, opened 2016-07-14
Fixed OOM killer setup

lslezak:
This is a fix up for commit 65a6b6f0.

The `oom_score_adj` file has a different range than the original `oom_adj`, value `-1000` is the equivalent of the original `-17`.

See the [kernel proc documentation](https://www.kernel.org/doc/Documentation/filesystems/proc.txt), section 3.1.

*Note: the second commit is actually just an update for the Rubocop config (https://github.com/yast/yast-devtools/pull/107).*

- 3.1.200

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jul 14 08:10:16 UTC 2016 - lslezak@suse.cz
+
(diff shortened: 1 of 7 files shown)


jreidinger:
LGTM, can this be root of issue on s390?

lslezak:
No, this actually won't affect that s390 bug. This code disables the OOM killer for the X server, in that bug kernel killed `mke2fs`...

PR #405, opened 2016-07-14
Configure rake to commit to YaST:storage-ng

ancorgs:


jreidinger:
LGTM

ancorgs:
Merging despite Travis, since breakage is because of the rubocop change. Let's not block because of that.

PR #406, opened 2016-07-14
update rubocop

jreidinger:


src/lib/installation/clients/inst_ask_online_update.rb
```diff
@@ -148,12 +148,12 @@ def main
           Wizard.ShowHelp(@help)
         elsif @ret == :next
           # Skipping online update
-          if Convert.to_boolean(UI.QueryWidget(Id(:noupdate), :Value))
-            Internet.do_you = false
+          Internet.do_you = if Convert.to_boolean(UI.QueryWidget(Id(:noupdate), :Value))
```
lslezak: We could simply it to

```
Internet.do_you = !UI.QueryWidget(Id(:noupdate), :Value)
```

src/lib/installation/clients/inst_license.rb
```diff
@@ -143,28 +143,27 @@ def main
...
         if @test_mode != true
           SCR.Execute(path(".target.bash"), "/sbin/halt -f -n -p")
         end
-        return :abort
-      elsif @ask_ret == :next
-        return :next
-      elsif @ask_ret == :accepted
-        return :next
+        :abort
+      when :next
```
lslezak: These two conditions should be merged.

src/lib/installation/copy_logs_finish.rb
```diff
@@ -46,6 +46,7 @@ def modes
       [:installation, :live_installation, :update, :autoinst]
     end
 
+    PREFIX_SIZE = 6 # "y2log-".size
```
lslezak: NP: I'd prefer `"y2log-".size` here, the evaluation of that short string is fast anyway.

lslezak:
LGTM

PR #407, opened 2016-07-18
Instsys cleanup

lslezak:
This PR adds a new installation code which removes files from the inst-sys when the system has low memory. If the system has enough memory (at least 1GB) the inst-sys is kept untouched. 

The inst-sys cleanup functionality will be called after partitioning and before installing the packages.
The partitioner might load some kernel modules during disk formatting and mounting the target system (e.g. the filesystem modules), we cannot move the cleanup earlier.

*Note: We would need to split the cleanup it into two parts which would make it more complicated, let's use this simplified version for now. If it is still not sufficient we can improve it later.*

## Removed Parts

- Deletes the `raw` libzypp cache (about 2.5MB when installing pure SLES without any addons and without the Updates from the registration server)
- Delets the kernel modules, all needed modules should be already loaded (29MB on x86_64, 5MB on s390)

## Test

I have tested this change in a patched (DUD) Beta4 installation in an x86_64 VirtualBox VM with 512MB RAM.

*Note: To work properly this update also needs a patch in the `control.xml` to call the new client. I'll adapt the control packages later in separate pull requests.*

### Result

The memory statistics logged by the test installation:

#### Before Freeing the Memory

```
             total       used       free     shared    buffers     cached
Mem:           487        478          8        154          7        256
-/+ buffers/cache:        214        272
Swap:         1245          0       1245
```

```
Filesystem                                   1M-blocks  Used Available Use% Mounted on
tmpfs                                              488    98       391  20% /
```

#### After Freeing the Memory

```
             total       used       free     shared    buffers     cached
Mem:           487        447         40        123          7        224
-/+ buffers/cache:        214        272
Swap:         1245          0       1245
```

```
Filesystem                                   1M-blocks  Used Available Use% Mounted on
tmpfs                                              488    67       422  14% /
```

#### The Difference

As you can see the system has about **32MB more memory available for the installer**. That's quite a lot on a system with 512MB RAM...

*Note: The result highly depends on the system and the architecture. On S390 the freed memory is much smaller because the kernel modules take just about 5MB (on x86_64 it is about 29MB).*


src/lib/installation/instsys_cleaner.rb
```diff
@@ -0,0 +1,145 @@
...
+
+      # unmount the loop device
+      `umount #{KERNEL_MODULES_MOUNT_POINT}`
+      if $CHILD_STATUS && !$CHILD_STATUS.success?
+        log.warn("Unmouting #{KERNEL_MODULES_MOUNT_POINT} failed")
+        return
+      end
+
+      # remove the loop device binding
+      `losetup -d #{Shellwords.escape(device)}`
```
mvidner: If you use [`Yast::Execute.locally`](http://www.rubydoc.info/github/yast/yast-yast2/Yast/Execute#locally-class_method) then you won't have to escape (but will have to care about exit codes if you don't want a popup).

test/helpers.rb
```diff
@@ -12,4 +12,12 @@ module Helpers
   def fixtures_dir(*dirs)
     FIXTURES_DIR.join(*dirs)
   end
+
+  # Read the fixture file
+  # @param *path [Array<String>] path components
```
mvidner: NP, since yardoc omits test/ anyway: without the star, `@param path ...`

mvidner:
`unmount_kernel_modules` is quite long and the intent is hard to see. I think this would read better:

```rb
mount_point = find_mount_point || return
image = losetup_backing_file(mount_point) || return
Yast::Execute.locally("umount", image, allowed_exitstatus: 1..255) # errors goto y2log
Yast::Execute.locally("losetup", "-d", image, allowed_exitstatus: 1..255)
 ::FileUtils.rm_rf(image)
```

test/instsys_cleaner_test.rb
```diff
@@ -15,8 +15,8 @@ def stub_logging
         expect(Yast::Mode).to receive(:installation).and_return(true)
 
         # mock the logged memory stats
-        allow(subject.class).to receive(:`).with("df -m").and_return("")
-        allow(subject.class).to receive(:`).with("free -m").and_return("")
+        allow(Yast::Execute).to receive(:locally).with("df", "-m").and_return("")
+        allow(Yast::Execute).to receive(:locally).with("free", "-m").and_return("")
```
mvidner: drop `.and_return("")` because this form will not return anything

mvidner:
Just one minor point in the tests, otherwise LGTM.

mvidner:
LGTM

PR #408, opened 2016-07-21
fix build in suse and keep in sync build requires and travis requires

jreidinger:


test/inst_complex_welcome_test.rb
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env rspec
 
+# fake Console class to avoid build dependency on yast2-country
+module Yast
+  Console = "fake"
```
lslezak: Maybe mock it this way: https://github.com/yast/yast-installation/blob/master/test/test_helper.rb#L17 ?
It should be consistent across modules...

jreidinger: problem is that it need to mock all used methods, so it can be very long.

lslezak:
LGTM

PR #409, opened 2016-07-22
Don't halt AutoYaST installation on self-update problem

imobachgs:
It should fix [bsc#988949](https://bugzilla.suse.com/show_bug.cgi?id=988949). When running AutoYaST, if the self-update repo cannot be probed, it just show a error and don't halt the installation (unless the profile is configured to stop on errors).

### DIFF
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -196,8 +196,8 @@ module Yast
     rescue ::Installation::UpdatesManager::CouldNotProbeRepo
       if Mode.auto
         Report.Warning(could_not_probe_repo_msg)
-      else
-        retry if remote_self_update_url? && configure_network?
+      elsif remote_self_update_url? && configure_network?
+        retry
       end
       false
     end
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -213,21 +217,11 @@ def remote_self_update_url?
...
-                 "\n" \
-                 "If you need a proxy server to access the update repository\n" \
-                 "then use the \"proxy\" boot parameter.\n" \
-                 "\n" \
-                 "Would you like to check your network configuration\n" \
-                 "and try installing the updates again?"), self_update_url)
-      )
+      msg = could_not_probe_repo_msg + "\n" \
+        "Would you like to check your network configuration\n" \
+        "and try installing the updates again?"
```
ancorgs: Is this translatable?

imobachgs: Oh, I forgot.

ancorgs:
LGTM

PR #410, opened 2016-07-29
Don't ask twice about multipath bsc#989770

ancorgs:
Goes together with https://github.com/yast/yast-storage/pull/217


jreidinger:
LGTM

PR #413, opened 2016-08-01
Use y2storage for hardware probing

ancorgs:
To check it works, just run the installer and search for these lines in the logs after the system analysis step:

```
probed devicegraph begin
[...here is the information about your hardware as it's loaded in memory...]
probed devicegraph end
```

Tested in qemu-kvm with IDE & SCSI. Should work in real hardware and VirtualBox as well.

This goes together with https://github.com/yast/yast-storage-ng/pull/87

### DIFF
--- a/src/lib/installation/clients/inst_system_analysis.rb
+++ b/src/lib/installation/clients/inst_system_analysis.rb
@@ -27,10 +27,12 @@
 
 require "yast"
 require "fileutils"
+require "y2storage"
 
 module Yast
   class InstSystemAnalysisClient < Client
     include Yast::Logger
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/inst_system_analysis.rb
```diff
@@ -262,7 +288,7 @@ def ActionHDDProbe
         drivers_info = ""
       end
 
-      if Builtins.size(targetMap) == 0
+      if devicegraph.disks.empty?
```
aschnell: The Devicegraph also has a function empty?. That would be closer at the original code.

aschnell:
LGTM

PR #415, opened 2016-08-03
UI and UX Optimizations in Welcome Screen

kobliha:
- Code that handles license checkbox moved to prevent from warnings"No widget ID..."
- The License Text widget is now initialized ASAP
- Changing the License widget when its text is actually known
- The new code has been tested and optimized especially for the user (UI, UX), while keeping the actual openQA-related bug in mind as well (bsc#980374)
- The client has a good test case which does not need to be extended
- @license_accepted variable has been replaced by InstData.product_license_accepted because of conflicts in data duplication

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Aug  3 17:06:46 CEST 2016 - locilka@suse.com
+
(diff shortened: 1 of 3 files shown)


imobachgs:
Definitively it's an improvement, thanks! LGTM.

kobliha:
@imobachgs Thank you :)!

PR #416, opened 2016-08-05
Don't save custom_url when SCC is used during self-update

imobachgs:
Fixes a problem introduced [here](https://github.com/yast/yast-installation/pull/401/commits/e27cc92d5d5f45d48808b4e1fcb54499778e6c50#diff-8aa15970732082407af8655c128f0913R168).

The registration URL used during self-update is stored to be the default option in the registration screen. However, due to the change mentioned above, it was stored accidentally converted from `nil` to `""` if SCC was used.

This patch just skip writing the URL if not needed.

### DIFF
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -168,7 +168,7 @@ module Yast
       registration = Registration::Registration.new(url == :scc ? nil : url.to_s)
       # Set custom_url into installation options
       Registration::Storage::InstallationOptions.instance.custom_url = registration.url
-      store_registration_url(url) if url.is_a?(URI)
+      store_registration_url(url) unless url == :scc
       registration.get_updates_list.map { |u| URI(u.url) }
     end
 
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -168,7 +168,7 @@ module Yast
       registration = Registration::Registration.new(url == :scc ? nil : url.to_s)
       # Set custom_url into installation options
       Registration::Storage::InstallationOptions.instance.custom_url = registration.url
-      store_registration_url(registration.url)
+      store_registration_url(url) if url.is_a?(URI)
       registration.get_updates_list.map { |u| URI(u.url) }
     end
 
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -168,7 +168,7 @@ module Yast
       registration = Registration::Registration.new(url == :scc ? nil : url.to_s)
       # Set custom_url into installation options
       Registration::Storage::InstallationOptions.instance.custom_url = registration.url
-      store_registration_url(url) unless url == :scc
+      store_registration_url(url) if url != :scc
       registration.get_updates_list.map { |u| URI(u.url) }
     end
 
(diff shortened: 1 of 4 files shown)


test/inst_update_installer_test.rb
```diff
@@ -254,24 +254,30 @@ def import(_args); end
...
+                expect(registration_class).to receive(:new).with(smt0.slp_url)
+                  .and_return(registration)
+                allow(manager).to receive(:add_repository)
+                subject.main
+              end
+
+              it "saves the registration URL to be used later" do
+                allow(manager).to receive(:add_repository)
+                expect(FakeInstallationOptions.instance).to_not receive(:custom_url=).with(nil)
+                expect(File).to receive(:write).with(/inst_update_installer.yaml/,
```
lslezak: NP: I'd be more strict about the file name, e.g. `/\/inst_update_installer.yaml\z/`.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -168,7 +168,7 @@ def self_update_url_from_connect
       registration = Registration::Registration.new(url == :scc ? nil : url.to_s)
       # Set custom_url into installation options
       Registration::Storage::InstallationOptions.instance.custom_url = registration.url
-      store_registration_url(registration.url)
+      store_registration_url(url) if url.is_a?(URI)
```
lslezak: IMHO `if url != :scc` would be more readable and similar to the comparison 3 lines above (the same style).

With this code you have to figure out what else it can be if not an `URI`, the meaning is a bit hidden.

test/inst_update_installer_test.rb
```diff
@@ -291,12 +297,19 @@ def import(_args); end
...
+              it "asks the SCC server for the updates URLs" do
                 expect(registration_class).to receive(:new).with(nil)
                   .and_return(registration)
                 allow(manager).to receive(:add_repository)
                 subject.main
               end
+
+              it "does not save the registration URL to be used later" do
+                allow(manager).to receive(:add_repository)
+                expect(FakeInstallationOptions.instance).to_not receive(:custom_url=).with(nil)
```
lslezak: I guess in this case the `custom_url=` method should be never called regardless the argument. This expectation allows calling `custom_url=` with anything but `nil` which is wrong.

test/inst_update_installer_test.rb
```diff
@@ -254,24 +254,30 @@ def import(_args); end
...
+              it "asks that SMT server for the updates URLs" do
+                expect(registration_class).to receive(:new).with(smt0.slp_url)
+                  .and_return(registration)
+                allow(manager).to receive(:add_repository)
+                subject.main
+              end
+
+              it "saves the registration URL to be used later" do
+                allow(manager).to receive(:add_repository)
+                expect(FakeInstallationOptions.instance).to_not receive(:custom_url=).with(nil)
```
lslezak: Is this expectation correct? See https://github.com/yast/yast-installation/pull/416/files#r73657445

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -168,7 +168,7 @@ def self_update_url_from_connect
       registration = Registration::Registration.new(url == :scc ? nil : url.to_s)
       # Set custom_url into installation options
       Registration::Storage::InstallationOptions.instance.custom_url = registration.url
-      store_registration_url(registration.url)
+      store_registration_url(url) unless url == :scc
```
lslezak: NP: I'd rather prefer `if url != :scc` but that depends on personal style, I'm fine with this.

imobachgs: Yes, I don't have a strong opinion but I'm used to `unless`. Anyway, I've changed it :)

lslezak:
LGTM, thanks!  :tada:

imobachgs:
Thanks! Merging...

PR #417, opened 2016-08-09
Remove old libstorage dependency

ancorgs:
To make travis work again we need to be able to build and install some basic packages without bringing yast2-storage in. We don't need all the code to actually work, just the one executed in our installation (already fixed/commented in https://github.com/yast/yast-installation/pull/403).

This pull request removes the dependency from the spec file and cleanups some bits of code. Hopefully enough to get travis working again.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug  9 14:51:06 UTC 2016 - ancor@suse.com
+
+- storage-ng: removed dependency from (old) yast2-storage, even
+  if it breaks some functionality.
+- 3.2.2
+
+-------------------------------------------------------------------
 Fri Jul 29 07:32:46 UTC 2016 - ancor@suse.com
 
 - If the user has skipped multipath activation, don't ask again
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -62,9 +62,6 @@ Conflicts:	yast2-mouse < 2.18.0
 # New API for ProductLicense
 Requires:	yast2-packager >= 3.1.96
 
-# Yast::Storage.multipath_off?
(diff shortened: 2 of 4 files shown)


package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug  9 14:51:06 UTC 2016 - ancor@suse.com
+
+- storage-ng: removed dependency from (old) yast2-storage, even
+  if it breaks some functionality.
```
lslezak: If it breaks the functionality then I'd comment the lines instead of removing them and put a `FIXME` note with description what needs to be fixed later.

package/yast2-installation.spec
```diff
@@ -62,9 +62,6 @@ Conflicts:	yast2-mouse < 2.18.0
 # New API for ProductLicense
 Requires:	yast2-packager >= 3.1.96
 
-# Yast::Storage.multipath_off?
-Requires:	yast2-storage >= 3.1.97
```
lslezak: Should we rather comment this out instead of removing?

ancorgs:
Makes sense. I will change that.

package/yast2-installation.spec
```diff
@@ -62,8 +62,10 @@ Conflicts:	yast2-mouse < 2.18.0
...
+# FIXME: some code present in this package still depends on the old yast2-storage
+# and will break without this dependency. That's acceptable at this point of the
+# migration to storage-ng. See installer-hacks.md in the yast-storage-ng repo.
```
lslezak: NP: I'd put the link https://github.com/yast/yast-storage-ng/blob/master/doc/installer-hacks.md here, I'm too lazy to look for a doc... :wink:

lslezak:
LGTM

PR #418, opened 2016-08-11
[WIP] Move self-update and software manager initialization

imobachgs:
The intention of this PR is just to discuss what's the best approach to move the self-update feature earlier during the installation. I'm showing some code which works although I think that the final solution could be different.

My first instinct was just to move the `InitInstallationRepositories` to the beginning of this client and adding a new step:

```ruby
  actions_todo      << _("Installer update")
  actions_doing     << _("Searching for installer updates...")
  actions_functions << fun_ref(method(:update_installer), "boolean ()")
```

But at this point, the add-ons initialization (`add_on_products.xml`) are already processed and I would like to do the update in a even earlier step. So that's the reason to call the "client" right after the real initialization. Anyway, the code needs some refactoring (it looks ugly).

On the other hand, we should remove references to `update_installer` in control files and that leads me to another possible approach: maybe we could move the software initialization to its own client (including calling the self-update).

Finally, I don't know why the progress is messed up after the software initialization step finishes.

### DIFF
--- a/src/lib/installation/clients/inst_system_analysis.rb
+++ b/src/lib/installation/clients/inst_system_analysis.rb
@@ -96,6 +96,10 @@ module Yast
       actions_doing = []
       actions_functions = []
 
+      actions_todo      << _("Initialize software manager")
+      actions_doing     << _("Initializing software manager...")
+      actions_functions << fun_ref(method(:InitInstallationRepositories), "boolean ()")
+
       Builtins.y2milestone("Probing done: %1", Installation.probing_done)
       # skip part of probes as it doesn't change, but some parts (mostly disks
       # that can be activated) need rerun see BNC#865579
@@ -134,10 +138,6 @@ module Yast
       actions_todo      << _("Search for system files")
       actions_doing     << _("Searching for system files...")
       actions_functions << fun_ref(method(:FilesFromOlderSystems), "boolean ()")
-
-      actions_todo      << _("Initialize software manager")
-      actions_doing     << _("Initializing software manager...")
-      actions_functions << fun_ref(method(:InitInstallationRepositories), "boolean ()")
 
       Progress.New(
         # TRANSLATORS: dialog caption
@@ -339,6 +339,7 @@ module Yast
         ret = false
       else
         @packager_initialized = true
+        return :restart_yast if update_installer == :restart_yast
         Packages.InitializeAddOnProducts
 
         # bnc#886608: Adjusting product name (for &product; macro) right after we
@@ -352,6 +353,13 @@ module Yast
       PackageCallbacks.RestorePreviousProgressCallbacks
 
       ret
+    end
+
+    # Runs installer's self-update
+    #
+    # @see inst_update_installer client
+    def update_installer
+      WFM.CallFunction("inst_update_installer", [])
     end
 
     def FilesFromOlderSystems


imobachgs:
Another approach could be trying to initialize software manager at the beginning of the `update_installer` client (the minimum) and leave the rest as is.

lslezak:
IMHO the proposed change does not help much. I mean the self update step is moved just a bit in the workflow.

Originally the `system_analysis` step was just before the `update_installer` step. Moving it at the beginning of the `system_analysis` step is a too small move forward IMHO. Moreover the `system_analysis` step is not used in the *autoinstallation* mode (or at least is missing in `control.xml`, I'm not sure if it called from inside the AY clients).

I think we should be more aggresive with moving the self update.

### Proposal

- Keep the self update functionality in a separate client (clear separation, for moving it just edit the `control.xml` file, no need for any code change)
- Add the Packager initialization (`Packages.InitializeCatalogs`, plus maybe some fancy stuff with the callbacks...)
- Move it in the workflow as soon as possible, we basically just need working network. That means we should move it just after the `setup_dhcp` step [here](https://github.com/yast/skelcd-control-SLES/blob/41a7d61e49966473c3eb9587311d82f5685101d8/control/control.SLES.xml#L718).

#### Disadvantages

- It would mean that the manual network config won't be possible, static IP setup could be possible only via linuxrc boot parameters. But I think that's acceptable, I guess static IP setup is not that common.
- It would also mean that we cannot easily support iSCSI, DASD,... but yesterday we agreed that we would initially support only HTTP(S) and FTP protocols.

#### Advantages

- At the end (before restarting YaST) we would just remove the added repository to avoid repository duplicates.
- The rest should just work™:
  - I think we would not need to store/restore anything as there is no user interaction yet (no user preference to keep).
  - The steps looks idempotent to me (e.g. `configure_dhcp` step skips already configured cards), running them again should not have any side effects. We would just check for restart at the self update step to avoid infinite loop.

## Opinions?

imobachgs:
@lslezak It sounds like a good plan (it worth a try), but I think that we should keep it at least after the `complex_welcome` because of the language selection. There is little user interaction during self-update (ideally no interaction at all), but certificates importing or SMT selection could happen. And at that point we should know the user's language.

teclator:
I like the option of a separate client.

@lslezak @imobachgs regarding to AutoYast, currently autoinit calls `system_analysis`, `iscsi-client-auto` and `fcoe-client-auto` just to take it in account and in autoupgrade it is part of the workflow but disable if autoyast options is passed or not, more info also [here](https://github.com/yast/skelcd-control-SLES/pull/65) and the root cause of this [bug](https://bugzilla.suse.com/show_bug.cgi?id=988287)

If i'm not wrong initially we put `self_update` before `system_analysis` but then we move it after `system_analysis` due to repositories initialization, so move this part or add it to a new client sounds fine.

So if we don't need to call disks activation because DASD,FCOE and ISCI will not be supported by self_update then we could call it just after `lan` so also the network could be configured if not obtained by dhcp or just add `lan` client into the new **client** proposed.

lslezak:
Yes, I was thinking about the language as well...

IMHO selecting the language later is OK mainly because:

- The self update should work automatically without any user interaction.
- The interaction will be needed only in special minor cases (mainly with a custom repo -  unknown SSL certificate, unsigned repository, unsigned RPMs...).
- There already is some *"Initializing"* dialog at the beginning in English (untranslated) and the language selection is initially in English. Displaying some English text, like download progress, should not be surprising for users.
- If the users really want to have translations early then the initial language can be selected in the DVD boot menu (using the `F2` key) or via the [`language`](https://en.opensuse.org/SDB:Linuxrc#p_lang) boot option.

Moreover the sooner we apply the self update the bigger possibility to fix a bug. It will minimize the cases when the self update cannot fix a bug (then only DUD will help).

imobachgs:
@lslezak Yes, you're right, I didn't think about the `F2` option. And yes, the self-update will need interaction only in quite specific cases. And of course, I agree with moving it as soon as possible so we can fix any installer bug using self-update :)

lslezak:
JFYI: Currently we do not support self update from disconnected machine, the documentation says [*If network is not available, the installer update will be skipped.*](https://github.com/yast/yast-installation/blob/master/doc/SELF_UPDATE.md#error-handling).

That means using a local disk for self update is even more difficult => one more reason for self update to officially support only HTTP(S)/FTP, at least at the beginning.

PR #419, opened 2016-08-16
Fixed test-case after implementing lazy loading in ProductLicense

kobliha:
- The test actually tests behavior of the ProductLicense (a bit) instead of testing the client only
- That's something I can't and will not fix now
- See also https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:J/yast2-installation/standard/x86_64

### DIFF
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -34,6 +34,7 @@ Source1:	YaST2-Second-Stage.service
 Source2:	YaST2-Firstboot.service
 
 BuildRequires:  docbook-xsl-stylesheets libxslt update-desktop-files yast2-core-devel
+BuildRequires:  yast2-packager >= 3.1.113
 BuildRequires:  yast2-devtools >= 3.1.10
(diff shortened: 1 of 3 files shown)


package/yast2-installation.spec
```diff
@@ -59,8 +60,8 @@ Requires:	yast2-pkg-bindings >= 3.1.33
 # Mouse-related scripts moved to yast2-mouse
 Conflicts:	yast2-mouse < 2.18.0
 
-# New API for ProductLicense
-Requires:	yast2-packager >= 3.1.96
+# Lazy loading in ProductLicense
+Conflicts:	yast2-packager < 3.1.113
```
mchf: it still ```Requires: yast2-packager``` (not only ```BuildRequires```) doesn't it?

kobliha: No, this is for installation, the rest of packages is not required either

kobliha: Oh, yes

mchf:
lgtm

PR #420, opened 2016-08-17
Twice url fix

jreidinger:
manual testing in process now.

### DIFF
--- a/src/lib/installation/clients/inst_extrasources.rb
+++ b/src/lib/installation/clients/inst_extrasources.rb
@@ -342,13 +342,9 @@ module Yast
         general = Pkg.SourceGeneralData(src)
         url = general["url"]
         next if url.nil? || url.empty?
-        if Mode.update && url =~ /^dir:\/+mnt\/+/
-          @local_urls[src] = url
-        end
+        @local_urls[src] = url if Mode.update && url =~ /^dir:\/+mnt\/+/
         # check for USB sources which should be disabled
-        if url.include?("device=/dev/disk/by-id/usb-")
-          @usb_sources[src] = url
-        end
+        @usb_sources[src] = url if url.include?("device=/dev/disk/by-id/usb-")
         url.chop! if url.end_with?("/") # remove trailing slash to normalize path (see bnc#970488)
         ret << url
       end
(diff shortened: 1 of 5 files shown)


src/lib/installation/clients/inst_extrasources.rb
```diff
@@ -229,7 +223,9 @@ def GetURLsToRegister(registered)
       end
 
       urls_from_control_file = Builtins.filter(urls_from_control_file) do |one_url|
-        if Builtins.contains(registered, Ops.get_string(one_url, "baseurl", ""))
+        url = one_url["base_url"] || ""
+        normalized_url = url.end_with?("/") ? url.chop : url
```
lslezak: I think we need to be more careful here, the trailing `/` might be important in some cases, like in `cd:/`.

It seems that we have to parse the URL and check if the trailing `/` is there. Then remove it and rebuild the URL again...

jreidinger: this normalized url is used only for comparison, for other purpose and later processing original is used

lslezak: Yes, I overlooked that, looks OK...

shundhammer:
This was there before this pull request, but it should be fixed before this is distributed among 80+ YaST subprojects:

In the "Bug Reports" section, "if you find a problem..." is duplicated.

shundhammer:
Other than that, LGTM.

aschnell:
Note: Technical the URLs only differing in a trailing slash are not identical. I once had the problem with some openSUSE repository that one of the URLs was redirected while the other one was not. On the other hand AFAIR libzypp cannot distinguish the two URLs.

lslezak:
Libzypp does not care about the URLs at all, you can even add a repository with the same URL as many times as you wish. The only result would be increased memory consumption because of the loaded duplicates and probably slower dependency solving.

jreidinger:
thinking about slash and different url: Even when url with and without trailing slash are technically different, I cannot image case when someone intentionally use different url in extra_urls in control file and in online repo and want to have it as two separated repositories, so I propose to ignore it.

lslezak:
I think that for handling the `control.xml` and the openSUSE repositories it is good enough. It does not fixes all possible issues but at least fixes the original bug. If there are more issues we can fix them later.

So I'm giving my LGTM here...

PR #421, opened 2016-08-18
Proposal store fix

jreidinger:
[bsc#994127](https://bugzilla.suse.com/show_bug.cgi?id=994127)

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug 18 11:52:39 UTC 2016 - jreidinger@suse.com
+
+- fix skipping of proposal returning empty hash otherwise it
+  creates non sense proposal entry in UI (bnc#994127)
+- 3.1.211
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


mvidner:
Importantly, Suresh said in the bug that an analogous fix (changing his clients to return `nil`) has fixed the problem.

jreidinger:
@mvidner yep, this fix just ensure that doc and code is in sync, so both ways can be used.

package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug 18 11:52:39 UTC 2016 - jreidinger@suse.com
+
+- fix skipping of proposal returning empty hash otherwise it
+  creates non sense proposal entry in UI (bnc#994127)
```
mvidner: I am not sure people will understand that without context. What about this:

> proposal entries were always displayed (with a  foo_proposal heading) even if empty hash was returned (bnc#994127)

mvidner:
I have changed the target to SP2 and made it mergeable.

mvidner:
LGTM

jreidinger:
@mvidner thanks

PR #423, opened 2016-08-19
Remove fixes for update installer not needed anymore.

teclator:
Tested update_installer manually in a normal installation and with it just after inst_dhcp client in the control.xml file.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Aug 19 11:15:21 UTC 2016 - kanderssen@suse.com
+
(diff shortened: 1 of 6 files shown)


package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Aug 19 11:15:21 UTC 2016 - kanderssen@suse.com
+
+- Removing remember state fix related to update_installre from
```
lslezak: This sounds very cryptic even for the YaST developers, rather use some description suitable for end users.

lslezak:
I'll merge the PR and update the changes.

PR #424, opened 2016-08-23
[WIP] Fetch profile from update installer

teclator:
I tried to do as minimal as possible not handling all the stuff that currently autoinit is handling, just enough to read the self_update_url.

So from here https://github.com/yast/yast-autoinstallation/blob/master/src/clients/inst_autoinit.rb#L196 to the end is not handled.

### DIFF
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -611,12 +611,16 @@ module Yast
       return current_profile if current_profile
 
       if !profile_valid_scheme?
-        log.info("The scheme used (#{AutoinstConfig.scheme}), is not supported in self update.")
+        Report.Warning("The scheme used (#{AutoinstConfig.scheme}), " \
+                       "is not supported in self update.")
         return nil
       end
 
-      log.info("Processing profile location...")
-      ProfileLocation.Process
+      Report.LogMessages(true)
+      Report.LogErrors(true)
+      Report.LogWarnings(true)
+
+      process_location
 
       if !current_profile
         secure_uri = Yast::URL.HidePassword(AutoinstConfig.OriginalURI)
@@ -625,12 +629,8 @@ module Yast
         return nil
       end
 
-      Report.LogMessages(true)
-      Report.LogErrors(true)
-      Report.LogWarnings(true)
-
       if !Profile.ReadXML(AutoinstConfig.xml_tmpfile)
-        Popup.Warning(_("Error while parsing the control file.\n\n"))
+        Report.Warning(_("Error while parsing the control file.\n\n"))
         return nil
       end
 
@@ -663,6 +663,24 @@ module Yast
 
       profile_prepare_reports
       profile_prepare_signatures
+    end
+
+    # It retrieves the profile and the user rules from the given location not
+    # blocking AutoYast
+    #
+    # @see ProfileLocation.Process
+    def process_location
+      # ProfileLocation reports an error in case that the profile was not
+      # available in the given URL. We change the timeout error to not block
+      # AutoYast during update_installer, just to this method.
+      report_settings = Report.Export
+
+      Report.Import(report_settings.merge("errors" => { "timeout" => 10 }))
+
+      log.info("Processing profile location...")
+      ProfileLocation.Process
+
+      Report.Import(report_settings)
     end
   end
 end
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -24,6 +24,7 @@ module Yast
 
     UPDATED_FLAG_FILENAME = "installer_updated".freeze
     REMOTE_SCHEMES = ["http", "https", "ftp", "tftp", "sftp", "nfs", "nfs4", "cifs", "smb"].freeze
+    PROFILE_VALID_SCHEMES   = ["http", "https", "ftp", "tftp", "sftp"].freeze
     REGISTRATION_DATA_PATH = "/var/lib/YaST2/inst_update_installer.yaml".freeze
 
     Yast.import "Pkg"
@@ -42,6 +43,9 @@ module Yast
     Yast.import "NetworkService"
     Yast.import "Mode"
     Yast.import "Profile"
+    Yast.import "ProfileLocation"
+    Yast.import "AutoinstConfig"
+
 
     def main
       textdomain "installation"
@@ -57,6 +61,13 @@ module Yast
       return :next if installer_updated?
 
       initialize_progress
+
+      if Mode.auto
+        log.info("Fetching profile")
+
+        return :next unless fetch_profile
+        Yast::Progress.NextStage
+      end
 
       # initialize packager, we need to load the base product name
       # to properly obtain the update URL from the registration server
@@ -527,6 +538,16 @@ module Yast
 
     # Show global self update progress
     def initialize_progress
+      stages = [
+        # TRANSLATORS: progress label
+        _("Add Update Repository"),
+        _("Download the Packages"),
+        _("Apply the Packages"),
+        _("Restart")
+      ]
+
+      stages.unshift(_("Fetching AutoYast Profile")) if Mode.auto && profile_given?
+
       # open a new wizard dialog with title on the top
       # (the default dialog with title on the left looks ugly with the
       # Progress dialog)
@@ -541,13 +562,7 @@ module Yast
       # max is 100%
       100,
       # stages
-      [
-        # TRANSLATORS: progress label
-        _("Add Update Repository"),
-        _("Download the Packages"),
-        _("Apply the Packages"),
-        _("Restart")
-      ],
+      stages,
       # steps
       [],
       # help text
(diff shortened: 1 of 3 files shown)


lslezak:
We actually need to additionally set the Report config and the signature callbacks.

The signature callbacks handle the GPG errors reported by libzypp, the profile allows e.g. to disable GPG checks if you use own unsigned installer updates or signed by your (non-SUSE) GPG key.

The Report can be configured to ignore errors and continue after a timeout. Again, when using your own installer updates this might be necessary.

The original AY code does not look nice (see [here](https://github.com/yast/yast-autoinstallation/blob/master/src/clients/inst_autoinit.rb#L201)):

```ruby
report = Profile.current["report"]
  if report && !report.has_key?( "yesno_messages" )
    report = Report.Export
    report["yesno_messages"] = report.fetch("errors",{})
  end
Report.Import(report)
```

I'd just copy&paste that so it behaves the same way (to be compatible).

For signature handling we would just need to call:
```ruby
AutoinstGeneral.SetSignatureHandling
```

However, before this the original AY code calls:
```ruby
AutoinstGeneral.Import(Profile.current.fetch("general",{}))
```

I'm not sure what it does and if that could influence the `AutoinstGeneral.SetSignatureHandling` call later. Please check, maybe it is needed, maybe not...

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -24,6 +24,7 @@ class InstUpdateInstaller
 
     UPDATED_FLAG_FILENAME = "installer_updated".freeze
     REMOTE_SCHEMES = ["http", "https", "ftp", "tftp", "sftp", "nfs", "nfs4", "cifs", "smb"].freeze
+    PROFILE_VALID_SCHEMES   = ["http", "https", "ftp", "tftp", "sftp"].freeze
```
lslezak: [get_file_from_url method](https://github.com/yast/yast-installation/blob/self_update_refactoring/src/lib/transfer/file_from_url.rb#L810) supports much more protocols, why is it limited to only these?

Of course, some protocols might not work properly in all cases (e.g. the disk protocol), but we should at least try it and do not give up too early.

teclator: True, i was influenced by this requirement:

- Support only HTTP, HTTPS and FTP self-update sources

But it was about self_update sources and not about autoyast profile, so i will update tit.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -565,5 +580,68 @@ def finish_progress
...
+
+    def fetch_profile
+      return current_profile if current_profile
+
+      if !profile_valid_scheme?
+        Report.Error(format(_("Not supported scheme %s.\n\n"), Yast::AutoinstConfig.scheme))
+        return nil
+      end
+
+      Yast::ProfileLocation.Process
```
lslezak: I was worried whether the AutoYaST rules are evaluated, but this call seems to handle that... :white_check_mark:

teclator: Yes, it does, basically we are trying to do the same that processProfile is doing removing all the parts that could be removed (but basically at the end maybe only the storage part), so maybe some part could be moved to the ProfileLocation or Profile modules and import them from both clients (autoinit and update_installer).

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -565,5 +580,68 @@ def finish_progress
...
+
+      while !current_profile
```
lslezak: I think asking for the profile when download fails does not make much sense in the self update.

The profile might be located on a disk which is not accessible yet (the drivers are loaded by `inst_system_analysis` in full AutoYaST [here](https://github.com/yast/yast-autoinstallation/blob/master/src/clients/inst_autoinit.rb#L89)). So maybe the self update is not able to load the profile but the full AutoYaST can do it later. And for self update we support only HTTP(S)/FTP anyway. 

So when we fail here we should skip loading the profile and fallback to the default self update URL. The user can always pass the self update URL or disable it completely via the boot options when using unsupported profile location. We just need to document this.

We should not block AutoYaST when self update cannot load the profile. Let's start with a simple case first, we can always improve it later when we get more feedback from the users...

teclator: Totally agree. 

As commented above in the case that only network protocols were to be used the profile should be accessible, and we could ask for the URI, but as you commented could be located on a cdrom or something like that.

About the disk I was not worried because in that case the plan was to just to continue without fetch the profile because as you have mentioned the disks will not be accessible until ```inst_system_analysis```.

In summary, by now we are going to process the profile location without block AutoYast using the default in case of not fetched, and avoiding just the ```AutoinstConfig.scheme == label`` that implies Storage as we can see [here](https://github.com/yast/yast-autoinstallation/blob/master/src/modules/ProfileLocation.rb#L126)

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -564,5 +578,57 @@ def finish_progress
...
+
+      unless profile_valid_scheme?
+        log.info("The scheme used (#{AutoinstConfig.scheme}), is not supported in self update.")
+        return nil
+      end
+
+      log.info("Processing profile location...")
+      ProfileLocation.Process
+
+      unless current_profile
```
lslezak: NP: Use `if !current_profile` here to be consistent.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -564,5 +578,57 @@ def finish_progress
...
+        log.info("Unable to load the profile from: #{AutoinstConfig.OriginalURI}")
```
lslezak: The URL can possibly contain user name and a password, use `Yast::URL.HidePassword` to hide it when logging.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -564,5 +578,57 @@ def finish_progress
...
+      !PROFILE_FORBIDDEN_SCHEMES.include? AutoinstConfig.scheme
+    end
+
+    def current_profile
+      return nil if Yast::Profile.current == {}
+
+      Profile.current
+    end
+
+    def fetch_profile
```
lslezak: Ideally this should be shared with the [inst_autoinit.rb client](https://github.com/yast/yast-autoinstallation/blob/master/src/clients/inst_autoinit.rb#L138), add at least a `TODO` note and mention where it was copied from so we can improve it later.

teclator: I also share that point of view, in the future we should move the responsibility of profile fetching to a reusable class to be more DRY and to not duplicate efforts.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -564,5 +577,92 @@ def finish_progress
...
+        return nil
+      end
+
+      log.info("Processing profile location...")
+      ProfileLocation.Process
+
+      if !current_profile
+        secure_uri = Yast::URL.HidePassword(AutoinstConfig.OriginalURI)
+        log.info("Unable to load the profile from: #{secure_uri}")
+
```
lslezak: I think we should display a warning message here saying that the profile could not be downloaded to get the installer update URL and that the default will be used.

We should use `Popup.TimedWarning` to not block the autoinstallation. (The Report module is not initialized yet.)

teclator: It is what I was looking for, I tried with Popup.Warning thinking that it uses the timeout defaults and in that case 10 seconds, but don't seems so.

I'm also preparing the methods so that can be easier to move into a specialized class in the future.

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -564,5 +577,110 @@ def finish_progress
...
+    # blocking AutoYast
+    #
+    # @see ProfileLocation.Process
+    def process_location
+      # ProfileLocation reports an error in case that the profile was not
+      # available in the given URL. We change the timeout error to not block
+      # AutoYast during update_installer, just to this method.
+      report_settings = Report.Export
+
+      Report.Import(report_settings.merge("errors" => { "timeout" => 10 }))
```
lslezak: If the self update is skipped in the end (e.g. because the URL points to an empty repository) then as a side effect of this code the error setting will be kept for the AY installation later.

We should set the defaults back after the self update client is finished to avoid side effects and possible regressions. (E.g. run `settings = Report.Export` at the beginning,  `Report.Import(settings)` at the end.)

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -564,5 +577,110 @@ def finish_progress
...
+
+      log.info("Processing profile location...")
+      ProfileLocation.Process
+
+      Report.Import(report_settings)
```
lslezak: Yes, it is restored back here, but when the self-update fails e.g. during applying the updates then `Report.Error` will block again. I'm not sure if that's desired...??

src/lib/installation/clients/inst_update_installer.rb
teclator: If it is very common, we could add a method that receive a block and just do the ```settings = Report.Export``` and ```Report.Import(settings)``` around.

src/lib/installation/clients/inst_update_installer.rb
imobachgs: The downside is that reports settings will be ignored. I'm fine with that as long as it's documented.

imobachgs:
This client is getting too big. Don't get me wrong: it was getting big even before moving it to the beginning of the installation (after adding the registration stuff). But now we also have code related to package management and AutoYaST profile handling.

We're almost on RC2 and I wouldn't refactor it right now. But maybe after RC2 is released we could have a look.

lslezak:
Yes, already mentioned at https://github.com/yast/yast-installation/pull/424#discussion_r76035084, the code is not ideal but for RC2 I would prefer minimal changes to the other packages. Let's move/share the code later...

src/lib/installation/clients/inst_update_installer.rb
lslezak: I have checked the AY code and it seems that until the `Report` settings are loaded from the profile  it uses the defaults (which block). So let's use the same approach here, if the profile cannot be loaded keep the defaults for the rest of the code.

lslezak:
LGTM

imobachgs:
LGTM

PR #425, opened 2016-08-24
add more logging to debug (bnc#994127)

jreidinger:
it do not deserve own changelog entry as it is just more debugging, so fine if it is send later to BS

### DIFF
--- a/src/lib/installation/proposal_runner.rb
+++ b/src/lib/installation/proposal_runner.rb
@@ -261,8 +261,9 @@ module Installation
 
     def pre_continue_handling
       @skip = if Yast::UI.WidgetExists(Id(:skip))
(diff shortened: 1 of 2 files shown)


src/lib/installation/proposal_runner.rb
```diff
@@ -261,6 +261,7 @@ def handle_skip
 
     def pre_continue_handling
       @skip = if Yast::UI.WidgetExists(Id(:skip))
+        log.info "there is skip widget with value #{Yast::UI.QueryWidget(Id(:skip), :Value)}."
```
lslezak: NP: maybe a bit more DRYer (+ use `inspect`):
```ruby
val = Yast::UI.QueryWidget(Id(:skip), :Value)
log.info "there is :skip widget with value #{val.inspect}"
val
```

lslezak:
LGTM

PR #426, opened 2016-08-24
Replace unicode bullet char by asterisk for ncurses (bsc#995082)

cwh42:
More like a hack, but it fixes the issue without big overhead.
roles-ncurses
roles-qt

### DIFF
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 # Copyright (c) 2016 SUSE LLC.
 #  All Rights Reserved.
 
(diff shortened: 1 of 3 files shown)


src/lib/installation/select_system_role.rb
```diff
@@ -1,3 +1,4 @@
+# coding: utf-8
```
lslezak: JFYI: UTF-8 is default for some time (since Ruby 2.0?), it's not needed anymore. On the other hand it does not harm...

lslezak:
LGTM, please resolve the conflict

lslezak:
Closing, merged in #428.

PR #427, opened 2016-08-25
Fix writing proposal

jreidinger:


imobachgs:
I'm missing a testcase. Otherwise, LGTM.

jreidinger:
will try to create one, but it is not trivial piece of code. Suresh test it manually.

imobachgs:
@jreidinger Ok, so let's try it manually (as the fix looks trivial). LGTM.

jreidinger:
after quick look it is too complex to test without another significant refactoring. We testing proposal_store, but proposal_runner is still too much coupled with a lot of other stuff.

kobliha:
@jreidinger Maybe openQA test can help? We need to prevent from happening this again...

jreidinger:
@kobliha we do not use this functionality, it is used by OES, so we can test only them

src/lib/installation/proposal_runner.rb
```diff
@@ -514,7 +514,7 @@ def write_settings
 
       log.info "Writting settings for proposal"
 
-      @store.proposal_names do |submod|
+      @store.proposal_names.each do |submod|
```
mvidner: This bug could be detected by static analysis if we were able to resolve the `proposal_names` method because it is already [documented to take no arguments](https://github.com/yast/yast-installation/blob/031cc091f08f5d6be8794400514ad940857dd72e/src/lib/installation/proposal_store.rb#L129) (while we were passing it a block).
Unfortunately the class of `@store` is determined at run time ([`@store = @store_class.new(...)`](https://github.com/yast/yast-installation/blob/031cc091f08f5d6be8794400514ad940857dd72e/src/lib/installation/proposal_runner.rb#L84)) so I am afraid this is too advanced for our tools.

PR #428, opened 2016-08-25
Resolve conflicts in PR#426

lslezak:
#426 cannot be merged because of a conflict, this is just a copy of #426 with merged changes from `master`.

### DIFF
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.1.213
+Version:        3.1.212
 Release:        0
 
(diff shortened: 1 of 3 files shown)


package/yast2-installation.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.1.211
+Version:        3.1.213
```
imobachgs: Version number does not match the one included in the changes file.

lslezak: Ooops, I probably called `rake version:bump` twice by mistake...

Good catch, thanks!

imobachgs:
LGTM

PR #429, opened 2016-08-26
Storage self update cleanup

imobachgs:


lslezak:
LGTM, but someone from the storage team should double check...

ancorgs:
LGTM

PR #431, opened 2016-08-29
Display a warning when using the fallback URL for self update (bsc#996179)

lslezak:
## Description

When the registration server fails or returns empty URL then the fallback URL from `control.xml` was silently used. This might be a bit misleading because users might think the SMT server is still used after confirming the error popup. It's not clear which URL is used in the end.

This fixes [bug#996179](https://bugzilla.suse.com/show_bug.cgi?id=996179).

## Original Behavior

*Click the screenshots to display them in full size.*

Note the default `https://updates.suse.com/` URL in the progress after confirming the error. If you do not look carefully you might still think the originally selected SMT is used, that's not true.

self_update_fallback

## Fixed

There is a popup warning displayed when the fallback URL is used.

*(The SSL import dialog is not displayed because this was tested it in the same installation instance, the certificate has been already imported.)*

self_update_fallback_popup

## Notes

- The `Report.LongWarning` call is used to print the message in a RichText widget to wrap long lines in ncurses UI: 
warning_ncurses
-  `Report.LongWarning` does not block in AutoYaST mode, the default URL is used automatically after a timeout.
- Additionaly the registration server is not contacted when network is not configured. In that case whole self-update is skipped.

### DIFF
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -210,7 +210,24 @@ module Yast
       # Set custom_url into installation options
       Registration::Storage::InstallationOptions.instance.custom_url = registration.url
       store_registration_url(url) if url != :scc
-      registration.get_updates_list.map { |u| URI(u.url) }
+      ret = registration.get_updates_list.map { |u| URI(u.url) }
+
+      if ret.empty?
+        # TRANSLATORS: error message
+        msg = _("<p>Cannot obtain the installer update repository URL\n" \
+          "from the registration server.</p>")
+
+        if self_update_url_from_control
+          # TRANSLATORS: part of an error message, %s is the default repository
+          # URL from control.xml
+          msg += _("<p>The default URL %s will be used.<p>") % self_update_url_from_control
+        end
+
+        # display the message in a RichText widget to wrap long lines
+        Report.LongWarning(msg)
+      end
+
+      ret
     end
 
     # Return the URL of the preferred registration server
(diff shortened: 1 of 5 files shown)


src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -210,7 +210,24 @@ def self_update_url_from_connect
       # Set custom_url into installation options
       Registration::Storage::InstallationOptions.instance.custom_url = registration.url
       store_registration_url(url) if url != :scc
-      registration.get_updates_list.map { |u| URI(u.url) }
+      ret = registration.get_updates_list.map { |u| URI(u.url) }
+
+      if ret.empty?
```
imobachgs: I would move this to a separate method to keep the original one rather small. What about something like `try_fallback_url` (sure you find a better name).

imobachgs:
LGTM

teclator:
LGTM, thnx for nice gifs and screenshots, and also for update the documentation ;)

imobachgs:
LGTM, thanks!

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -415,17 +437,17 @@ def configure_network?(reason)
...
+    # * Self-update feature is enabled and the repository URL is defined
     #
     # @return [Boolean] true if the update should be performed; false otherwise.
     #
     # @see #installer_updated?
     # @see #self_update_enabled?
     # @see NetworkService.isNetworkRunning
     def try_to_update?
-      !installer_updated? && self_update_enabled? && NetworkService.isNetworkRunning
+      NetworkService.isNetworkRunning && !installer_updated? && self_update_enabled?
```
imobachgs: Strictly speaking, I'm not sure if we need this as we check it on L61.

lslezak: Yes, not really needed, but I fixed the order so it does not break later after some changes or refactoring. Just to be on the safe side...

teclator:
LGTM

lslezak:
Thanks, merged! :tada:

PR #434, opened 2016-09-06
Fix version number in changes file

imobachgs:
Minor fix to changes file to match package's version.


mvidner:
LGTM. I have checked that the spec file has the correct version.

imobachgs:
Thanks!

PR #443, opened 2016-10-06
setting popup timeout in autoyast upgrade mode

schubi2:

PR #441, opened 2016-09-29
Translate snapper snapshots

imobachgs:
Translate description of Snapper snapshots according to the [solution proposed](https://bugzilla.suse.com/show_bug.cgi?id=988700#c5) by @mvidner to bsc#988700.

PR #437, opened 2016-09-21
bnc_999953; fixed crash in showing proposal

schubi2:

PR #444, opened 2016-10-06
[Review] Request from 'schubi2' @ 'yast/yast-installation/review_161006_setting_popup_timeout_in_autoyast_upgrade'

schubi2:
  * cc3f179 merged with merge_after_SP2_release
  * 151ad12 setting popup timeout in autoyast upgrade mode (#443)
  * 5d016df Translate snapper snapshots (#441)
  * e3637b7 bnc_999953; fixed crash in showing proposal (#437)

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,15 +1,16 @@
 -------------------------------------------------------------------
 Thu Oct  6 12:55:58 CEST 2016 - schubi@suse.de
 
-- AutoYaST upgrade: Setting timeout for errors, messages, ...
-  popup correctly (bnc#999895).
-- 3.2.0
+- AutoYaST upgrade: Do not override the Report module settings in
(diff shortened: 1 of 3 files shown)


package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Oct  6 12:55:58 CEST 2016 - schubi@suse.de
+
+- AutoYaST upgrade: Setting timeout for errors, messages, ...
+  popup correctly (bnc#999895).
```
lslezak: Well, the description does not match the code change well. You removed some code, the there is nothing set. It keeps the defaults set somewhere else.

The better description IMHO would be something like  `Do not override the Report module setting in the AutoYaST upgrade mode, keep the defaults`.

lslezak: or `keep the previous settings`.

PR #445, opened 2016-10-06
Set theme for ncurses

imobachgs:
Set theme for ncurses in order to honor the `Y2COLORMODE` setting. I've decided to add a `set_inst_ncurses_env` in make it symmetric to the Qt scenario.

Depends on https://github.com/yast/yast-yast2/pull/501 and it's related to https://github.com/libyui/libyui-qt/pull/60 and https://github.com/libyui/libyui-qt/pull/61.

### DIFF
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.2.3
+Version:        3.2.4
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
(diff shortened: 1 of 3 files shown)


ancorgs:
Shouldn't we update the version in the yast2 dependency?

package/yast2-installation.spec
```diff
@@ -46,8 +46,8 @@ BuildRequires: yast2 >= 3.1.180
 # AutoinstSoftware.SavePackageSelection()
 Requires:       autoyast2-installation >= 3.1.105
 
-# Moved proc_modules.scr
-Requires:       yast2 >= 3.1.186
+# Added Screenmode support
```
ancorgs: I would prefer a more "grep-able" comment. Something mentioning set_inst_ncurses_env

ancorgs:
LGTM

imobachgs:
Thanks!

PR #446, opened 2016-10-13
Fix version number in spec file

imobachgs:
Latest version in changes file is 3.2.3. I guess some merge went wrong :)


ancorgs:
If no automatic checker complains, LGTM

imobachgs:
`rake osc:build` works. Let's see if it also works on OBS/IBS.

PR #452, opened 2016-10-31
Support for read-only proposal modules (FATE#321739)

lslezak:
This implements the support for the read-only proposal modules, [FATE#321739](https://fate.suse.com/321739).

# Manual Testing

## Control File Example

The code has been tested with software and firewall proposal modules set to read-only mode:

```xml
<proposal_module>
  <name>software</name>
  <presentation_order>35</presentation_order>
  <read_only config:type="boolean">true</read_only>
</proposal_module>
<proposal_module>
  <name>firewall_stage1</name>
  <presentation_order>50</presentation_order>
  <read_only config:type="boolean">true</read_only>
</proposal_module>
```

## The Displayed Proposal

The software and firewall proposal modules then do not display the links in the section header (note the different color), the modules are not reachable:

read_only_proposal

However the links returned in the proposal summary are still displayed. But clicking them results in a warning popup and the click is ignored:

read_only_proposal_link

### Text Mode

The same situation in text mode, the title cannot be activated:

read_only_proposal_text

And the activating the links also displays the warning popup:
read_only_proposal_link_text

Additionaly in the text mode the *Change...* menu button does not contain the read-only modules, the software and the firewall setting are omitted:

read_only_proposal_menu_text

## Disadvantages of the Current Implementation

- Users cannot enter the read-only proposals to see the details. e.g. if the package proposal is read-only the user cannot see which exact packages will be installed.
- Some modules might ask the user to run the module (e.g. when the software selection is too big and does not fit the available space it asks for deselecting some packages), but that's not possible.
- Clicking the links in the proposal does not look nice (they should be rather hidden instead of displaying the warning popup after clicking them).

## Future Enhancements

The proposal handler now sends a new `read_only` flag to the proposal clients.

We could use this feature later to improve the disadvantages mentioned above, the clients would know that they are running in the read-only mode and should not return clickable links and adjust the displayed text accordingly.

### DIFF
--- a/src/lib/installation/proposal_store.rb
+++ b/src/lib/installation/proposal_store.rb
@@ -40,6 +40,7 @@ module Installation
       Yast.import "Mode"
       Yast.import "ProductControl"
       Yast.import "Stage"
+      Yast.import "Report"
 
       textdomain "installation"
 
@@ -237,9 +238,16 @@ module Installation
     def title_for(client)
       description = description_for(client)
 
-      description["rich_text_title"] ||
+      title = description["rich_text_title"] ||
         description["rich_text_raw_title"] ||
         client
+
+      return title unless read_only?(client)
+
+      # remove any HTML links if the proposal is read only,
+      # use the non-greedy .*? repetition to handle
+      # the "<a>foo</a> <a>bar</a>" case correctly
+      title.gsub(/<a.*?>(.*?)<\/a>/, "\\1")
     end
 
     # Returns the read-only flag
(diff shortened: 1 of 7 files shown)


src/lib/installation/proposal_store.rb
```diff
@@ -273,6 +297,24 @@ def client_for_link(link)
       matching_client.first
     end
 
+    def read_only_proposals
+      return @read_only_proposals unless @read_only_proposals.nil?
+
+      @read_only_proposals = []
+      log.info "all proposals: #{properties["proposal_modules"]}"
```
jreidinger: is it needed to be repeated? I think it is already in log

src/lib/installation/proposal_store.rb
```diff
@@ -273,6 +297,24 @@ def client_for_link(link)
...
+
+      properties.fetch("proposal_modules", []).each do |proposal|
+        next unless proposal["read_only"]
+
+        name = proposal["name"]
+        name += "_proposal" unless name.end_with?("_proposal")
```
jreidinger: I think it is already in multiple place, so maybe time for some helper?

jreidinger: with helper it can look like `@read_only_proposals << client_name(proposal["name"])`

jreidinger:
few ideas how to improve it, but in general it looks good

jreidinger:
LGTM

PR #454, opened 2016-11-02
Updated the control file documentation

lslezak:
- Updated the proposal part
- Removed the Automatic Configuration part (not used anymore, the seconds stage has been dropped completely)
- Updated self-update documentation

For better review you can check the [rendered version](https://github.com/yast/yast-installation/blob/012e83131b4a1e129140c253aae2f5ca5af21b81/doc/control-file.md#proposals).

### DIFF
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -1101,7 +1101,7 @@ repository should be defined in the control file.
     <self_update_url>http://updates.opensuse.org/$arch/leap-42.1-installer-update</self_update_url>
 ```
 
-This is the fallback which is used when the self-update repository is not\
+This is the fallback which is used when the self-update repository is not
 specified on the boot command line or when there registration module is not available.
 
 See more details in the [self-update documentation](./SELF_UPDATE.md).
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -312,7 +312,7 @@ Details:
   (the code actually allows writing it but that makes the definitions longer
   and less readable).
 - `presentation_order` (integer, optional) - the number defines the display order
-  in the proposal, the absolute value is not important, important is the relative
+  in the proposal, the absolute value is not important, it depends on the relative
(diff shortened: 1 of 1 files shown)


doc/control-file.md
```diff
@@ -268,30 +275,47 @@ possible using the control file.
...
-Currently, proposals names and captions are fixed and cannot be changed. It
-is not possible to create a special proposal screen, instead those
-available should be used: *network*, *hardware*, *service*. All proposal script
-names are listed without *_proposal* suffix. If a *proposed_module* is called
-*example*, then installer looks for *example_proposal* script.
+- `name` (string) - proposal client name without the `_proposal` suffix
+  (the code actually allows writing it but that makes the definitions longer
+  and less readable).
+- `presentation_order` (integer, optional) - the number defines the display order
+  in the proposal, the absolute value is not important, important is the relative
```
imobachgs: np: `important, important` looks strange.

doc/control-file.md
```diff
@@ -1158,20 +1095,16 @@ This is the full list of SLE-12 SP2:
...
 ```
-    <self_update_url>http://updates.opensuse.org/$arch/leap-42.1.dud</self_update_url>
+    <self_update_url>http://updates.opensuse.org/$arch/leap-42.1-installer-update</self_update_url>
 ```
 
-The URL can contain a variable `$arch` that will be replaced by the system's
-architecture, such as `x86_64`, `s390x`, etc. You can find more information in
-the [Arch
-module](http://www.rubydoc.info/github/yast/yast-yast2/Yast/ArchClass).
+This is the fallback which is used when the self-update repository is not\
```
imobachgs: Why is this backslash needed?

lslezak: Typo, already fixed

imobachgs:
LGTM

PR #457, opened 2016-11-10
[RFC] Proposal error popup

jreidinger:
I will add changes, when it is final and also create sr against CASP first.

glob content:

Some errors in proposal is non-blocking, but is serious and often lead to problems later in installation. In past it happen that people overlook such errors. So to even more high-light it, we decided to require one more confirmation when such situation happen. You can see how it will look like in following screenshots:
error_proposal_qt
error_proposal_ncurses

### DIFF
--- a/src/lib/installation/proposal_errors.rb
+++ b/src/lib/installation/proposal_errors.rb
@@ -1,0 +1,51 @@
+require "yast"
+
+Yast.import "UI"
+Yast.import "Label"
+Yast.import "Mode"
+Yast.import "Popup"
+
+module Installation
+  class ProposalErrors
+    include Yast::I18n
+    include Yast::Logger
+
+    ERROR_PROPOSAL_TIMEOUT = 60
+
+    def initialize
+      textdomain "installation"
+      @errors = []
+    end
+
+    # clears previously stored errros
+    def clear
+      @errors = []
+    end
+
+    # appends new error with given message
+    def append(message)
+      @errors << message
+    end
+
+    # returns true if there is no error or user approved stored errors
+    def approved?
+      return true if @errors.empty?
+
+      headline = _("Error Found in Installation Settings")
+      text = _("The following errors were found in the configuration proposal.\n" \
+        "If you continue with the installation it may not be successful.\n" \
+        "Errors:\n")
+      sep = Yast::UI.TextMode ? "-" : "•"
+      text += "#{sep} " + @errors.join("\n#{sep} ")
+
+      if Yast::Mode.auto
+        !Yast::Popup.TimedErrorAnyQuestion(headline, text,
+          Yast::Label.BackButton, Yast::Label.ContinueButton, :focus_yes,
+          ERROR_PROPOSAL_TIMEOUT)
+      else
+        !Yast::Popup.ErrorAnyQuestion(headline, text,
+          Yast::Label.BackButton, Yast::Label.ContinueButton, :focus_yes)
+      end
+    end
+  end
+end
--- a/src/lib/installation/proposal_runner.rb
+++ b/src/lib/installation/proposal_runner.rb
@@ -22,6 +22,7 @@
 require "yast"
 
 require "installation/proposal_store"
+require "installation/proposal_errors"
 
 module Installation
   # Create and display reasonable proposal for basic
@@ -66,6 +67,7 @@ module Installation
       # BNC #463567
       @submods_already_called = []
       @store_class = store
+      @errors = ProposalErrors.new
     end
 
     def run
@@ -278,6 +280,8 @@ module Installation
         return nil
       end
 
+      return nil unless @errors.approved?
+
       if Yast::Stage.stage == "initial"
         input = Yast::WFM.CallFunction("inst_doit", [])
       # bugzilla #219097, #221571, yast2-update on running system
@@ -441,6 +445,7 @@ module Installation
         end
       end
 
+      @errors.clear
       @store.make_proposals(
         force_reset:      force_reset,
         language_changed: language_changed,
@@ -469,6 +474,7 @@ module Installation
 
       if !warning.empty?
         level = prop["warning_level"] || :warning
+        log.info "proposal returns warning with level #{level} and msg #{warning}"
 
         case level
         when :notice
@@ -476,6 +482,7 @@ module Installation
         when :warning
           warning = Yast::HTML.Colorize(warning, "red")
         when :error
+          @errors.append(warning)
           warning = Yast::HTML.Colorize(warning, "red")
         when :blocker, :fatal
           @have_blocker = true
--- a/test/proposal_errors_test.rb
+++ b/test/proposal_errors_test.rb
@@ -1,0 +1,33 @@
+#! /usr/bin/env rspec
+
+require_relative "./test_helper"
+
+require "installation/proposal_errors"
+
+describe ::Installation::ProposalErrors do
+  describe "#approved?" do
+    it "returns true if there is no error stored" do
+      expect(Yast::Popup).to_not receive(:ErrorAnyQuestion)
+      expect(subject.approved?).to eq true
+    end
+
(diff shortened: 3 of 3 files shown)


src/lib/installation/proposal_runner.rb
```diff
@@ -463,19 +466,52 @@ def make_proposal(force_reset, language_changed)
...
+
+      headline = _("Error found when proposing configuration")
+      text = _("Following errors found during proposal. If you continue with " \
+        "installation, there is a high risk of later issues when performing " \
+        "installation.\nErrors:\n>")
+      text += "- " + proposal_errors.join("\n- ")
+
+      if Yast::Mode.auto
+        !Yast::Popup.TimedErrorAnyQuestion(headline, text,
+          Yast::Label.BackButton, Yast::Label.ContinueButton, :focus_yes,
```
lslezak: Um, I'd prefer `no` as the default pre-selected button. The default should be safe.

src/lib/installation/proposal_runner.rb
```diff
@@ -463,19 +466,52 @@ def make_proposal(force_reset, language_changed)
...
+
+    ERROR_PROPOSAL_TIMEOUT = 60
+
+    def error_proposal_aproved?
+      return true if proposal_errors.empty?
+
+      headline = _("Error found when proposing configuration")
+      text = _("Following errors found during proposal. If you continue with " \
+        "installation, there is a high risk of later issues when performing " \
+        "installation.\nErrors:\n>")
```
lslezak: The `>` character at the end looks like a typo...

src/lib/installation/proposal_runner.rb
jreidinger: default is in fact BackButton, as it is first one. I do it this way, due to timed one, as default is false there

src/lib/installation/proposal_runner.rb
jreidinger: I forgot to push fix I did

src/lib/installation/proposal_runner.rb
```diff
@@ -463,19 +466,52 @@ def make_proposal(force_reset, language_changed)
...
+    ERROR_PROPOSAL_TIMEOUT = 60
+
+    def error_proposal_aproved?
+      return true if proposal_errors.empty?
+
+      headline = _("Error found when proposing configuration")
+      text = _("Following errors found during proposal. If you continue with " \
+        "installation, there is a high risk of later issues when performing " \
+        "installation.\nErrors:\n")
+      text += "-> " + proposal_errors.join("\n-> ")
```
lslezak: In Qt we should use some Unicode character, like we do e.g. for the role descriptions (https://github.com/yast/skelcd-control-SLES/blob/master/control/control.SLES.xml#L393). Using plain ASCII in GUI does not look nice...

Use ASCII only in text mode if Unicode does not work properly there (test it in installation).

src/lib/installation/proposal_runner.rb
```diff
@@ -463,19 +466,53 @@ def make_proposal(force_reset, language_changed)
...
+      headline = _("Error found when proposing configuration")
+      text = _("Following errors found during proposal. If you continue with " \
+        "installation, there is a high risk of later issues when performing " \
+        "installation.\nErrors:\n")
+      sep = Yast::UI.TextMode ? "-" : "•"
+      text += "#{sep} " + proposal_errors.join("\n#{sep} ")
+
+      if Yast::Mode.auto
+        !Yast::Popup.TimedErrorAnyQuestion(headline, text,
+          Yast::Label.BackButton, Yast::Label.ContinueButton, :focus_yes,
```
lslezak: A popup with `Back` and `Continue` buttons is quite unusual in YaST. We already have `Popup.ContinueCancel` call, so I think the `Continue` and  `Cancel` buttons would be more standard here.

jreidinger: well, following dialog to confirm dialog is back/install, so it looks consistent to me.

lslezak: I see.... OK, maybe we should ask Ken for feedback.

lslezak:
The error message is truncated in text mode. Please use some `*Long*` or `*Any*` method with `RichText` widget.

jreidinger:
I do not find appropriate one that have also autoyast Timed version and that is question with changable buttons. Any idea which one to use?

jreidinger:
@lslezak btw why it ignore newlines?

jreidinger:
updated, newline added to ncurses, also title updated according to Ken comment. New ncurses screenshot added.

src/lib/installation/proposal_runner.rb
```diff
@@ -463,19 +466,53 @@ def make_proposal(force_reset, language_changed)
...
+      @proposal_errors << msg
+    end
+
+    ERROR_PROPOSAL_TIMEOUT = 60
+
+    def error_proposal_aproved?
+      return true if proposal_errors.empty?
+
+      headline = _("Error Found in Installation Settings")
+      text = _("Following errors found in configuration proposal. If you continue with " \
```
lslezak: I'm not sure about the wording: `If you continue with installation ... when performing installation.` It sounds a bit repeated to me.

@mvidner any suggestion?

src/lib/installation/proposal_runner.rb
```diff
@@ -463,19 +466,53 @@ def make_proposal(force_reset, language_changed)
...
+      @proposal_errors = []
+    end
+
+    def add_proposal_error(msg)
+      @proposal_errors << msg
+    end
+
+    ERROR_PROPOSAL_TIMEOUT = 60
+
+    def error_proposal_aproved?
```
mvidner: "Error proposal" means that we propose an error. I cannot make a proper name for this method unless we move the 1st line out, so line 281 becomes
```rb
if !proposal_errors.empty? && !approve_of_proposal_errors?
  return nil
end
```
Note "a**pp**rove".

src/lib/installation/proposal_runner.rb
mvidner: > **The** following errors **were** found in **the** configuration proposal. If you continue with **the** installation it may not be successful.
Errors:

src/lib/installation/proposal_runner.rb
mvidner: IMHO the Back/Install dialog is nonstandard on purpose as it marks the point of no return. OTOH this one should use a familiar Continue-Cancel pattern.

src/lib/installation/proposal_runner.rb
```diff
@@ -463,19 +466,53 @@ def make_proposal(force_reset, language_changed)
       nil
     end
 
+    attr_reader :proposal_errors
+
+    def clean_proposal_errors
```
mvidner: clean (očistit) → clear (vyklidit), which is also what the Ruby stdlib prefers

jreidinger:
@mvidner as I already commented about back/Continue. It is a bit similar to Next/Back following it when you confirm you really want do installation. So I find it a bit inconsistent and also cancel looks like you cancel whole installation which is wrong. You click on "Install" button ( or upgrade ), it will show this popup, so you go back proposal screen.

mvidner:
OK, I don't care too much about the popup buttons, they are not visible in the default workflow anyway.

src/lib/installation/proposal_errors.rb
```diff
@@ -0,0 +1,50 @@
...
+    def clear
+      @errors = []
+    end
+
+    # appends new error with given message
+    def append(message)
+      @errors << message
+    end
+
+    ERROR_PROPOSAL_TIMEOUT = 60
```
lslezak: NP: Just move the constant definition to the top, it's a bit hidden between the method definitions...

jreidinger:
@mvidner changes at first go to CSAP. I create it against master, so it is easier for manuall testing, blog entry and making screenshots for it.

PR #459, opened 2016-11-11
network.service removed in YaST2-Firstboot.service

schubi2:


teclator:
So, after all the tests done by @schubi2 it seems that it is not needed anymore. Taking this in account:

LGTM

PR #463, opened 2016-11-17
Documentation for new storage related parameters in control.xml (Fate#321736)

shundhammer:


package/yast2-installation.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.1.216.1.3
+Version:        3.1.216.1.4
```
ancorgs: In my opinion this should be 3.1.217.1

Now that we merged the "merge_after_SP2_release" branch there is no need anymore for the fifth number. Everything from 3.1.217 is also in this branch, so let's sync versions please.

shundhammer: ok

ancorgs:
LGTM

PR #464, opened 2016-11-18
add missing file

jreidinger:


lslezak:
Um, I guess we will need a version update + changes to make the autosubmission work, the package is already on the CASP medium...

PR #467, opened 2016-11-22
merge SP2 to master

jreidinger:


doc/SELF_UPDATE.md
```diff
@@ -5,10 +5,11 @@ system installation. This feature will help to solve problems with the
 installation even after the media has been released. Check
 [FATE#319716](https://fate.suse.com/319716) for a more in-depth rationale.
 
-## Disabling Updates
```
imobachgs: No decision has been made about this, so in TW it's still set as enabled by default (although no repository is set). The initial value is set by Linuxrc and, in master, is set to `1`.

Maybe we could set it to `0` by default also in TW to keep it in sync with SLE 12 SP2 and Leap 42.2.

kobliha: I'd love to see self-update enabled by default in SP3 / TW in the future. So let's keep it as it is.

PR #469, opened 2016-12-05
add new keyboard and root pwd installation client

jreidinger:
https://trello.com/c/zG6NNbwv

RFC for now, tested manually, screenshot in trello. spec file update and also changes will come later


mvidner:
spec file should be adjusted:
- version
- dependency on users and country

src/lib/installation/clients/inst_keyboard_rootpwd.rb
```diff
@@ -0,0 +1,79 @@
...
+
+require "users/widgets"
+require "y2country/widgets"
+
+module Installation
+  # This library provides a simple dialog for setting new password for the
+  # system adminitrator (root) and keyboard layout.
+  # The new password is not stored here, just set in UsersSimple module
+  # and stored later during inst_finish.
+  class InstKeyboardRootpwd
```
jreidinger: abbreviate fixed. Regarding inheritance from `Yast::Client` it is used mainly for ycp clients which adds all possibly used helpers for easier transpilation. Now I think it is not needed to inherit it.
Well, regarding name with "Client" in it or in its namespace, I try to be consistent with rest of files in this directory which also do not use it. If there is some final agreement, I think it make sense to switch all of it, to have consistent naming.

src/Makefile.am
```diff
@@ -38,6 +38,7 @@ client_DATA = \
   clients/inst_initialization.rb \
   clients/inst_installation_options.rb \
   clients/inst_instsys_cleanup.rb \
+  clients/inst_keyboard_rootpassword.rb \
```
mvidner: An underscore is missing between `root` and `password` ... :duck:

jreidinger:
note: travis failed as new country is not there yet

package/yast2-installation.changes
```diff
@@ -1,5 +1,12 @@
 -------------------------------------------------------------------
-Mon Dec 13 10:54:21 UTC 2016 - kanderssen@suse.com
```
teclator: 🙈

PR #470, opened 2016-12-07
add roles additional dialogs support

jreidinger:
initial RFC, before merge abort have to be solved and also specification for additional_dialogs have to be written as there is more requirements for given dialogs.

### DIFF
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -849,11 +849,12 @@ Section *supported\_desktops* contains list of one or more
 System Roles, if defined in the control file, are presented during
 the first stage of the installation. The user will select one of them
 and they will affect the proposed configuration of Partitioning and Software.
-Selected role can also define additional dialogs that is shown when given role
-is selected. It is common installation dialog with abort, cancel and next
-layout. It supports and use all parameters from GetInstArgs module. When going
-back, it will first shown last additional dialog and when going back through
-all additional dialogs, it will show again roles selection.
+
+A role can also define additional dialogs that are shown when a given role is
+selected. It is a common installation dialog with *Abort*, *Cancel* and *Next*
+buttons. It supports and uses all parameters from the **GetInstArgs** module.
+When going back, it will first shown the last additional dialog and when going
+back through all additional dialogs, it will show again the roles selection.
 
 They were requested in FATE#317481 and they are an evolution of the earlier
 concept of Server Scenarios used in SLE 11.
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -399,6 +399,9 @@ These options usually enable or disable some installation feature.
 -   (boolean) *require\_registration* - Require registration of add-on
     product (ignored in the base product).
 
+-   (boolean) *readonly\_timezone* - Suggested timezone cannot be
+    changed by the user.
+
 ### Installation Helpers
 
 In the *globals* section, there are also helper variables for the
@@ -608,6 +611,9 @@ This is a list of supported entries in *software*:
 
 -   *default\_desktop* - defines a desktop selected by default by the
     installation.
+
+-   *clone\_install\_recommended\_default* - Default entry for
+    "install_recommended" in created AutoYaST configuration file.
 
 Additionally, you can configure how updating of packages should be
 performed. The following options are available:
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -859,7 +859,7 @@ and they will affect the proposed configuration of Partitioning and Software.
 A role can also define additional dialogs that are shown when a given role is
 selected. It is a common installation dialog with *Abort*, *Cancel* and *Next*
 buttons. It supports and uses all parameters from the **GetInstArgs** module.
-When going back, it will first shown the last additional dialog and when going
+When going back, it will first show the last additional dialog and when going
 back through all additional dialogs, it will show again the roles selection.
 
 They were requested in FATE#317481 and they are an evolution of the earlier
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -849,6 +849,11 @@ Section *supported\_desktops* contains list of one or more
 System Roles, if defined in the control file, are presented during
 the first stage of the installation. The user will select one of them
 and they will affect the proposed configuration of Partitioning and Software.
+Selected role can also define additional dialogs that is shown when given role
+is selected. It is common installation dialog with abort, cancel and next
+layout. It supports and use all parameters from GetInstArgs module. When going
+back, it will first shown last additional dialog and when going back through
+all additional dialogs, it will show again roles selection.
 
 They were requested in FATE#317481 and they are an evolution of the earlier
 concept of Server Scenarios used in SLE 11.
@@ -873,6 +878,7 @@ Example:
         <software>
           <default_patterns>base Minimal kvm_server</default_patterns>
         </software>
+        <additional_dialogs>kvm_setup,virt_manager_setup </additional_dialogs>
       </system_role>
     </system_roles>
 
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -1018,6 +1018,18 @@ layout
 
 ### Partitioning
 
+*proposal_settings_editable* (boolean, default _true_) is specifies if the user
+should be able to change the configuration of the storage proposal in the
+installer: What filesystem to use for the root partition, if there should be a
+separate /home partition, if LVM or encryption should be used. For some
+products this is not desired.
+
+*expert_partitioner_warning* (boolean, default _false_) specifies if there
+should be an extra warning pop-up dialog when the user enters the expert
+partitioner dialog during installation, for example because the product has
+special requirements for partitioning (Btrfs to support snapshots etc.).
+
+
 #### Subvolumes
 
 This section is used to specify what subvolumes should be created if Btrfs is
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -108,6 +108,7 @@ module Installation
 
   private
 
+    # gets array of clients to run for given role
     def additional_clients_for(role_id)
       clients = raw_roles.find { |r| r["id"] == role_id }["additional_dialogs"]
       clients ||= ""
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -27,7 +27,7 @@ Yast.import "ProductFeatures"
 
 module Installation
   # marker exception when additional dialog return abort
-  class AbortException < RuntimeError;end
+  class AbortException < RuntimeError; end
 
   class SelectSystemRole < ::UI::InstallationDialog
     class << self
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -19,6 +19,8 @@
 
 require "yast"
 require "ui/installation_dialog"
+
+Yast.import "GetInstArgs"
 Yast.import "Popup"
 Yast.import "ProductControl"
 Yast.import "ProductFeatures"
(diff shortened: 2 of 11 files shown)


doc/control-file.md
```diff
@@ -849,6 +849,11 @@ Section *supported\_desktops* contains list of one or more
 System Roles, if defined in the control file, are presented during
 the first stage of the installation. The user will select one of them
 and they will affect the proposed configuration of Partitioning and Software.
+Selected role can also define additional dialogs that is shown when given role
```
mvidner: Please start a new paragraph here.

*A* role can also define additional dialogs that **are** shown when *a* given role is selected. It is *a* common installation dialog with *A*bort, *C*ancel and *N*ext *buttons*. It supports and use**s** all parameters from *the* GetInstArgs module. When going back, it will first *show~~n~~* *the* last additional dialog and when going back through all additional dialogs, it will show again *the* roles selection.

doc/control-file.md
```diff
@@ -873,6 +878,7 @@ Example:
         <software>
           <default_patterns>base Minimal kvm_server</default_patterns>
         </software>
+        <additional_dialogs>kvm_setup,virt_manager_setup </additional_dialogs>
```
mvidner: Spaces and commas as separators are a mess in the control file format (just search for "separat"). Can we allow either and use spaces in the example?

src/lib/installation/select_system_role.rb
```diff
@@ -19,17 +19,27 @@
...
 require "yast"
 require "ui/installation_dialog"
+
+Yast.import "GetInstArgs"
 Yast.import "Popup"
 Yast.import "ProductControl"
 Yast.import "ProductFeatures"
 
 module Installation
+  # marker exception when additional dialog return abort
```
mvidner: Do we really want to use an exception to handle user input? Why not a boolean or a symbol flag? Or, if we need to do along jump, `throw` and `catch`.
http://stackoverflow.com/questions/51021/what-is-the-difference-between-raising-exceptions-vs-throwing-exceptions-in-ruby

src/lib/installation/select_system_role.rb
```diff
@@ -42,7 +52,13 @@ def run
         return :auto # skip forward or backward
       end
 
+      if Yast::GetInstArgs.going_back
+        return :next if run_clients(self.class.original_role_id, going_back: true)
```
mvidner: If we're going back, it returns `:next`? That does not sound right.

src/lib/installation/select_system_role.rb
```diff
@@ -84,11 +100,61 @@ def next_handler
...
     end
 
   private
 
+    # @return true if clients successfully go to next dialog after roles
+    def run_clients(role_id, going_back: false)
+      clients = raw_roles.find { |r| r["id"] == role_id }["additional_dialogs"]
+      clients ||= ""
+      clients = clients.split(",").map!(&:strip)
+
```
mvidner: A finer style point, I think from here on it should be a separate method that takes a client list and does not care about roles.

doc/control-file.md
jreidinger: what is benefit for it? is it required?

src/lib/installation/select_system_role.rb
jreidinger: well, abort is for me exceptional state, not common work-flow

src/lib/installation/select_system_role.rb
jreidinger: no, run_clients return true if user want to go forward...so if you go back and in that dialog press next, it will return next, otherwise it continue with drawing this dialog. But when thinking about it I have to check how it behave when there is no clients.

jreidinger: ah, it behave correctly, so it behave as it should

src/lib/installation/select_system_role.rb
jreidinger: make sense.

doc/control-file.md
mvidner: Not required, it is a potential improvement in the "leave things better than you found them" category.

src/lib/installation/select_system_role.rb
mvidner: It is a less common, but still perfectly expected state. Please avoid exceptions for this. https://github.com/bbatsov/ruby-style-guide#no-exceptional-flows

src/lib/installation/select_system_role.rb
```diff
@@ -84,11 +100,61 @@ def next_handler
...
+    rescue AbortException
+      finish_dialog(:abort)
+      return
     end
 
   private
 
+    # @return true if clients successfully go to next dialog after roles
+    def run_clients(role_id, going_back: false)
```
mvidner: JFYI, the generic installation variant of this is [ProductControl.RunFrom](https://github.com/yast/yast-yast2/blob/0dd92b3f76f367256f754f035d52eabe49593936/library/control/src/modules/ProductControl.rb#L1222)

src/lib/installation/select_system_role.rb
```diff
@@ -84,11 +100,61 @@ def next_handler
 
       apply_role(role_id)
 
+      # if run clients goes back, then show again this dialog
```
mvidner: How does `create_dialog; return` show this dialog again?

src/lib/installation/select_system_role.rb
```diff
@@ -84,11 +100,61 @@ def next_handler
...
       super
+
+    rescue AbortException
+      finish_dialog(:abort)
+      return
     end
 
   private
 
+    # @return true if clients successfully go to next dialog after roles
```
mvidner: The return value depends on
- are there any clients?
- are we going back?

all 3 of which are booleans and I am lost.

I think the API would be much clearer if this method returned `:next`, `:back` or `:abort` (which solves my exception objection too).

src/lib/installation/select_system_role.rb
jreidinger: Well, I am thinking about it whole weekend and in fact I think it will be even better if all abort after abort_handler confirmation raise this exception, because it is not common flow from my POV. It is single action -> abort and die ( even in this case, because abort handler is called after clicking in client ). Common flow for wizard is from my POV go forward and go backward. Kill wizard is not from my POV common flow.

src/lib/installation/select_system_role.rb
jreidinger: thanks for mentioning it should be placed here in comment.

src/lib/installation/select_system_role.rb
jreidinger: Consider flow here:

- client is called. It creates its own window wizard window. You press next and client return next. But installation clients do not open nor close wizard windows nor clean it, so result is that in this case that still first dialog is shown here. But when we call create_dialog, it will draw current dialog ( and for `InstallationDialog` class it do not create wizard as proper installation dialog ). And return is needed, to avoid calling super, which return `:next` after clicking on next button as we cancel this action and we are back in input loop.
What do you think helps with understanding this? I know it is non-trivial state and work-flow.

jreidinger:
@mvidner OK, so any suggestion what helps to make implementation more clean? What about moving such functionality in own class with more understandable methods like `redraw_dialog?`, `aborted?` or `go_forward?` ?

doc/control-file.md
jreidinger: well, that is probably caused by different view on "leave things better" as I am not convinced ( as opposed to some team members ) that less strictly format is better as it can confused from my POV reader why somewhere is used spaces and somewhere comma. I am completely fine with changing comma to space, but I think we should allow only one separator.

src/lib/installation/select_system_role.rb
jreidinger: in fact no, previously I use return nil here and result is that it do not draw dialog and show first additional one and not this one as Wizard have different content and nothing overwrite it. What can be solution is to call 
```
  Yast::Wizard.SetTitleIcon(title_icon) if title_icon
  Yast::Wizard.SetContents(
    dialog_title,
    dialog_content,
    help_text,
    Yast::GetInstArgs.enable_back,
    Yast::GetInstArgs.enable_next || Yast::Mode.normal
  )
```

which sets again wizard content.

mvidner: Er, no. Instead, a recursive `finish_dialog(run)` should do it.

src/lib/installation/select_system_role.rb
mvidner: I think you are generalizing too much. Here, exceptions are not necessary because you always rescue them directly in the calling method. I have committed an exceptionless way here: https://github.com/mvidner/yast-installation/commit/5d09e5cf04e89b4f8bd9fb038dfdc89c6784a7e0 in https://github.com/mvidner/yast-installation/commits/roles_additional_dialogs_casp

doc/control-file.md
mvidner: I see. It is a small matter not worth arguing now.

src/lib/installation/select_system_role.rb
```diff
@@ -84,11 +99,58 @@ def next_handler
...
+      result = run_clients(additional_clients(role_id))
+      # We show the main role dialog; but the additional clients have
+      # drawn over it, so do it again, and propagate its result.
+      result = self.class.run if result == :back
+      finish_dialog(result)
     end
 
   private
 
+    def additional_clients(role_id)
```
teclator: NP: I prefer additional_clients_for(role_id).

BTW some documentation will be nice

test/select_system_role_test.rb
```diff
@@ -65,6 +73,33 @@
         expect(subject.run).to eq(:back)
       end
 
+      context "when role contain additional dialogs" do
+        it "show first dialog when going forward" do
```
teclator: `it shows the first dialog`

doc/control-file.md
```diff
@@ -856,6 +856,12 @@ System Roles, if defined in the control file, are presented during
 the first stage of the installation. The user will select one of them
 and they will affect the proposed configuration of Partitioning and Software.
 
+A role can also define additional dialogs that are shown when a given role is
+selected. It is a common installation dialog with *Abort*, *Cancel* and *Next*
+buttons. It supports and uses all parameters from the **GetInstArgs** module.
+When going back, it will first shown the last additional dialog and when going
```
teclator: There is still "shown" instead of "show"

jreidinger:
when approved I plan to do final testing in real VM

src/lib/installation/select_system_role.rb
```diff
@@ -84,11 +99,63 @@ def next_handler
 
       apply_role(role_id)
 
-      super
+      result = run_clients(additional_clients_for(role_id))
+      # We show the main role dialog; but the additional clients have
+      # drawn over it, so do it again, and propagate its result.
+      result = self.class.run if result == :back
```
jreidinger: Here is a bug revealed from real testing.
Problem is in this scenario:

1. going to dialog after all additional dialogs
2. now going back thrue all additional dialogs to original roles ( works great )
3. go again forward with next
4. and on first additional dialog go back
5. now bug appear as it shown the last additional dialog.

Explanation - going_back is set to thrue, so when this run is called, then https://github.com/yast/yast-installation/pull/470/files#diff-77dcc9aa9ffe2786edbd6aa5beda2f4dR52 this line in run is in affect and show the last additional dialog instead of drawing roles dialog.

mvidner: I see :-( I think we should look at the main workflow with ProductControl.RunFrom and add-on products. Maybe there is an analogous bug there, or maybe the implementations can be unified easily enough, or maybe it is best to keep this bug as is.

jreidinger: well main workflow do not have this problem as it have going back, if we have our internal going_back as global variable we can detect it, but I do not like it. I still think that the easiest way how to solve it is to simply call `create_dialog` and return without finish as it was before as it behave correctly, even not so nicely from design perspective.

jreidinger: @teclator it does not help, it works as expected. clients size condition is passed correctly ( it is set to `:back` which is correct. Problem here is that run, which start again when going back additional dialogs.

jreidinger:
@mvidner thanks

PR #472, opened 2016-12-07
bsc#1013976 in CASP

imobachgs:
Merge #471 into the `SLE-12-SP2-CASP` branch.


jreidinger:
LGTM, just please verify with @kobliha that we really want by default disabled self update for CASP.

imobachgs:
@jreidinger Ok. I'll merge this change (as it fixes a bug) and I'll try to get more information about the self-update thingie. As it's base in `SP2`, I assume that it's disabled by default (and we're just updating the documentation in this PR).

PR #474, opened 2016-12-07
New worker role dialog.

teclator:
Trello card: https://trello.com/c/OFmpN4xq/785-5-casp-role-specific-dialog

By now it just write the log file when going next.

registerworker

### DIFF
--- a/src/lib/installation/dialogs/worker_role.rb
+++ b/src/lib/installation/dialogs/worker_role.rb
@@ -1,0 +1,61 @@
+# Copyright (c) 2016 SUSE LLC.
+#  All Rights Reserved.
+
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of version 2 or 3 of the GNU General
+#  Public License as published by the Free Software Foundation.
+
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.   See the
+#  GNU General Public License for more details.
+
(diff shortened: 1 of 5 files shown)


src/lib/installation/dialogs/worker_role.rb
```diff
@@ -0,0 +1,58 @@
...
+module Yast
+  class WorkerRoleDialog < ::UI::InstallationDialog
+    def initialize
+      super
+
+      textdomain "installation"
+    end
+
+    def next_handler
+      log.info("Worker role dialog should register with the admin dashboard")
```
jreidinger: well, when you want to log TODO. I prefer to use "error" level for it and maybe also add it as comment, so it can be easily grepped.

src/lib/installation/dialogs/worker_role.rb
```diff
@@ -0,0 +1,58 @@
...
+          VSpacing(1),
+          VBox(
+            InputField(Id(:dashboard_url), Opt(:hstretch), _("Dashboard U&RL"))
+          )
+        )
+      )
+    end
+
+    def dialog_title
+      _("Worker Role")
```
jreidinger: Does it make sense this title? what about "Register Worker" ?

src/lib/installation/dialogs/worker_role.rb
teclator: Updated with your suggestions.

src/lib/installation/dialogs/worker_role.rb
teclator: Updated

teclator:
@jreidinger 

registerworker

jreidinger:
LGTM

PR #476, opened 2016-12-12
Symbols; recursion.

mvidner:


src/lib/installation/select_system_role.rb
```diff
@@ -100,29 +98,31 @@ def next_handler
...
-
-    rescue AbortException
-      finish_dialog(:abort)
-      return
     end
 
   private
 
-    # @return true if clients successfully go to next dialog after roles
+    # @return [Symbol] what should the caller do next
```
jreidinger: well, I just do not like wording. method should not tell caller what to do. We should say something like return symbol for direction what clients return + abort when abortion is in process. What do you think?

jreidinger:
thanks

PR #479, opened 2016-12-14
Set root subvolume read-only for CASP if configured (Fate#321755)

shundhammer:
https://trello.com/c/sLkMAwRU/479-3-casp-321755-microos-switch-root-subvolume-to-read-only-s

This PR is the code and XML documentation part.
PR for the XML schema and the CASP control.xml (skelcd-CASP) will follow.

Please notice the PBI called for a simple solution. This approach will fail silently if the configuration parameter is set in control.xml, but the root filesystem type is something other than Btrfs.


wfeldt:
lgtm

PR #482, opened 2016-12-16
relax dependency due to splitted country for SLE12 and TW

jreidinger:


package/yast2-installation.spec
```diff
@@ -83,8 +83,8 @@ Requires: yast2-network >= 3.1.143
 
 # new root password cwm widget
 Requires:	yast2-users >= 3.2.7
-# new keyboard layout cwm widget
-Requires:	yast2-country >= 3.2.5
+# new keyboard layout cwm widget, but keep dependency compatible with SLE12
+Requires:	yast2-country >= 3.1.30
```
ancorgs: AFAICS, the widget was introduced in 3.1.33.1, not in 3.1.30.

On the other hand, travis fails.

jreidinger: I know, but I if we in SP3 use SP2 version and not CASP version, we have problem. Travis failed due to not available package built yet. problem is that it is used only in casp, but testsuite require all files, even that which is not used for given product.

ancorgs: But if we want the master version of yast2-installation to work in in SP3... shouldn't we ensure SP3 version of yast2-country includes the widget? Otherwise it will be broken anyways.

PR #483, opened 2016-12-19
Enable RO proposal in case of error

mchf:
https://trello.com/c/sEVjWKcJ/780-3-casp-read-only-software-selection

Proposals in control file currently have format:
```
<proposal_module>
  <name>software</name>
  <read_only config:type="boolean">true</read_only>
</proposal_module>
```
this PR changes this to
```
<proposal_module>
  <name>software</name>
  <!-- soft means that proposal becomes active when an error happens,
         hard means that proposal is always RO -->
  <read_only>soft</read_only>
</proposal_module>
```

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Dec 19 13:23:12 UTC 2016 - mfilka@suse.com
+
+- Made user's interaction possible in case of error in read-only
+  proposal.
+- 3.1.217.9 
+
+-------------------------------------------------------------------
 Wed Dec 14 15:47:02 UTC 2016 - jreidinger@suse.com
 
 - Change layout of new installation dialog according to UX team
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,8 +1,9 @@
 -------------------------------------------------------------------
 Mon Dec 19 13:23:12 UTC 2016 - mfilka@suse.com
 
-- Made user's interaction possible in case of error in read-only
-  proposal.
+- fate#321739
+  - Made user's interaction possible in case of error in read-only
+    proposal.
 - 3.1.217.9 
 
 -------------------------------------------------------------------
--- a/src/lib/installation/proposal_runner.rb
+++ b/src/lib/installation/proposal_runner.rb
@@ -807,7 +807,7 @@ module Installation
       title = @store.title_for(submod)
 
       # do not add a link if the module is read-only or link is already included
-      heading = if @store.read_only?(submod) || title.include?("<a")
+      heading = if @store.read_only_no_recovery?(submod) || title.include?("<a")
         title
       else
         Yast::HTML.Link(title, @store.id_for(submod))
--- a/src/lib/installation/proposal_runner.rb
+++ b/src/lib/installation/proposal_runner.rb
@@ -807,7 +807,7 @@ module Installation
       title = @store.title_for(submod)
 
       # do not add a link if the module is read-only or link is already included
-      heading = if @store.read_only_no_recovery?(submod) || title.include?("<a")
+      heading = if @store.hard_read_only?(submod) || title.include?("<a")
         title
       else
         Yast::HTML.Link(title, @store.id_for(submod))
--- a/src/lib/installation/proposal_runner.rb
+++ b/src/lib/installation/proposal_runner.rb
@@ -365,6 +365,11 @@ module Installation
       workflow_sequence
     end
 
+    # Checks if given proposal map contains an error report
+    #
+    # @param [Hash] proposal map as returned by make_proposal
+    # @return [Boolean] true if the map reports an issue in proposal
+    # @see ProposalClient#make_proposal
     def proposal_failed?(proposal)
       proposal && [:blocker, :fatal, :error].include?(proposal["warning_level"])
     end
--- a/src/lib/installation/proposal_runner.rb
+++ b/src/lib/installation/proposal_runner.rb
@@ -365,6 +365,10 @@ module Installation
       workflow_sequence
     end
 
+    def proposal_failed?(proposal)
+      proposal && [:blocker, :fatal, :error].include?(proposal["warning_level"])
+    end
+
     def make_proposal(force_reset, language_changed)
       tab_to_switch = 999
       current_tab_affected = false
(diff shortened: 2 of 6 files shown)


src/lib/installation/proposal_store.rb
```diff
@@ -300,13 +320,19 @@ def client_for_link(link)
...
 
         name = full_module_name(proposal["name"])
-        @read_only_proposals << name
+
+        case proposal["read_only"]
+        when "hard"
+          @read_only_proposals[:hard] << name
+        when "soft"
+          @read_only_proposals[:soft] << name
+        end
```
lslezak: The `else` branch is missing, it should at least log the unknown value. It would be better to put it in one of the list otherwise it will be completely lost. Probably the `:soft` is a good default as it allows user to fix a possible issue without locking him out.

src/lib/installation/proposal_store.rb
```diff
@@ -250,12 +250,32 @@ def title_for(client)
...
+    # "hard" read-only means that the proposal is always read-only
+    # "soft" read-only means that the proposal is made changeable when an error
+    # in proposal is detected.
     #
     # @param [String] client
-    # @return [String] a title provided by the description API
-    def read_only?(client)
-      read_only_proposals.include?(client)
+    # @return [Boolean] if the client is marked as "hard" read only
+    def read_only_no_recovery?(client)
```
lslezak: I think `hard_read_only?` and `soft_read_only?` names would sound better and better match the description.

package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Dec 19 13:23:12 UTC 2016 - mfilka@suse.com
+
+- Made user's interaction possible in case of error in read-only
+  proposal.
```
lslezak: Add a FATE reference.

mvidner:
This only makes sense when other pieces of the puzzle are present, but there are no references to them
- bnc/fate
- trello
- other PRs

src/lib/installation/proposal_store.rb
```diff
@@ -250,12 +250,40 @@ def title_for(client)
...
+
+    # Checks if the client's proposal is configured as "hard" read-only
+    #
+    # "hard" read-only means that the proposal is always read-only
+    # "soft" read-only means that the proposal is made changeable when an error
+    # in proposal is detected.
+    #
+    # @param [String] client
+    # @return [Boolean] if the client is marked as "hard" read only
+    def read_only_no_recovery?(client)
```
lslezak: I think `hard_read_only?` and `soft_read_only?` names would sound better and better match the description.

mvidner:
In particular, I am missing the logic that says: "an error has in fact occurred so let's make a read-only part modifiable now".

src/lib/installation/proposal_store.rb
```diff
@@ -300,13 +328,19 @@ def client_for_link(link)
...
 
       properties.fetch("proposal_modules", []).each do |proposal|
         next unless proposal["read_only"]
 
         name = full_module_name(proposal["name"])
-        @read_only_proposals << name
+
+        case proposal["read_only"]
+        when "hard"
+          @read_only_proposals[:hard] << name
```
lslezak: There is an inconsistency with the PR description, it lists `<read_only>soft</read_only>` config option (which is a string), here in the code is `:hard` symbol.

Either change the code to `"hard"` or change the configuration to `<read_only config:type="symbol">soft</read_only>`.

mvidner: No, it's OK. The condition uses a string.

lslezak: Oh, I overlooked that, OK...

src/lib/installation/proposal_runner.rb
```diff
@@ -807,7 +807,7 @@ def html_header(submod)
       title = @store.title_for(submod)
 
       # do not add a link if the module is read-only or link is already included
-      heading = if @store.read_only?(submod) || title.include?("<a")
+      heading = if @store.hard_read_only?(submod) || title.include?("<a")
```
lslezak: It seems it does not handle the soft read only case correctly. For soft RO the proposal should be RO, RW only when there is an error. The error state for the soft case is not handled at all here.

mchf:
There is no need to adjust old testsuite. ```read_only?``` still exists and still use boolean.

lslezak:
I see, then please add new test for the added functionality.

And Travis fails because of `cannot load such file -- y2country/widgets (LoadError)`, no idea why...

lslezak:
OK, this looks better, thanks!

test/proposal_runner_test.rb
```diff
@@ -75,6 +75,30 @@ class C < ::Installation::ProposalStore; end
...
     end
 
+    describe "#html_header" do
+      it "returns clickable header when forced" do
+        allow(Yast::GetInstArgs).to receive(:proposal).and_return("software")
+
+        expect_any_instance_of(::Installation::ProposalRunner)
+          .to receive(:submod_descriptions_and_build_menu)
+          .and_return(false)
+        expect_any_instance_of(::Installation::ProposalStore)
```
lslezak: The test is OK, but as you mock most (all?) of the `::Installation::ProposalStore` functionality it would be nice to add also some tests to https://github.com/mchf/yast-installation/blob/ae795f1094d361c5db76aace03776951e6923323/test/proposal_store_test.rb#L566, esp. for the soft RO case. TIA!

lslezak:
Almost there... :wink:

Please add also a test with soft RO and and as error state. (Check that the header is clickable and the client can be called.)

mchf:
Related PR: https://github.com/yast/yast-installation-control/pull/33

mchf:
Related PR: https://github.com/yast/skelcd-control-CASP/pull/16

test/proposal_runner_test.rb
```diff
@@ -150,6 +149,45 @@ class C < ::Installation::ProposalStore; end
...
+            { "name" => "software", "presentation_order" => "15", "read_only" => "soft" }
+          ]
+        }.freeze
+        let(:proposals) { [["software_proposal", 15]] }
+        let(:software_header) {
+          Yast::HTML.Heading(Yast::HTML.Link("Software", "software"))
+        }
+
+        it "makes a proposal" do
+          allow(subject)
```
lslezak: NP: I'd a comment here that this is needed to allow the other calls than the one `expect`ed later, otherwise it looks a bit confusing...

mchf:
ok, I'll wait for https://github.com/yast/yast-installation/pull/485, rebase and then we will see.

lslezak:
Please remove the last commit (`Use Docker at Travis`), I have fixed a packaging issue in the Travis packages. It works now.

mchf:
@lslezak thanks, last commit removed. However the build this time failed in jenkins (seems like an accessibility issue)

lslezak:
We can ignore the Jenkins failure, it tries to build against CASP in IBS which cannot work from the public Jenkins. Moreover after switching to Docker at Travis we do not this Jenkins job anymore

mchf:
Thanks for reviews

PR #485, opened 2017-01-06
Use Docker at Travis

lslezak:


lslezak:
The `yast-installation-image` is the image name built in the previous step (`before_install`), the `yast-travis-ruby` is a script included in the base `yastdevel/ruby` image (just check the comment above :wink: )

mchf:
ok thanks, for explanation ... and you know comments and doc are last instance where to look ;-)

PR #487, opened 2017-01-12
CASP new all-in-one overview dialog

mvidner:
- https://fate.suse.com/322328
- https://trello.com/c/XDgl2dj7/816-13-casp-new-all-in-one-dialog

### TODOs 

These TODOs are part of the above Feature / PBI, but are NOT part of this PR. We want another PR for these before the end of the sprint.

- Registration does nothing (which includes not updating the software proposal)
- Dashboard URL appears, but does nothing (and it is not a mandatory field as it should be)
- Errors in any proposal are silently ignored. In particular that includes partitioning, bootloader, software (which is a hidden one)
- The installation confirmation dialog is missing (it was part of inst_proposal which we replaced)

### Screenshots

casp-overview5-gui
casp-overview6-tui


src/lib/mocked_widgets.rb
```diff
@@ -0,0 +1,110 @@
+require "yast"
```
jreidinger: do we really want to have this living in top level? Even if it is temporary

jreidinger: and second note, this won't work, as CASP still use autotools so Makefile.am modification is needed. ( Yes, annoying and I already change it for master )

src/lib/mocked_widgets.rb
```diff
@@ -0,0 +1,110 @@
...
+
+module Widgets
+  class Overview < CWM::CustomWidget
+    def contents
+      VBox(
+        PushButton(Id(button_id), self.label),
+        Label("* foo"),
+        Label("* bar")
+      )
+    end
```
jreidinger: maybe generate it somehow from proposal client? e.g. for bootloader it make some sense as it already provide such info.

src/lib/mocked_widgets.rb
mvidner: so I should move it to
- `src/lib/installation/widgets/mocked.rb` and split off
- `src/lib/installation/widgets/system_role.rb` right away, OK?

src/lib/mocked_widgets.rb
mvidner: Right, I have also thought about that. Let's see how it works out.

src/lib/mocked_widgets.rb
jreidinger: for now I am fine with `src/lib/installation/widgets.rb` and if it get more complex we can have `src/lib/installation/widgets/system_role.rb` and original `src/lib/installation/widgets.rb` will just require it. Just idea, but anything that is not top level is good :)

shundhammer:
Yikes... I am aware that the CASP/CaaS people really want only one dialog, but are they really prepared to become the honor nominees of the UI hall of shame for it? Seriously, that's at least two dialogs rolled into one:

- The dialog where the user (the admin really) **has** to enter some values
- The installation summary lookalike where changing things is purely optional

IMHO it's not a good idea to muddle that up into one dialog. Our usability experts will get a fit, and rightfully so.

src/lib/installation/clients/inst_casp_overview.rb
```diff
@@ -128,12 +128,12 @@ def content
...
-          Installation::Widgets::Overview.new(client: "partitions_proposal"),
-          Installation::Widgets::Overview.new(client: "bootloader_proposal")
+          Installation::Widgets::PartitioningOverview.new,
+          Installation::Widgets::BootloaderOverview.new
         ),
         lower_right: VBox(
-          Installation::Widgets::Overview.new(client: "network_proposal"),
-          Installation::Widgets::Overview.new(client: "kdump_proposal")
+          Installation::Widgets::NetworkOverview.new,
+          Installation::Widgets::KdumpOverview.new
```
jreidinger: Reason is simple. widget_id is if not defined, taken from class name. So what you need to do is to add to def initialize line `self.widget_id = client`

src/lib/installation/widgets/overview.rb
```diff
@@ -92,5 +92,16 @@ def initialize
...
       end
     end
+
+    class InvisibleSoftwareOverview < Overview
+      def initialize
+        super(client: "software_proposal")
+      end
+
+      def contents
+        _ = items
```
jreidinger: why not simple `items` with comment that it will make proposal?

jreidinger: and I miss that validation we are talking about.

src/lib/installation/clients/inst_casp_overview.rb
mvidner: @jreidinger ah, thanks! fixed.

src/lib/installation/widgets/overview.rb
mvidner: I meant to express that we discard the return value, but I may as well say that in a comment
```rb
items     # propose software, but discard the returned summary
Empty()
```

Yes, the validation is missing. I will check the card requirements and either add TODOs or implement it before merging.

mvidner:
I have squashed the history a bit.

PR #488, opened 2017-01-12
adapt services according to settings in roles

jreidinger:


src/lib/installation/services.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module Installation
+  # Represents services manipulation in installation.
+  class Services
+    class << self
+      include Yast::Logger
+
+      def enabled
+        @enabled ||= []
+        @enabled
```
imobachgs: This line is not necessary (as the previous one returns the value).

src/lib/installation/services.rb
```diff
@@ -0,0 +1,34 @@
+require "yast"
+
+Yast.import "SystemdService"
+
+module Installation
+  # Represents services manipulation in installation.
+  class Services
```
imobachgs: Does the API in https://github.com/yast/yast-services-manager/blob/master/src/modules/services_manager.rb fit? What do you think about extending it?

I guess it's not an option and that's the reason to create a new class specific to the installation process. In that case, maybe we could document it.

src/lib/installation/services.rb
jreidinger: you are right, previously it was more complex stuff there

src/lib/installation/services.rb
```diff
@@ -0,0 +1,34 @@
...
+    class << self
+      include Yast::Logger
+
+      def enabled
+        @enabled ||= []
+        @enabled
+      end
+
+      def enabled=(services)
+        if !services.is_a?(::Array)
```
jreidinger: well, we started in ruby-bindings with it. It helps user of class to notice wrong usage. @mvidner what do you think?

src/lib/installation/services.rb
jreidinger: well, to be honst I am not sure if services-manager is in insts-sys for now? and also it looks like module for autoinstallation, at least it is my impression, so I do not want to interfere with it.

imobachgs: I see your point. On the other hand, the module is in the installation media (for SLE 12 SP2 and even for CASP) and it's the one behind the `System -> Services Manager` YaST2 option, so it also have a UI.

jreidinger: you are right, it is there. So I again check this file and in fact this module is more useful https://github.com/yast/yast-services-manager/blob/master/src/modules/services_manager_service.rb but it have one quite big problem. It is not adapted to work in insts-sys, it probably often fail with attempt to call systemd calls not available in chroot env. I am not sure if it is worth this work to adapt it to work properly in this env.

src/lib/installation/clients/inst_extrasources.rb
```diff
@@ -245,7 +245,7 @@ def GetURLsToRegister(registered)
     # Register the installation sources in offline mode (no network connection required).
     # The repository metadata will be downloaded by sw_single (or another yast module) when the repostory is enabled
     #
-    # @param list <map> list of the sources to register
+    # @param url_list [Array<Hash>] list of the sources to register
```
imobachgs: What about describing the content of those hashes? At least the keys the have.

src/lib/installation/select_system_role.rb
```diff
@@ -187,6 +189,17 @@ def apply_role(role_id)
       features = features.dup
       NON_OVERLAY_ATTRIBUTES.each { |a| features.delete(a) }
       Yast::ProductFeatures.SetOverlay(features)
+      adapt_services(role_id)
+    end
+
+    def adapt_services(role_id)
```
imobachgs: I would add some documentation to this method.

src/lib/installation/services.rb
imobachgs: it's still not needed, right?

src/lib/installation/services.rb
imobachgs: I agree it can be helpful, but it's not the usual practice in the Ruby world. On the other hand, it won't hurt and I'm not against it.

test/services_test.rb
```diff
@@ -0,0 +1,51 @@
+#! /usr/bin/env rspec
+
+require_relative "./test_helper"
+
+require "installation/services"
+
+describe ::Installation::Services do
+  before do
+    # simulate non used class
+    described_class.instance_variable_set(:@enabled, nil)
```
imobachgs: np: I would rather prefer having a `reset` method or something similar, although that method would be only used during tests.

src/lib/installation/clients/inst_extrasources.rb
jreidinger: it would be nice, I just fix errors reported by yardoc, but I can try to dig a bit and describe it

src/lib/installation/select_system_role.rb
jreidinger: yes, make sense

src/lib/installation/services.rb
jreidinger: damn, I did commit at different pc and do not push it

test/services_test.rb
jreidinger: well, I have my inner rule to never affect production code by tests. I believe that API should make sense in target domain and should be as minimal as possible. I still remember old Yast code which have on every second line `if Mode.Test` and it looks horrible.
So to sum it up, I think it is good idea to add such method only if there is real usage, but now it is always set, so no real usage.

imobachgs: Fine with me. Anyway, I think that the `Mode.Test` thingie is a different issue: the behaviour changes when the system is under testing.

PR #491, opened 2017-01-16
Do not retry release notes download after getting a 4xx

imobachgs:
Fixes [bsc#1015794](https://bugzilla.suse.com/show_bug.cgi?id=1015794).

### DIFF
--- a/src/lib/installation/clients/inst_download_release_notes.rb
+++ b/src/lib/installation/clients/inst_download_release_notes.rb
@@ -29,6 +29,12 @@ module Yast
   class InstDownloadReleaseNotesClient < Client
     include Yast::Logger
 
+    # When cURL returns one of those codes, the download won't be retried
+    # @see man curl
+    #  7 = Failed to connect to host.
+    # 28 = Operation timeout.
+    CURL_GIVE_UP_RETURN_CODES = [7, 28].freeze
+
     # Download all release notes mentioned in Product::relnotesurl_all
     #
     # @return true when successful
@@ -111,10 +117,7 @@ module Yast
             InstData.release_notes[product["short_name"]] = SCR.Read(path(".target.string"), filename)
             InstData.downloaded_release_notes << product["short_name"]
             break
-          # exit codes (see "man curl"):
-          #  7 = Failed to connect to host.
-          # 28 = Operation timeout.
-          elsif ret == 7 || ret == 28
+          elsif CURL_GIVE_UP_RETURN_CODES.include?(ret)
             log.info "Communication with server for release notes download failed, skipping further attempts."
             InstData.stop_relnotes_download = true
             break
--- a/src/lib/installation/clients/inst_download_release_notes.rb
+++ b/src/lib/installation/clients/inst_download_release_notes.rb
@@ -101,6 +101,12 @@ module Yast
           # Where we want to store the downloaded release notes
           filename = Builtins.sformat("%1/relnotes",
             SCR.Read(path(".target.tmpdir")))
+
+          if InstData.failed_release_notes.include?(url)
+            log.info("Skipping download of already failed release notes at #{url}")
+            next
+          end
+
           # download release notes now
           cmd = Builtins.sformat(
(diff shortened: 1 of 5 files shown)


src/lib/installation/clients/inst_download_release_notes.rb
```diff
@@ -113,8 +116,9 @@ def download_release_notes
             break
           # exit codes (see "man curl"):
           #  7 = Failed to connect to host.
+          # 22 = Not found or returned another error with HTTP code being 400 or above.
```
lslezak: Please move the exit code comment above (to the list definition where the values are used).

test/lib/inst_download_release_notes_test.rb
```diff
@@ -0,0 +1,151 @@
...
+    let(:proxy) { double("proxy", "Read" => nil, "enabled" => false) }
+    let(:curl_code) { CURL_SUCCESS_CODE }
+    let(:language) { "en_US" }
+
+    before do
+      stub_const("Yast::Proxy", proxy)
+      allow(Yast::Pkg).to receive(:ResolvableProperties).with("", :product, "")
+        .and_return([product])
+
+      allow(Yast::SCR).to receive(:Execute).and_call_original
```
lslezak: NP: What are the other cases which we need to call? I do not like calling the real SCR in tests, ideally all SCR calls should be mocked.

- Running real SCR is not safe.
  - `Write`might change something in the system
  - `Read` reads from the current system and makes the test fragile, it may suddenly fail in another environment (Travis, different machine, root vs non-root user...).
- The test might run too long, stubbing the system access makes it faster (I have seen a test where SLP discovery was not mocked and it was called several time in the testsuite...)

If there are just few cases I'd rather mock them instead of this *blank cheque* approach...

test/lib/inst_download_release_notes_test.rb
```diff
@@ -0,0 +1,151 @@
...
+      allow(Yast::SCR).to receive(:Execute)
+        .with(Yast::Path.new(".target.bash"), /curl.*relnotes/)
+        .and_return(curl_code)
+
+      allow(Yast::SCR).to receive(:Read).and_call_original
```
lslezak: NP: The same here? Can we mock the calls instead?

test/lib/inst_download_release_notes_test.rb
```diff
@@ -0,0 +1,152 @@
...
+Yast.import "InstData"
+Yast.import "Pkg"
+Yast.import "Language"
+
+describe Yast::InstDownloadReleaseNotesClient do
+  CURL_NOT_FOUND_CODE = 22
+  CURL_SUCCESS_CODE = 0
+  CURL_HARD_ERROR = 7
+
+  subject(:client) { described_class.new }
```
imobachgs: I see your point, but I like to describe what I'm testing (and in this case it's a client). There's an [interesting article](http://blog.davidchelimsky.net/blog/2012/05/13/spec-smell-explicit-use-of-subject/) about that written by David Chelimsky, one of the former RSpec developers (you can skip to the last part).

test/lib/inst_download_release_notes_test.rb
imobachgs: Yes, you're right. The problem is that our modules do a lot of things and, for example, a call introduced in some of our dependencies can make this test to fail.

But I've checked what you said and, in this case, you're right. So I've removed the *blank cheque* approach :)

lslezak: Just a note: The problem with adding new SCR calls in the dependent code can be avoided by stubbing the code at a higher level, eg. with

```
def init
  SCR.Read(".foo")
  SCR.Read(".bar")
end
```

do not mock `SCR.Read` but whole `init`. Introducing yet another SCR call there will not break it.

test/lib/inst_download_release_notes_test.rb
lslezak: I see, thanks for the link.

PR #497, opened 2017-01-23
Use network proposal

teclator:
Use the new network proposal added in yast/yast-network#491


mvidner:
Thanks!

In the next step I need to move this out of the `mocked.rb` file, but that should not block this PR

PR #499, opened 2017-01-24
Use Registration::Widgets::RegistrationCode and removed mocked widgets

teclator:
It depends on:

https://github.com/yast/yast-registration/pull/292


mvidner:
Superseded by #504

PR #501, opened 2017-01-25
Disable self update via AutoYaST profile

imobachgs:
Using the following snippet allows to disable the self-update using an AutoYaST profile:

```xml
<general>
  <self_update config:type="boolean">false</self_update>
</general>
```

I'm also removing the step `Fetching AutoYaST Profile`. Why? Because I think that we should show the same UI if you disable the self-update via Linuxrc (no progress is shown) or via a profile. It can be confusing to see the `Updating Installer` screen (although for 1 sec) if you explicitly has disabled it.

### DIFF
--- a/doc/SELF_UPDATE.md
+++ b/doc/SELF_UPDATE.md
@@ -9,14 +9,17 @@ installation even after the media has been released. Check
 
 Starting in SUSE Linux Enterprise 12 SP3, self-update is enabled by default.
 However, it can be disabled by setting `self_update=0` boot option. If you're
-using AutoYaST, is also possible to disable this feature using the `self_update`
-element in `general` section of the profile:
+using AutoYaST, it is also possible to disable this feature using the
+`self_update` element in the `general` section of the profile:
 
    ```xml
    <general>
      <self_update config:type="boolean">false</self_update>
    </general>
    ```
+
+Please, take into account that self-update will be disabled if any of those
+values is set.
 
 During the rest of this document it is assumed that self-update is enabled.
 
--- a/doc/SELF_UPDATE.md
+++ b/doc/SELF_UPDATE.md
@@ -7,9 +7,18 @@ installation even after the media has been released. Check
 
 ## Disabling Updates
 
-Self-update is enabled by default. However, it can be disabled by setting
-`self_update=0` boot option. In the following sections it is assumed that this
-feature is enabled.
+Starting in SUSE Linux Enterprise 12 SP3, self-update is enabled by default.
+However, it can be disabled by setting `self_update=0` boot option. If you're
+using AutoYaST, is also possible to disable this feature using the `self_update`
+element in `general` section of the profile:
+
+   ```xml
+   <general>
+     <self_update config:type="boolean">false</self_update>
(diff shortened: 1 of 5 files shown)


doc/SELF_UPDATE.md
```diff
@@ -7,9 +7,18 @@ installation even after the media has been released. Check
 
 ## Disabling Updates
 
-Self-update is enabled by default. However, it can be disabled by setting
-`self_update=0` boot option. In the following sections it is assumed that this
-feature is enabled.
+Starting in SUSE Linux Enterprise 12 SP3, self-update is enabled by default.
+However, it can be disabled by setting `self_update=0` boot option. If you're
```
lslezak: I think we should document the priorities. E.g. what happens if both options (the boot option *and* the AY profile) are used together?

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -149,6 +148,16 @@ def disabled_in_linuxrc?
       Linuxrc.InstallInf("SelfUpdate") == "0"
     end
 
+    # Determines whether self-update feature is disable via AutoYaST profile
```
lslezak: Typo `disabled`

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -149,6 +148,16 @@ def disabled_in_linuxrc?
...
+    #
+    # @return [Boolean] true if self update has been disabled by AutoYaST profile
```
lslezak: I thin the `false` for non AY mode should be documented as well...

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -149,6 +148,16 @@ def disabled_in_linuxrc?
...
+    def disabled_in_profile?
+      return false unless Mode.auto
```
lslezak: This can be actually simplified, the AY mode is checked already in the caller, here I'd simply return the value from the profile (or `nil` if not specified). The behavior would be simpler, currently it checks for the mode and additionally treats missing key as `false`...

imobachgs: About the `auto` mode you're right: I moved the call to this method and I forgot about the `Model.auto`.

I think it's ok to reduce the returned value to only `true` and `false`. By default, the `self_update` is considered to be enabled.

doc/SELF_UPDATE.md
```diff
@@ -7,9 +7,21 @@ installation even after the media has been released. Check
...
+`self_update` element in the `general` section of the profile:
+
+   ```xml
+   <general>
+     <self_update config:type="boolean">false</self_update>
+   </general>
+   ```
+
+Please, take into account that self-update will be disabled if any of those
+values is set.
```
lslezak: Um, the word `set` is a bit misleading here, I guess it means if the self-update is disabled if any option disables the self-update.

With word `set` it sounds like if it is defined, meaning that `true` value does not work correctly.

imobachgs:
Thanks!

PR #502, opened 2017-01-26
All in one errors

jreidinger:


src/lib/installation/widgets/overview.rb
```diff
@@ -50,19 +54,38 @@ def label
...
             { "simple_mode" => true }
           ])
-        d["label_proposal"]
+        if d["warning"] && !d["warning"].empty? && d["warning_level"] != :notice
+          Yast::Popup.LongError(
+            "Problem found when proposing %{client}:<br>" \
+            "Severity: %{severity}<br>" \
+            "Message: %{message}" %
+            ({
+              client: proposal_client,
```
mvidner: `proposal_client` is an implementation detail so we should not show it to the user. `label` is human readable, and that also means we should memoize it like we do for `items`

jreidinger: good idea, I do not like it before.

src/lib/installation/clients/inst_casp_overview.rb
```diff
@@ -66,6 +74,15 @@ def run
           # do not store stuff when just redrawing
           skip_store_for: [:redraw]
         )
+        blocker = blocking_widgets.find(&:blocking?)
+        if blocker
+          # %title stands for title of problematic part
```
mvidner: Just use `%s` when there is only one string to substitute? Less room for errors both for us and the translator.

> %s is a heading of a problematic section, like "Partitioning" or "Network"

jreidinger: ok

src/lib/installation/clients/inst_casp_overview.rb
```diff
@@ -66,6 +74,15 @@ def run
...
+          Yast::Popup.Error(
+            _("%{title} blocks installation. Please solve problem here before proceeding.") %
```
mvidner: > %s blocks **the** installation. Please resolve **the** problem **there** before proceeding.

PR #504, opened 2017-01-27
Fix redraw for system roles, move software proposal and use real registration widget

jreidinger:


package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jan 27 09:11:32 UTC 2017 - jreidinger@suse.com
+
+- Respect newly registered update repos in software proposal and
```
mvidner: += Do register.

mvidner:
@jreidinger I don't understand why this PR apparently undoes the error handling that you added in #502

mvidner:
Aaah, it uses `validate` now!

jreidinger:
@mvidner yes, I use validate as more standard way to do it and also no need to redraw and redraw again when you click it

mvidner:
The code looks good. Let me test it with a DUD.

jreidinger:
I test it expect registration widget.

mvidner:
Thanks!
But it's still failing because it is calling the older User widget API that does not expect the additional flag we're passing.

PR #506, opened 2017-01-27
Added installation confirmation to all-in-one-dialog (FATE#322328)

teclator:
https://trello.com/c/XDgl2dj7/816-13-casp-new-all-in-one-dialog
https://fate.suse.com/322328
Main PR: https://github.com/yast/yast-installation/pull/487

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jan 27 15:55:33 UTC 2017 - kanderssen@suse.com
+
+- Ask for installation confirmation in all-in-one dialog before
+  proceed with it. (Fate#3222328)
+- 3.1.217.16
+
+-------------------------------------------------------------------
 Fri Jan 27 09:11:32 UTC 2017 - jreidinger@suse.com
 
 - Respect newly registered update repos in software proposal and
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/inst_casp_overview.rb
```diff
@@ -80,13 +80,16 @@ def run
...
+        # report problems with sofware proposal
+        if [:blocker, :fatal].include?(d["warning_level"])
+          # %s is a problem description
+          Yast::Popup.Error(
+            _("Software proposal failed. Cannot proceed with installation.\n%s") % d["warning"]
+          )
+        # continue only if confirmed
+        elsif Yast::WFM.CallFunction("inst_doit", []) == :next
+          break
+        end
```
jreidinger: code wise it looks good, but in general it is against idea of one click install. Please verify with @jsrain and @kobliha that it is what they really want.

mvidner: Good point, but I think we should have this.
If product mgmt decide to remove the brakes, we can revert this later.

package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jan 27 15:55:33 UTC 2017 - kanderssen@suse.com
+
+- Ask for installation confirmation in all-in-one dialog before
+  proceed with it. (Fate#3222328)
```
mvidner: Tooooo many 22222s.  Use (`FATE#322328`)

teclator: Probably that was because of CaaaaaaSP ;) I will fix it thnx for the catch.

PR #507, opened 2017-01-30
Implement all-in-one overview for CASP/CaaSP (FATE#322328) (CASP->master)

mvidner:
Merge #487 ... #506

### DIFF
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.2.17
+Version:        3.2.18
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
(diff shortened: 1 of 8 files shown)


package/yast2-installation.spec
```diff
@@ -104,8 +104,8 @@ Conflicts:	yast2-core < 2.17.10
 # Top bar with logo
 Conflicts:	yast2-ycp-ui-bindings < 3.1.7
 
-# RegserviceSelectionDialog
-Conflicts:  yast2-registration < 3.1.179
+# new registration widget
+Conflicts:  yast2-registration <= 3.1.190
```
teclator: Why not 3.2.1 added in yast/yast-registration#295

mvidner:
@teclator Thank you for the fix!
Hmm,  I don't like that we are adding some buildrequires. I think I'll try removing them and mocking the APIs instead... BUT that should not block merging this.

PR #510, opened 2017-02-01
Casp services

jreidinger:
tested manually

https://fate.suse.com/321738

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Feb  1 14:20:43 UTC 2017 - jreidinger@suse.com
+
+- Enable CaaSP specific services on the installed system
+  (FATE#321738)
+- 3.1.217.18
+
+-------------------------------------------------------------------
 Tue Jan 31 16:06:40 UTC 2017 - jreidinger@suse.com
 
 - Do not crash if role contain own service to enable (bsc#1022762)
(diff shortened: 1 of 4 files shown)


imobachgs:
The code looks ok but I'm not sure if I like this solution:

1) I don't like having a hardcoded list of services. I would move the list to the control file.
2) As it's only in the one-dialog, AutoYaST won't benefit of it. But maybe AutoYaST is out of the scope here.

Well, that's just an opinion but I would like to discuss it :)

jreidinger:
@imobachgs well, if it will be in control file it means we need to adapt control file schema and also control file beside others, but on other hand we need to add such support as generic support and not as caasp only. And if we do it, then there is high risk, that they start to use it instead of systemd-presets ( which in PBI is mentioned that for this purpose cannot be used ). So I would like to keep it as exception for caasp only as it have tricky services.
Regarding autoyast, it is mentioned, that autoyast is not affected, as it should have these services in autoyast template.

mvidner:
https://fate.suse.com/321738 explicitly calls for the list of the services to come from the control file. Neither the code nor the commit message explain why we hardcode it instead.

jreidinger:
@mvidner I added comment and to be honest I reuse this FATE from roles services PBI. This request come via instant messaging from kukuk and to be honest original proposal was to run it in shell script :) This way it can be at least debugged.

imobachgs:
The dialog is CaaSP specific and won't be reusable at all. But now it's documented so I'm fine. The code looks good and I'm only missing a unit test :)

PR #511, opened 2017-02-01
release download: break downloading if the host/proxy cannot be resolved

schubi2:


src/lib/installation/clients/inst_download_release_notes.rb
```diff
@@ -106,16 +106,20 @@ def download_release_notes
           )
           ret = SCR.Execute(path(".target.bash"), cmd)
           log.info("Downloading release notes: #{cmd} returned #{ret}")
+          # exit codes (see "man curl"):
+          skip_ret = {
```
imobachgs: I would use a constant. In fact, the [master branch version of this file](https://github.com/yast/yast-installation/blob/master/src/lib/installation/clients/inst_download_release_notes.rb#L36) uses one. What about merging them (I mean, using a constant but with the values you proposed)?

PR #513, opened 2017-02-03
Configure minion master location (Fate#321738)

teclator:
trello card: https://trello.com/c/hBDzHqGs

Maybe starts to make sense an specific class for system roles where store the role state. By now I just obtain the current selection from the widgets when finish is call.

Field validation:

caspi_validation

After installation:

configwritten

**[UPDATE]**

At the end we have decided to add the specific SystemRole class.


src/lib/installation/clients/roles_finish.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+
+      def write
+        log.info("The current role is: #{current_role}")
+        if current_role == "worker_role"
+          master_conf = ::Installation::CFA::MinionMasterConf.new
+          begin
+            master_conf.load
+          rescue Errno::ENOENT
+            log.info("Fail does not exist, will be created")
```
mchf: i'd prefer more accurate log message.

src/lib/installation/clients/roles_finish.rb
```diff
@@ -0,0 +1,63 @@
...
+          master_conf.master = master
+          master_conf.save
+        end
+      end
+
+    private
+
+      # Obtains the current role from the role selection widget
+      def current_role
+        ::Installation::Widgets::SystemRole.original_role_id
```
mchf: ehm, is ```::Installation``` needed? I've thought, we're in same namespace here ... I just don't like those ```::``` spaghetti.

src/lib/installation/widgets/system_role.rb
```diff
@@ -25,23 +25,46 @@
...
       def store
-        # TODO: implement it together with init and some validation
+        self.class.location = value
+      end
+
+      def init
+        self.value = self.class.location
+      end
+
+      def validate
```
mchf: missing a documentation. At least it should mention that it can raise a popup.

src/lib/installation/clients/roles_finish.rb
teclator: I was modifying the namespaces and was using absolute path at the beginning

src/lib/installation/widgets/system_role.rb
```diff
@@ -48,6 +48,10 @@ def init
         self.value = self.class.location
       end
 
+      # If the value is not a valid IP or a valid FQDN it display an error
```
mchf: AFAIK, "displays"

mchf:
would it be possible to add some "smoke" test?

teclator:
Sure, it is still in RFC, so if you are fine with implementation I will add them.

src/lib/installation/widgets/system_role.rb
```diff
@@ -40,16 +40,20 @@ def label
         "Controller Node"
       end
 
+      # It stores the content of the value if validates
```
mchf: ehm ... what value?

teclator: It is an InputField widget so its value...

src/lib/installation/clients/roles_finish.rb
```diff
@@ -0,0 +1,62 @@
...
+# current contact information at www.suse.com.
+# ------------------------------------------------------------------------------
+
+require "installation/finish_client"
+require "installation/widgets/system_role"
+require "installation/cfa/salt"
+
+module Installation
+  module Clients
+    class RolesFinish < ::Installation::FinishClient
```
mchf: i'm trying to play martin's role here ... but it would be nice to have a high level desc of the class

src/lib/installation/clients/roles_finish.rb
```diff
@@ -25,6 +25,11 @@
 
 module Installation
   module Clients
+    # This is a step of base installation finish and is reponsable of write the
```
mchf: NP: responsible ;-)

teclator: arff, this is one of my recurrent typos :(

src/lib/installation/clients/roles_finish.rb
```diff
@@ -0,0 +1,67 @@
...
+require "installation/cfa/salt"
+
+module Installation
+  module Clients
+    # This is a step of base installation finish and is responsible of write the
+    # specific configuration for the current system role.
+    #
+    # It has been added for CaaSP Roles (FATE#321754) and currently only
+    # the 'worker_role' has an special behavior.
+    class RolesFinish < ::Installation::FinishClient
```
jreidinger: To be honest I do not like this. In general I think it is not good idea to query widget from finish client. Also do not forget that it is run also from autoinstallation. I think the best way to achieve it, is to create in lib configuration file, that holds info what needs to be done and then finish client if needed, run it.
Also name indicate it is needed for all roles, but this is caasp specific one. So I prefer own name, like salt_finish.

src/modules/InstData.rb
```diff
@@ -109,6 +110,7 @@ def main
     publish variable: :start_mode, type: "string"
     publish variable: :selected_desktop, type: "string"
     publish variable: :current_systasks_status, type: "map <string, boolean>"
+    publish variable: :current_role_options, type: "map <string, string>"
```
jreidinger: why it is needed?

src/lib/installation/clients/roles_finish.rb
teclator: I also don't like to query the widgets explicitly, and during the development first idea was to use InstData but then just use the current widgets as it already was remembering the selected role.

It is not needed for all roles as currently only do something special for the worker role, but probably it will not be salt specific and more things will be needed based on the role.

Which makes sense for me is to have a special class for the SystemRole which maintains the role selected and other things that we are currently defining in the profile like the services.

The role could implement the method finish what could be called from this client, so just instance the specifc class based on role name and call the finish method of it.

src/modules/InstData.rb
teclator: sorry, was added as commented before for remember the selected role and other options and not ask the widgets directly, but as that info was already handled by them just decided to use by now and then refactor when needed.

src/lib/installation/clients/roles_finish.rb
teclator: auto-installation from other hand does not use roles, so it could be skipped in that case

jreidinger: @teclator I am looking at it from https://github.com/yast/yast-installation/pull/515 perspective, where we have different widget for roles, so reusability is harder. And it also tight this finish client and SystemRole widget. So idea with special class for system roles is fine for me.

teclator: @jreidinger I have updated the PR based on your comments

src/lib/installation/clients/roles_finish.rb
```diff
@@ -38,7 +38,7 @@ def title
 
       def write
         # Finish installation for the selected role
-        SystemRole.installation_finish
+        SystemRole.finish(SystemRole.current) if SystemRole.current
```
jreidinger: why not finish without parameter that handle inside if it is current role nil?

src/lib/installation/system_role.rb
```diff
@@ -22,51 +22,132 @@
...
 
-    def initialize
-      @roles = Yast::ProductControl.productControl.fetch("system_roles", [])
-      @selected = nil
-      @options = {}
+    def initialize(id, label, description, services, options = {})
+      @id          = id
+      @label       = label
+      @description = description
+      @services    = services || []
```
jreidinger: if services can be nil, why not also use default parameter? and in general, why not use named parameters?

src/lib/installation/clients/roles_finish.rb
teclator: just to avoid the call, but could be, and was also my first option

src/lib/installation/system_role.rb
```diff
@@ -22,51 +22,132 @@
...
+          Object.const_get(handler).run
+        rescue NameError, LoadError
+          log.info("There is no special finisher for #{role_id}")
+        end
+      end
     end
 
-    def services
-      all.find { |r| r["id"] == selected  }["services"] || []
+    def option(*args)
```
jreidinger: to be hnest I do not get goal of this method, it looks strange from my POV

src/lib/installation/system_role.rb
```diff
@@ -0,0 +1,162 @@
...
+
+    def overlay_features
+      features = self.class.raw_roles.find { |r| r["id"] == id }.dup
+
+      NON_OVERLAY_ATTRIBUTES.each { |a| features.delete(a) }
+      Yast::ProductFeatures.SetOverlay(features)
+    end
+  end
+
+  module SystemRoleHandlers
```
imobachgs: I would move this module (and its classes) to a separated directory.

teclator: yep, as it was only one I maintained it here, but could be moved to a system_roles directory in handlers.rb

src/lib/installation/system_role.rb
teclator: basically it is for set or get any option abstracting the user from the implementation and without inheritance

option("controller_node") => "salt.suse.de" (GET)
option("controller_node", "salt") => "salt"     (SET)

@imobachgs also propose set_option or get_option, some proposal?

jreidinger: @teclator I think separated methods are much better and cleaner. another option in my mind is `option(key)/assign(key, value)`

src/lib/installation/clients/roles_finish.rb
```diff
@@ -0,0 +1,44 @@
...
+    # the 'worker_role' has an special behavior.
+    class RolesFinish < ::Installation::FinishClient
+      def title
+        textdomain "installation"
+        _("Writing specific role configuration  ...")
+      end
+
+      def write
+        # Finish installation for the selected role
+        SystemRole.finish(SystemRole.current)
```
jreidinger: is there any use case to finish it with other then current role? if not, then parameter is useless

src/lib/installation/system_role.rb
```diff
@@ -0,0 +1,169 @@
...
+        if !role_id
+          log.info("There is no role selected so nothing to do")
+          return
+        end
+
+        class_name_role = role_id.split("_").map(&:capitalize).join
+        handler = "::Installation::SystemRoleHandlers::#{class_name_role}Finish"
+
+        begin
+          Object.const_get(handler).run
```
jreidinger: I expect that there is not so often to have handler and I really dislike having exception as part of common workflow. What about
```
if Object.const_defined?(handler)
  Object.const_get(handler).run
else
  log.info("There is no special finisher for #{role_id}")
end
```

src/lib/installation/system_role.rb
```diff
@@ -0,0 +1,169 @@
...
+      end
+    end
+
+    def [](key)
+      @options[key]
+    end
+
+    def []=(key, value)
+      @options[key] = value
+    end
```
jreidinger: NP: these two methods are in fact delegation, unless you plan to add some logging

src/lib/installation/system_roles/handlers.rb
```diff
@@ -0,0 +1,42 @@
...
+module Installation
+  module SystemRoleHandlers
+    class WorkerRoleFinish
+      def self.run
+        role = SystemRole.find("worker_role")
+        master_conf = CFA::MinionMasterConf.new
+        master = role["controller_node"]
+        begin
+          master_conf.load
+        rescue Errno::ENOENT
```
jreidinger: yeah, I plan to add to CFA file handler method `exist?` that can be used to query if file exists.

src/lib/installation/widgets/system_role.rb
```diff
@@ -22,32 +22,67 @@
...
+      def init
+        self.value = role["controller_node"]
+      end
+
+      # If the value is not a valid IP or a valid FQDN it displays a popup
+      # error and returns false, in other case it just returns true.
+      #
+      # @return <Boolean> false if not a valid IP or FQDN; true otherwise
+      def validate
+        unless Yast::IP.Check(value) || Yast::Hostname.CheckFQ(value)
```
jreidinger: what about `return true if Yast::IP.Check(value) || Yast::Hostname.CheckFQ(value)` and continue with error?

src/lib/installation/widgets/system_role.rb
```diff
@@ -75,75 +105,44 @@ def opt
       end
 
       def init
-        self.class.original_role_id ||= roles_description.first[:id]
-        self.value = self.class.original_role_id
+        self.value = ::Installation::SystemRole.current
```
jreidinger: this breaks previous behavior. request was to select first one when first shown. In your case it is nil and nothing is shown.

teclator: true

src/lib/installation/system_role.rb
```diff
@@ -0,0 +1,169 @@
...
+
+        begin
+          Object.const_get(handler).run
+        rescue NameError, LoadError
+          log.info("There is no special finisher for #{role_id}")
+        end
+      end
+    end
+
+    def [](key)
```
imobachgs: np: documentation

src/lib/installation/system_role.rb
```diff
@@ -0,0 +1,169 @@
...
+      @options[key]
+    end
+
+    def []=(key, value)
```
imobachgs: Documentation

src/lib/installation/system_role.rb
```diff
@@ -0,0 +1,169 @@
...
+      @options[key] = value
+    end
+
+    # Enables the role services defined in the control file
+    def adapt_services
```
imobachgs: I'm not sure if it belongs to this class. But I cannot think of a better place at this time.

src/lib/installation/system_role.rb
```diff
@@ -0,0 +1,169 @@
...
+      "additional_dialogs",
+      "id",
+      "services"
+    ].freeze
+
+    private_constant :NON_OVERLAY_ATTRIBUTES
+
+    # Overlays the product features for this role with the configuration
+    # obtained from the control file
+    def overlay_features
```
imobachgs: Same as above. But, for the time being, it's fine as it is.

src/lib/installation/cfa/salt.rb
```diff
@@ -9,6 +9,9 @@ class MinionMasterConf < ::CFA::BaseModel
       attributes(master: "master")
 
       # Configuration parser
+      #
+      # FIXME: At this time, we're using Augeas' cobblersettings lense,
+      # as long as YAML is not supported. We should use another parser.
```
jreidinger: any link why YAML is not supported? we need some resource to verify if it can be changed or not.

imobachgs: To be precise, YAML is supported but it expects the three dashes (`---`) at the beginning of the file. Am I wrong?

src/lib/installation/system_role.rb
```diff
@@ -32,14 +32,30 @@ module Installation
   # installation.
   class SystemRole
     include Yast::Logger
-
-    attr_accessor :id, :label, :description, :services
-
-    def initialize(id:, label: id, description: nil, services: [])
+    extend Forwardable
+
+    attr_accessor :id, :label, :description
```
jreidinger: still documentation for attributes missing - see for example https://github.com/yast/yast-bootloader/blob/e0ce36408473ea7d88de1eb54fdda7696b6585ec/src/lib/bootloader/sysconfig.rb#L16

src/lib/installation/system_role.rb
```diff
@@ -66,6 +85,11 @@ def all
       end
 
       # Fetchs the roles from the control file and returns them as they are.
+      #
+      # @example
+      #   SystemRole.raw_roles #=> [{ "id" => "role_one" }]
+      #
+      # @return [Array<Hash>]
```
jreidinger: specify that it return empty array if not defined in control file. It is always good to know defaults, that it cannot return nil

src/lib/installation/system_role.rb
```diff
@@ -32,14 +32,30 @@ module Installation
...
+
+    attr_accessor :id, :label, :description
+
+    # All the special attributes for a given role are delegated to @options
+    def_delegators :@options, :[], :[]=
+
+    # Constructor
+    #
+    # Only the id, label and description are allowed to be initialized other
+    # options has to be set explicitly.
```
imobachgs: np: have

src/lib/installation/system_role.rb
```diff
@@ -32,14 +32,30 @@ module Installation
...
+    #
+    # @example SystemRole with 'services' and a 'controller node' defined
+    #   @role = SystemRole.new(id: "test_role")
```
imobachgs: 👍

src/lib/installation/cfa/salt.rb
teclator: Although it is yaml format it does not use the headers --- and it can't be used, this class is basically reused from previous work for configuration management https://github.com/imobachgs/yast-cm/blob/master/src/lib/cm/cfa/minion.rb

src/lib/installation/widgets/system_role.rb
```diff
@@ -50,21 +50,20 @@ def init
         self.value = role["controller_node"]
       end
 
-      # If the value is not a valid IP or a valid FQDN it displays a popup
-      # error and returns false, in other case it just returns true.
+      # It returns true if the value is a valid IP or a valid FQDN, if not it
+      # displays a popup error.
       #
-      # @return <Boolean> false if not a valid IP or FQDN; true otherwise
+      # @return <Boolean> true if valid IP or FQDN
```
imobachgs: Use `[]` instead of `<>`. I mean: `@return [Boolean] ...`.

src/lib/installation/cfa/salt.rb
jreidinger: yep, this needs to be in comment, so we know it what we looking if we can finally fix FIXME :)

test/system_role_test.rb
```diff
@@ -0,0 +1,125 @@
...
+      }
+    ]
+  end
+
+  before do
+    allow(described_class).to receive(:raw_roles).and_return(system_roles)
+  end
+
+  describe ".raw_roles" do
+    it "fetchs the roles from the control file and returns them" do
```
imobachgs: np: You could simply say: "returns the roles from the control file".

src/lib/installation/cfa/salt.rb
teclator: Exactly, I commented it as an answer to a previous @jreidinger comment. And we were talking about it some days ago, you suggested that maybe in the future we could just reuse the yaml.lns without the header condition

src/lib/installation/system_role.rb
```diff
@@ -34,23 +34,28 @@ class SystemRole
...
+    attr_accessor :description
 
     # All the special attributes for a given role are delegated to @options
     def_delegators :@options, :[], :[]=
 
     # Constructor
     #
-    # Only the id, label and description are allowed to be initialized other
-    # options has to be set explicitly.
+    # Only the id, label and description are allowed to be initialized others
```
imobachgs: np: other :)

package/yast2-installation.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Feb  8 16:42:29 UTC 2017 - kanderssen@suse.com
+
+- CaaSP all-in-one-dialog: added validation to the controller node
+  location field, modified label and write it to the minion
+  master.conf file at the end of the installation
```
jreidinger: ID is missing, bnc or fate number

teclator: Good catch, originally I used this Fate#321738 for the PR I will used it also for the changelog, but will ask @jsrain .

src/lib/installation/system_role.rb
```diff
@@ -0,0 +1,201 @@
...
+
+module Installation
+  # This class fetchs and stores the roles declared in the installation control
+  # file. It works as a storage for them and for the role selected during the
+  # installation.
+  class SystemRole
+    include Yast::Logger
+    extend Forwardable
+
+    # @return [String]
```
jreidinger: I expect here a bit more, like what id, what it represents, is it id of role or id of object? same for label and description. Like is description and label localized? and what about id?

teclator: for label and description I wanted to make it open, although we translate them from the control file, for id of course I could add a comment about saying that is role specific id

package/yast2-installation.changes
teclator: Done

teclator:
@jreidinger @imobachgs @mchf thnx for the code reviews!

PR #514, opened 2017-02-03
Downloading release notes: Do not download them at all during AutoYaST and use the directory.yast file to check avialable translations (bsc#1009276)

jsrain:


src/lib/installation/clients/inst_download_release_notes.rb
```diff
@@ -38,6 +38,47 @@ class InstDownloadReleaseNotesClient < Client
...
+    # @param proxy the proxy URL to be passed to curl
+    #
+    # @return true when successful
+    def download_release_notes_index(url_base, proxy)
+      url_index = url_base + "/directory.yast2"
+      log.info("Index with available files: #{url_index}")
+      filename = Builtins.sformat("%1/directory.yast", SCR.Read(path(".target.tmpdir")))
+      # download the index with much shorter time-out
+      cmd = Builtins.sformat(
+        "/usr/bin/curl --location --verbose --fail --max-time 30 --connect-timeout 15  %1 '%2' --output '%3' > '%4/%5' 2>&1",
```
lslezak: Some parts are similar to the `download_release_notes` function below. I'd suggest to share some code, at least the `curl` command building. If we need to change e.g. the timeout defaults we will need to touch two places which is fragile.

src/lib/installation/clients/inst_download_release_notes.rb
```diff
@@ -38,6 +38,47 @@ class InstDownloadReleaseNotesClient < Client
...
+        url_index,
+        String.Quote(filename),
+        String.Quote(Directory.logdir),
+        "curl_log"
+      )
+      ret = SCR.Execute(path(".target.bash"), cmd)
+      log.info("Downloading release notes index: #{cmd} returned #{ret}")
+      if ret == 0
+        log.info("Release notes index downloaded successfully")
+        index_file = SCR.Read(path(".target.string"), filename)
```
lslezak: I think you can use plain `File.read(filename)` here.

src/lib/installation/clients/inst_download_release_notes.rb
```diff
@@ -38,6 +38,47 @@ class InstDownloadReleaseNotesClient < Client
...
+        if index_file.nil? || index_file.empty?
+          log.info("Release notes index empty, not filtering further downloads")
+          return nil
```
lslezak: `return nil` is repeated several times, I think it's enough to put it once at the end of the function.

src/lib/installation/clients/inst_download_release_notes.rb
```diff
@@ -96,9 +137,24 @@ def download_release_notes
...
+          log.info("Skipping release notes download due to previous download issues")
+          break
+        end
+
         url_template = url_base + filename_templ
         log.info("URL template: #{url_base}")
         [Language.language, Language.language[0..1], "en"].uniq.each do |lang|
+          if !rn_filter.nil?
+            filename = Builtins.sformat(filename_templ, lang)
+            filename = Builtins.substring(filename, 1)
```
lslezak: `Builtins` should be avoided where possible, either use [this Ruby trick](http://stackoverflow.com/questions/3614389/what-is-the-easiest-way-to-remove-the-first-ch) for removing the first character of or remove this line and change the condition to `if !rn_filter.include?(filename[1..-1])`.

package/yast2-installation.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Fri Feb  3 10:07:35 UTC 2017 - jsrain@suse.cz
+
+- Downloading release notes: Do not download them at all during
+  AutoYaST (bsc#1009276)
+- Use the directory.yast file to check avialable
```
lslezak: Typo `available`.

src/lib/installation/clients/inst_download_release_notes.rb
jsrain: I intentionally have shorter time-out here

src/lib/installation/clients/inst_download_release_notes.rb
jsrain: Would be, but I find some value in explaining why nil is returned - together with the comnent above.

src/lib/installation/clients/inst_download_release_notes.rb
lslezak: OK

PR #517, opened 2017-02-10
[RFC] Do not reset all-in-one dialog when sub dialog is closed

mchf:
[bsc#1022546](https://bugzilla.suse.com/show_bug.cgi?id=1022546)

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Mon Feb 13 14:49:14 UTC 2017 - mfilka@suse.com
+
+- bnc#1022546
+  - update all-in-one's subdialog (network/kdump/partitioning/...)
+    overview when the dialog is closed and do not lose another
+    content already entered by user.
+- 3.1.217.23
+
+-------------------------------------------------------------------
 Mon Feb 13 14:12:59 CET 2017 - schubi@suse.de
 
 - Calling services-manager_finish in order to write AutoYaST
--- a/src/lib/installation/clients/inst_casp_overview.rb
+++ b/src/lib/installation/clients/inst_casp_overview.rb
@@ -68,12 +68,12 @@ module Installation
         ret = Yast::CWM.show(
           content,
           # Title for installation overview dialog
-          caption:        _("Installation Overview"),
+          caption:      _("Installation Overview"),
           # Button label: start the installation
-          next_button:    _("Install"),
+          next_button:  _("Install"),
           # do not show abort and back button
-          abort_button:   "",
-          back_button:    ""
+          abort_button: "",
+          back_button:  ""
         )
 
         # TODO: check exception type
@@ -147,7 +147,7 @@ module Installation
       controller_node = Installation::Widgets::ControllerNodePlace.new
 
       kdump_overview = Installation::Widgets::Overview.new(client: "kdump_proposal")
-      bootloader_overview = Installation::Widgets::Overview.new(client: "bootloader_proposal", on_redraw: [ kdump_overview ])
+      bootloader_overview = Installation::Widgets::Overview.new(client: "bootloader_proposal", on_redraw: [kdump_overview])
 
       quadrant_layout(
         upper_left:  VBox(
@@ -162,7 +162,7 @@ module Installation
           Tune::Widgets::SystemInformation.new
         ),
         upper_right: VBox(
-          Installation::Widgets::Overview.new(client: "partitions_proposal", on_redraw: [ bootloader_overview ]),
+          Installation::Widgets::Overview.new(client: "partitions_proposal", on_redraw: [bootloader_overview]),
           bootloader_overview
         ),
         lower_right: VBox(
--- a/src/lib/installation/clients/inst_casp_overview.rb
+++ b/src/lib/installation/clients/inst_casp_overview.rb
@@ -146,8 +146,8 @@ module Installation
     def content
       controller_node = Installation::Widgets::ControllerNodePlace.new
 
-      bootloader_overview = Installation::Widgets::Overview.new(client: "bootloader_proposal")
       kdump_overview = Installation::Widgets::Overview.new(client: "kdump_proposal")
+      bootloader_overview = Installation::Widgets::Overview.new(client: "bootloader_proposal", on_redraw: [ kdump_overview ])
 
       quadrant_layout(
         upper_left:  VBox(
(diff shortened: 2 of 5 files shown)


src/lib/installation/clients/inst_casp_overview.rb
```diff
@@ -69,11 +69,11 @@ def run
...
           abort_button:   "",
-          back_button:    "",
-          # do not store stuff when just redrawing
-          skip_store_for: [:redraw]
+          back_button:    ""
         )
-        next if ret == :redraw
+
+        # TODO: check exception type
+        raise ArgumentError, "Unexpected return value" if ret != :next
```
jreidinger: that looks wrong, CWM can return also back and abort.

jreidinger: but it is not there, so probably make sense

src/lib/installation/widgets/overview.rb
```diff
@@ -80,9 +78,18 @@ def items
         @items = d["label_proposal"]
       end
 
+      def redraw
+        raise ArgumentError, "Missing replace point reference, cannot update the overview" if !@replace_point
```
jreidinger: that looks really wrong, you initialize it in constructor, so how it can be nil? that looks very paranoic

src/lib/installation/widgets/overview.rb
```diff
@@ -107,6 +114,19 @@ def button_id
...
+
+      def widget
+        VBox(
+          Left(PushButton(Id(button_id), label)),
+          * items.map { |i| Left(Label(" * #{i}")) }
+        )
+      end
+
+      # Flush cache(s)
+      def reset
```
jreidinger: what is purpose of this method?

src/lib/installation/widgets/overview.rb
```diff
@@ -80,9 +78,18 @@ def items
...
+        raise ArgumentError, "Missing replace point reference, cannot update the overview" if !@replace_point
+
+        Yast::UI.ReplaceWidget(Id(@replace_point), widget)
+      end
+
       def handle(_event)
         Yast::WFM.CallFunction(proposal_client, ["AskUser", {}])
-        :redraw
+
+        redraw
```
jreidinger: well, this have one problem. As I write to PBI problem is that result of AskUser can also modify result of other overviews. so in such case reset and then call to redraw make sense. But I suggest to use reinit instead of reset and reinit simple call again this proposal client and fills items and blocking.

src/lib/installation/widgets/overview.rb
mchf: yes, it is paranoic ... I'll drop it

src/lib/installation/widgets/overview.rb
mchf: ah, ok. Will study internals more.

mchf: updating other overviews is TBD.

src/lib/installation/clients/inst_casp_overview.rb
```diff
@@ -151,12 +155,12 @@ def content
           Tune::Widgets::SystemInformation.new
         ),
         upper_right: VBox(
-          Installation::Widgets::Overview.new(client: "partitions_proposal"),
-          Installation::Widgets::Overview.new(client: "bootloader_proposal")
+          Installation::Widgets::Overview.new(client: "partitions_proposal", on_redraw: [ kdump_overview, bootloader_overview ]),
```
jreidinger: I think that kdump also depends on bootloader. maybe it should be recursive...so partitioning > bootloader > kdump. So if bootloader changed, redraw kdump. if partitioning changed redraw bootloader and that cause kdump redraw

mchf: that's the main problem .. i don't know what depends on what ;-)

src/lib/installation/clients/inst_casp_overview.rb
mvidner: I don't quite see the point of this. Anyway, `ret` is not an argument, so I would use `RuntimeError` instead of `ArgumentError`.

mchf: I've used that mainly during development. I'll drop it.

src/lib/installation/widgets/overview.rb
```diff
@@ -33,20 +33,22 @@ class Overview < CWM::CustomWidget
       attr_reader :proposal_client
 
       # @param client [String] A proposal client implementing simple_mode,
+      # @param redraw [Array<Overview>] list of other Overview clients. In case
+      # of :redraw action every of these clients will be redrawn too. Caller is
+      # responsible for not creating circular dependencies.
       #   eg. "bootloader_proposal"
-      def initialize(client:)
+      def initialize(client:, on_redraw: [])
```
mvidner: Please rename `on_redraw` to `redraw`
(or make it a lambda: `on_redraw: -> { kdump_overview.redraw }`)

BTW this looks a bit like the [Observer pattern, for which there is a standard library module](http://www.rubydoc.info/stdlib/observer/Observable).

mvidner:
Please don't forget the .changes

src/lib/installation/widgets/overview.rb
```diff
@@ -33,20 +33,22 @@ class Overview < CWM::CustomWidget
...
       #   eg. "bootloader_proposal"
-      def initialize(client:)
+      def initialize(client:, redraw: [])
         textdomain "installation"
         @proposal_client = client
+        @replace_point = "rp_" + client
         # by default widget_id is the class name; must differentiate instances
         self.widget_id = "overview_" + client
         @blocking = false
+        @overviews_for_redraw = on_redraw
```
mvidner: boom

package/yast2-installation.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Mon Feb 13 14:49:14 UTC 2017 - mfilka@suse.com
+
+- bnc1022546
+  - update all-in-one's subdialog (network/kdump/partitioning/...)
+    overview when the dialog is closed and do not loose another
```
mvidner: lose

package/yast2-installation.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Mon Feb 13 14:49:14 UTC 2017 - mfilka@suse.com
+
+- bnc1022546
```
mvidner: bnc1 -> bnc#1 please:
```console
$ rake check:changelog
...
Stopping, missing new bugzilla or fate entry in the *.changes file.
e.g. bnc#<number>, fate#<number>, boo#<number>, bsc#<number>, bgo#<number>, cve-<number>
```

mchf:
@mvidner typos fixed

mchf:
thanks for reviews

PR #521, opened 2017-02-13
Fix dashed url (bsc#1024965)

teclator:
This is an easy fix allowing dashes url although the best option is to use a custom salt lense. It is valid because we don't expect to update the file but just create it.

### DIFF
--- a/src/lib/installation/system_roles/handlers.rb
+++ b/src/lib/installation/system_roles/handlers.rb
@@ -19,11 +19,14 @@
 # current contact information at www.suse.com.
 # ------------------------------------------------------------------------------
 
+require "yast"
 require "installation/cfa/salt"
 
 module Installation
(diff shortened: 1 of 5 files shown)


src/lib/installation/system_roles/handlers.rb
```diff
@@ -19,11 +19,14 @@
 # current contact information at www.suse.com.
 # ------------------------------------------------------------------------------
 
+require "yast"
 require "installation/cfa/salt"
 
 module Installation
   module SystemRoleHandlers
     class WorkerRoleFinish
+      include Yast::Logger
```
jreidinger: why logger is needed?

teclator: Well log.info is ussed

PR #522, opened 2017-02-14
Self-update refactoring

imobachgs:
## Improving the self-update feature usability

As you may know, YaST includes a [nice feature](https://fate.suse.com/319716) which allows the installer to update itself in order to solve bugs in the installation media. This mechanism has been included in SUSE Linux Enterprise 12 SP2, although it's not enabled by default (you need to pass an extra option `selfupdate=1` to make it work).

So after getting some feedback, we're working toward fixing some usability problems. The first of them is that, in some situations, the self-update mechanism is too intrusive.

Consider the following scenario: you're installing a system behind a firewall which prevents the machine to connect to the outside network. As the SUSE Customer Center will be unreachable, YaST complains about not being able to get the list of repositories for the self-update. And, after that, you get another complain because the fallback repository is not accessible. Two error messages and 2 timeouts.

And the situation could be even worse if you don't have access to a DNS server (add another error message).

So after some discussion we've decided to show such errors only if the user has specified SMT or a custom self-update repository (which is also possible). In any other case, the error is logged and the self-update is skipped completely.

You can find further information in our [Self-Update Use Cases and Error Handling dcument](https://gist.github.com/imobachgs/cb03b6a87dd5e31c89a9463d30db4252).

During upcoming sprints, we'll keep working on making the self-update feature great!

## Not for the blog

The idea of this PR is to extract some logic from the InstUpdateInstaller because, IMHO, it has too many responsabilities.

* `UpdateRepositoriesFinder` takes care of finding the update repositories (checking `Linuxrc`, the control file, SCC, etc.). It returns an array of `UpdateRepository` objects instead of just plain strings.
* `UpdateRepository` gets a new method `user_defined?` which determines whether is a user defined repository or not. Additionally, it also gets a `remote?` method and that responsability is also removed from the client.
* Storing and loading the used registration server is still done by the client.

### DIFF
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -231,7 +231,7 @@ module Yast
 
     rescue ::Installation::UpdatesManager::CouldNotFetchUpdateFromRepo
       if repo.user_defined?
-      # TRANSLATORS: %s is an URL
+        # TRANSLATORS: %s is an URL
         Report.Error(format(_("Could not fetch update from\n%s.\n\n"), repo.uri))
       end
       false
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -24,7 +24,6 @@ module Yast
     include Yast::I18n
 
     UPDATED_FLAG_FILENAME = "installer_updated".freeze
-    REMOTE_SCHEMES = ["http", "https", "ftp", "tftp", "sftp", "nfs", "nfs4", "cifs", "smb"].freeze
     PROFILE_FORBIDDEN_SCHEMES = ["label"].freeze
     REGISTRATION_DATA_PATH = "/var/lib/YaST2/inst_update_installer.yaml".freeze
 
@@ -106,7 +105,7 @@ module Yast
     # * Could not fetch update from repository: report the user about
     #   this error.
     # * Repository could not be probed: suggest checking network
-    #   configuration if URL has a REMOTE_SCHEME.
+    #   configuration if URL is remote.
     #
     # @return [Boolean] true if installer was updated; false otherwise.
     def update_installer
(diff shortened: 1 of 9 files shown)


src/lib/installation/update_repository.rb
```diff
@@ -118,6 +134,7 @@ def initialize(uri, instsys_parts_path = Pathname("/etc/instsys.parts"))
     # @see Yast::Pkg.ResolvableProperties
     def packages
       return @packages unless @packages.nil?
+      add_repo if repo_id.nil?
```
lslezak: The `add_repo` is already called inside `repo_id`, moreover `repo_id` cannot return `nil`, it either returns the ID or raises an exception. It looks like some initialization method should be used here...

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -162,27 +171,11 @@ def disabled_in_profile?
...
-      # load the base product from the installation medium,
-      # the registration server needs it for evaluating the self update URL
-      add_installation_repo
-      @default_self_update_urls = self_update_url_from_connect
-      return @default_self_update_urls unless @default_self_update_urls.empty?
-      @default_self_update_urls = Array(self_update_url_from_control)
+    def update_repositories
+      return @update_repositories if @update_repositories
+      @update_repositories = update_repositories_finder.updates
+      log.info("self-update repositories are #{@update_repositories.inspect}")
```
lslezak: Be careful about logging the repositories, if you use a custom repository then the URL might contain sensitive information (password). So I think the `UpdateRepository` class should implement some custom logging, see the recent discussion at yast-devel@.

src/lib/installation/update_repository.rb
```diff
@@ -220,6 +247,32 @@ def empty?
...
+    #
+    # @see Pkg.UrlSchemeIsRemote
+    def remote?
+      Yast::Pkg.UrlSchemeIsRemote(uri.scheme)
+    end
+
+    # Redefines the inspect method to avoid logging passwords
+    #
+    # @return [String] Debugging information
+    def inspect
```
lslezak: I think the `to_s` should redefined as well to handle also the simple `log.info "#{update_repo}"` case.

lslezak: Oh, ignore this, it seems that in this case the value is not logged...

imobachgs: Yes, it's not needed, but I've added a test to be on the safe side with `inspect` and `to_s`.

PR #523, opened 2017-02-14
adapted requires

schubi2:


package/yast2-installation.spec
```diff
@@ -73,7 +73,7 @@ Requires:	initviocons
 Requires:       yast2-proxy
 
 # Systemd default target and services
```
imobachgs: Maybe you could update this comment? Why do we need this particular version?

PR #532, opened 2017-03-06
Merge of SLE-12-SP2_CASP into master

mchf:


package/yast2-installation.changes
```diff
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Thu Mar  6 13:33:41 UTC 2017 - mfilka@suse.com
+
+- fate#322328
+  - adapted all-in-one dialog according to comments from UX
+- bnc#1022784
+  - Downloading release notes: Do not try again if the host or the
+    proxy cannot be resolved (schubi@suse.de)
```
mvidner: No, this is already mentioned at 3.2.19 below. The fact that it doesn't show up in the diff should have warned you.

package/yast2-installation.changes
```diff
@@ -1,4 +1,14 @@
...
+- 3.1.25
```
mvidner: 3.2.25

package/yast2-installation.changes
mchf: ah sorry

mchf: local diff when merging confused me a bit ;-)

mchf:
thanks for the review

PR #534, opened 2017-03-08
Add system roles through addons

imobachgs:
This PR adds support to allow add-ons to add new roles and requires https://github.com/yast/yast-yast2/pull/542.

Apart from that, I adapted the `SelectSystemRole` dialog to use the `SystemRole` class written by @teclator.

### DIFF
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -1778,27 +1778,6 @@ modules specified in the add-on product control file.
     </clone_modules>
 ```
 
-### Adding a system role
-
-This example shows how to add an additional `example_role`.
-
-```xml
-    <system_roles>
-      <insert_system_roles config:type="list">
-        <insert_system_roles>
-          <system_roles config:type="list">
-            <system_role>
-              <id>example_role</id>
-            </system_role>
-          </system_roles>
-        </insert_system_roles>
-      </insert_system_roles>
-    </system_roles>
-```
-
-NOTE: do not forget to add labels and descriptions to the `texts` section. Check
-*System Roles* section for more details.
-
 ### Example of OES 1.0
 
 The network code is instructed to force a static IP address.
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -1508,13 +1508,13 @@ section, which defines changes to the existing workflow and proposals.
             </inst_finish>
             <system_roles>
               <insert_system_roles config:type="list">
-                <insert_system_roles>
+                <insert_system_role>
                   <system_roles config:type="list">
                     <system_role>
                       <id>additional_role</id>
                     </system_role>
                   </system_roles>
-                </insert_system_roles>
+                </insert_system_role>
               </insert_system_roles>
             </system_roles>
         </update>
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -1303,6 +1303,12 @@ the installation workflow. Basically, adding proposal has two steps:
 Is possible as replacing or removing a step of the installation
 workflow.
 
+#### Adding new system roles
+
+Add-ons are allowed to define additional system roles. Those roles will be shown
+at the bottom of the list of already existing roles. For the time being,
+modifying or removing system roles is not possible.
+
 ### File layout
 
 #### Add-on Product CD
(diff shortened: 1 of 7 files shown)


src/lib/installation/select_system_role.rb
```diff
@@ -46,7 +47,7 @@ def initialize
     end
 
     def run
-      if raw_roles.empty?
+      if roles(true).empty?
```
mvidner: A boolean argument is quite obscure here. I would use a symbol argument `roles(:including_hamlet)` or another method, `roles_including_hamlet`.

doc/control-file.md
```diff
@@ -1500,6 +1506,17 @@ section, which defines changes to the existing workflow and proposals.
                     <module>before_umount_2</module>
                 </before_umount>
             </inst_finish>
+            <system_roles>
+              <insert_system_roles config:type="list">
+                <insert_system_roles>
```
mvidner: I don't think it is possible to name the item exactly the same as the collection.

doc/control-file.md
```diff
@@ -1500,6 +1506,17 @@ section, which defines changes to the existing workflow and proposals.
...
+                  <system_roles config:type="list">
```
mvidner: Why are we nesting a `system_roles` list within an `insert_system_roles` list?

doc/control-file.md
```diff
@@ -1500,6 +1506,17 @@ section, which defines changes to the existing workflow and proposals.
                     <module>before_umount_2</module>
                 </before_umount>
             </inst_finish>
+            <system_roles>
```
mvidner: This element name collides with the existing `system_roles` collection below.

doc/control-file.md
```diff
@@ -1761,6 +1778,27 @@ modules specified in the add-on product control file.
     </clone_modules>
 ```
 
+### Adding a system role
+
+This example shows how to add an additional `example_role`.
```
mvidner: I don't see what value this section is adding compared to the description added above. In particular, it omits any context, like which element should be the parent.
The problem probably lies with the existing structure of this document where a table of contents is missing and I hardly know the section nesting. A weak point of (Github flavored) Markdown?

doc/control-file.md
```diff
@@ -1500,6 +1506,17 @@ section, which defines changes to the existing workflow and proposals.
...
+              <insert_system_roles config:type="list">
+                <insert_system_roles>
+                  <system_roles config:type="list">
+                    <system_role>
+                      <id>additional_role</id>
+                    </system_role>
+                  </system_roles>
+                </insert_system_roles>
+              </insert_system_roles>
+            </system_roles>
```
mvidner: The inner `system_roles` container is an existing pattern so IMHO it should stay, and wrapping it in an `insert_system_roles` element is OK, but I would drop the outer 2 elements.

doc/control-file.md
imobachgs: When parsing the file, it will work as the name of the collection elements are not important. It could be `item_list`, `insert_system_roles` or `whatever`. I'm not sure if it could be a problem when validating.

I was thinking about using the singular form `insert_system_role` but, as you can be adding multiple roles, I decided to go for the plural one. But I'm open to suggestions.

doc/control-file.md
imobachgs: Because I think that we'll want to support scenarios like this:

```xml
<insert_system_roles config:type="list">
  <insert_system_roles>
    <before>normal_role</before>
    <system_roles config:type="list">
      <system_role><id>additional_role</id></system_role>
    </system_roles>
  </insert_system_roles>
</insert_system_roles>
```

It's similar to what we have when inserting modules:

```xml
<insert_modules config:type="list">
  <insert_module>
    <before>perform</before>
      <modules config:type="list">
        <module>
          <label>Perform Installation</label>
          <name>i1_netprobe</name>
        </module>
     </modules>
   </insert_module>
</insert_modules>
```
```

doc/control-file.md
imobachgs: See previous comment. Sure, we could simplify but I would like to stick with a more extensible syntax.

doc/control-file.md
imobachgs: Yes, you're right. I'll get rid of it.

doc/control-file.md
mvidner: I see. Being consistent with this existing design is a convincing argument.

With consistency in mind, please change the nested `insert_system_roles` item to singular.

Please make a PR to update the schema: https://github.com/yast/yast-installation-control/blob/82ec69490046295bc130ba88f7ea320718a0f363/control/control.rnc#L811

imobachgs:
I've created a PR to update the schema: https://github.com/yast/yast-installation-control/pull/40

PR #537, opened 2017-03-15
Redistribute roles buttons space

imobachgs:
This PR improves how role buttons are distributed in textmode. The point is that adding a role didn't fit in the 80x25 mode. Now separation is adjusted and, if roles don't fit even without separation, the description and the introductory text will be omitted.

## All roles fit
selection-role-dialog-textmode-fit

## All roles still fit
selection-role-dialog-textmode-still-fit

## Not enough space
selection-role-dialog-textmode-out-of-space

If roles don't fit even without the description, the [introductory text will be also omitted](https://github.com/yast/yast-installation/pull/537/files#diff-b9cac9df8ec348a3c81180e449f766e8R234).

## More than 80x25
selection-role-dialog-textmode-plenty-of-space

## Graphic mode
selection-role-dialog-graphic-mode

### DIFF
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -75,7 +75,6 @@ module Installation
       HSquash(
         VBox(
           Left(Label(Yast::ProductControl.GetTranslatedText("roles_text"))),
-          VSpacing(2),
           role_buttons
         )
       )
@@ -169,20 +168,13 @@ module Installation
       Yast::ProductFeatures.ClearOverlay
     end
 
+    # Returns the content for the role buttons
+    #
+    # @return [Yast::Term] Role buttons
+    #
+    # @see role_buttons_options
     def role_buttons
-      ui_roles = roles.each_with_object(VBox()) do |role, vbox|
-        # FIXME: following workaround can be removed as soon as bsc#997402 is fixed:
-        # bsc#995082: System role descriptions use a character that is missing in console font
-        description = Yast::UI.TextMode ? role.description.tr("•", "*") : role.description
-        vbox << Left(RadioButton(Id(role.id), role.label))
-        vbox << HBox(
-          HSpacing(Yast::UI.TextMode ? 4 : 2),
-          Left(Label(description))
-        )
-        vbox << VSpacing(2)
-      end
-
-      RadioButtonGroup(Id(:roles), ui_roles)
+      role_buttons_content(role_buttons_options)
     end
 
     # Applies given role to configuration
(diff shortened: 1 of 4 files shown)


jreidinger:
In general I suggest in the last case to remove also description after header what is system roles.

src/lib/installation/select_system_role.rb
```diff
@@ -215,5 +207,62 @@ def roles(refresh: false)
...
+      RadioButtonGroup(Id(:roles), VBox(VSpacing(margin), *ui_roles))
+    end
+
+    # Space for roles
+    #
+    # @return [Hash] Options to distribute roles
+    #
+    # @see available_lines_for_roles
+    def role_buttons_options
+      return { indentation: 4, margin: 2 } unless Yast::UI.TextMode
```
jreidinger: looks like constant for me

imobachgs:
@jreidinger I've made an additional improving and, in the last case, separation is also removed.

jreidinger:
@imobachgs I see but I suggest to also remove text "System roles are predefined ..." it is two lines which we can save.

src/lib/installation/select_system_role.rb
```diff
@@ -215,5 +207,62 @@ def roles(refresh: false)
...
+      default_opts = { indentation: 2, margin: 1 }
+      separation = 2.downto(0).find do |l|
+        needed_lines_for_roles + (roles.size * l) + default_opts[:margin] <= available_lines_for_roles
```
jreidinger: too long line. Maybe precompute `needed_lines_for_roles + default_opts[:margin]` ?

src/lib/installation/select_system_role.rb
```diff
@@ -215,5 +207,62 @@ def roles(refresh: false)
...
+      end
+      opts = separation.nil? ? { description: false } : { separation: separation }
```
jreidinger: so separation is 2 if even 0 do not fit?

src/lib/installation/select_system_role.rb
```diff
@@ -215,5 +207,62 @@ def roles(refresh: false)
...
+      merged_opts
+    end
+
+    # Number of required lines to show roles information
+    #
+    # Title + Descriptions
+    #
+    # @return [Integer] Number of lines needed to display roles information
+    def needed_lines_for_roles
+      @needed_lines_for_roles ||= roles.size + roles.map(&:description).reduce(0) do |sum, description|
```
jreidinger: what about `@needed_lines_for_roles ||= roles.size + roles.map{ |r| r.description.lines }.reduce(:+)` ?

src/lib/installation/select_system_role.rb
```diff
@@ -215,5 +207,62 @@ def roles(refresh: false)
...
+        sum + description.split("\n").size
+      end
+    end
+
+    # Space taken by header, description and footer and not available for the roles buttons
+    RESERVED_LINES = 9
```
jreidinger: I worry that description size is depending on content in control file and it can be also translated, so it can do wrong results.

src/lib/installation/select_system_role.rb
```diff
@@ -215,5 +207,62 @@ def roles(refresh: false)
...
+    # Returns an estimation of the available space for displaying the roles buttons
+    #
+    # @return [Integer] Estimated amount of available space
+    def available_lines_for_roles
+      @available_lines_for_roles ||= Yast::UI.GetDisplayInfo().fetch("Height") - RESERVED_LINES
```
jreidinger: why using fetch here?

jreidinger:
what I miss in general is some test cases. It is quite complex computing, so I will more trust into it if it is backed up with some test cases.

okurz:
does this have any effect on how role buttons are distributed in the GUI mode? Because the new desktop selection screen also in graphics mode looks like mostly whitespace and some cramped together text in the middle of the screen.

imobachgs:
@jreidinger I've made several improvements and I've updated the screenshots.

@okurz AFAIK, that's a [a different dialog](https://github.com/yast/yast-installation/blob/252bad28dbfc5bfd4954858bbb862438c0a3a21a/src/lib/installation/dialogs/desktop_roles.rb) so it should not be affected.

src/lib/installation/select_system_role.rb
```diff
@@ -215,5 +202,113 @@ def roles(refresh: false)
...
+      minimal_space_opts = { description: false, separation: 0 }
+      required_lines = roles.size + intro_text.lines.size
+      opts =
+        if required_lines + margin <= available_lines_for_roles
+          { margin: margin, show_intro: true }
+        elsif required_lines <= available_lines_for_roles
+          { margin: 0, show_intro: true }
+        else
+          { margin: 0, show_intro: false }
+        end
```
jreidinger: to be honest I found this a bit too much separated, in fact I see there cases:

1. `required_lines + margin >= available_lines_for_roles` then just count separation and show rest
2. `required_lines + margin < available_lines_for_roles` ( equal is wrong here, it cannot happen ) is no separation and no description
3. `required_lines <= available_lines_for_roles` nullify margin
4. rest do not show even intro.

I think these four cases should be together. now it looks like first cases is a bit special, which I do not see.
and btw if you have it, then first case do not need to be iteration, but simple int division with 
`separation = min(2, (available_lines_for_roles - required_lines)/roles.size)`

also I think that merge is not needed. In general you can set default and then in each case modify it.

src/lib/installation/select_system_role.rb
```diff
@@ -215,5 +202,113 @@ def roles(refresh: false)
...
+    # Number of required lines to show roles information and buttons
+    #
+    # Title + Descriptions
+    #
+    # @return [Integer] Number of lines needed to display roles information
+    def needed_lines_for_roles
+      return @needed_lines_for_roles if @needed_lines_for_roles
+      texts = roles.map(&:description)
+      texts << intro_text unless intro_text.nil?
+      lines = texts.compact.map(&:lines).reduce(:+).size
```
jreidinger: why there is a last size??? I do not get it

test/select_system_role_test.rb
```diff
@@ -159,4 +159,98 @@
...
+        expect(subject).to receive(:VSpacing).with(2) # margin
+        expect(subject).to receive(:RadioButton).with(anything, system_roles[0].label) # rol label
+        expect(subject).to receive(:Label).with(/#{system_roles[0].description}/) # rol description
+        expect(subject).to receive(:VSpacing).with(2) # separator
+        expect(subject).to receive(:RadioButton).with(anything, system_roles[1].label) # rol label
+        expect(subject).to receive(:Label).with(/#{system_roles[1].description}/) # rol description
+        subject.dialog_content
+      end
+    end
+  end
```
jreidinger: :+1: for tests

src/lib/installation/select_system_role.rb
imobachgs: Because at that point you get an array of lines and you're interested in the number of lines.

Anyway, I've moved the `size` to and renamed the var (lines quantity) to make it more clear.

PR #538, opened 2017-03-16
Allow proceeding after a blocking condition has been fixed

mvidner:
[bsc#1029291](https://bugzilla.suse.com/show_bug.cgi?id=1029291)

### DIFF
--- a/test/widgets_overview_test.rb
+++ b/test/widgets_overview_test.rb
@@ -1,0 +1,104 @@
+#! /usr/bin/env rspec
+
+require_relative "test_helper"
+
+require "installation/widgets/overview"
+
+# TODO: move these shared examples somewhere to be available to everyone
+RSpec.shared_examples "CWM::AbstractWidget" do
+  describe "#label" do
+    it "produces a String" do
+      expect(subject.label).to be_a String
+    end
+  end
+
+  describe "#handle" do
+    it "produces a Symbol or nil" do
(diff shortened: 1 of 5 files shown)


mvidner:
Tested manually on a CaaSP installation

test/widgets_overview_test.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+  end
+
+  describe "#handle" do
+    it "produces a Symbol or nil" do
+      expect(subject.handle(:dummy_event)).to be_a(Symbol).or be_nil
+    end
+  end
+
+  describe "#validate" do
```
jreidinger: in fact this is wrong, validate does not have to be defined at all. If it is not defined, then widget do not have validation.

test/widgets_overview_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+# TODO: move these shared examples somewhere to be available to everyone
+RSpec.shared_examples "CWM::AbstractWidget" do
+  describe "#label" do
+    it "produces a String" do
+      expect(subject.label).to be_a String
+    end
+  end
+
+  describe "#handle" do
```
jreidinger: same here, handle do not have to be defined

jreidinger: in context of abstract widget, here it is of course defined.

PR #539, opened 2017-03-17
taking /usr/bin/findmnt

schubi2:


src/lib/transfer/file_from_url.rb
```diff
@@ -489,13 +464,10 @@ def get_file_from_url(scheme:, host:, urlpath:, localfile:,
...
-            # this is workaround for bnc#849767
-            # because of changes in autoyast startup this code is now
-            # called much sooner (before Storage stuff is initialized)
-            # call dummy method to trigger Storage initialization
-            dummy = Storage.GetUsedFs()
-            mp = Storage.DeviceMounted(Ops.add("/dev/", _Host2))
-            already_mounted = !Builtins.isempty(mp)
+            # checking if device has already been mounted. Taking new mountpoint
+            # if it has already been done.
+            mp = `/usr/bin/findmnt --first-only --output=target /dev/#{_Host2}`.split.last
```
lslezak: Use `/dev/#{Shellwords.escape(_Host2)}` to avoid problems with unusual file names (spaces) and avoid shell injection vulnerability. (`require "shellwords"` is needed see https://docs.ruby-lang.org/en/2.2.0/Shellwords.html.)

src/lib/transfer/file_from_url.rb
```diff
@@ -402,32 +401,8 @@ def get_file_from_url(scheme:, host:, urlpath:, localfile:,
...
-              "file  %1 can't be retrieved",
-              Ops.add(Ops.add(mount_point, "/"), _Path)
-            )
-          else
-            @GET_error = ""
-            ok = true
-          end
-          SCR.Execute(path(".target.umount"), mount_point)
-        end
+        @GET_error = _("Floppy is not supported anymore.")
```
ancorgs: NP: Maybe, to make this more understandable for users and translators the message should be a little more descriptive. I don't know, something like "URLs starting with 'floppy:/' are not longer supported".

src/lib/transfer/file_from_url.rb
```diff
@@ -489,13 +464,10 @@ def get_file_from_url(scheme:, host:, urlpath:, localfile:,
...
-            # because of changes in autoyast startup this code is now
-            # called much sooner (before Storage stuff is initialized)
-            # call dummy method to trigger Storage initialization
-            dummy = Storage.GetUsedFs()
-            mp = Storage.DeviceMounted(Ops.add("/dev/", _Host2))
-            already_mounted = !Builtins.isempty(mp)
+            # checking if device has already been mounted. Taking new mountpoint
+            # if it has already been done.
+            mp = `/usr/bin/findmnt --first-only --output=target /dev/#{_Host2}`.split.last
+            already_mounted = !mp.nil?
```
lslezak: Um, for not mounted or not existing devices it returns empty string, not nil:

```
irb(main):001:0> `/usr/bin/findmnt --first-only --output=target /dev/foo`
=> ""
```

lslezak: Moreover the output contains a header and a trailing new line:

```
irb(main):002:0> `/usr/bin/findmnt --first-only --output=target /dev/sdb1`
=> "TARGET\n/\n"
```

So you need to use the `--noheadings` option and call `strip` on the result:

```
irb(main):002:0> `/usr/bin/findmnt --noheadings --first-only --output=target /dev/sdb1`.strip
=> "/"
```

lslezak: Ah, there is `split.last` at the end... This actually removes the header. So it is OK.

Sorry for the confusion.

PR #540, opened 2017-03-20
Allow proceeding after a blocking condition has been fixed (master)

mvidner:
forward port of #538 
[bsc#1029291](https://bugzilla.suse.com/show_bug.cgi?id=1029291)

### DIFF
--- a/test/widgets_overview_test.rb
+++ b/test/widgets_overview_test.rb
@@ -1,0 +1,128 @@
+#! /usr/bin/env rspec
+
+require_relative "test_helper"
+
+require "installation/widgets/overview"
+
+# TODO: these shared examples will soon be available in yast2.rpm
+# and then we can just `require "cwm/rspec"`
+RSpec.shared_examples "CWM::AbstractWidget" do
+  context "these methods are only tested if they exist" do
+    describe "#label" do
+      it "produces a String" do
+        next unless subject.respond_to?(:label)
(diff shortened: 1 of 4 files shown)


test/widgets_overview_test.rb
```diff
@@ -0,0 +1,128 @@
...
+    before do
+      allow(Yast::WFM).to receive(:CallFunction)
+        .with("adventure", ["MakeProposal", { "simple_mode" => true }])
+        .and_return(proposal_oops, proposal)
+    end
+
+    describe "#validate" do
+      it "first returns false, then returns true" do
+        subject.contents
+        expect(subject.validate).to be false
```
lslezak: NP: Um, what about using the `change` expectation here and changing the description to "user correction makes the invalid proposal valid"?

Something like
```ruby
expect {subject.handle(:hire_mr_baggins)}.to change {subject.validate}.from(false).to(true)
```

See the [doc](https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher#expect-change).

mvidner: Thanks, this is new to me!

It is available since RSpec 3.0, and CaaSP has 3.3 so we could use it even there.
https://www.relishapp.com/rspec/rspec-expectations/v/3-0/docs/built-in-matchers/change-matcher

PR #542, opened 2017-03-23
Sync storage-ng with master

ancorgs:


package/yast2-installation.changes
```diff
@@ -1,10 +1,345 @@
...
+Tue Mar 21 11:59:26 UTC 2017 - ancor@suse.com
+
+- Added some extra space to the openSUSE desktop/roles selection
+  dialog (improvement on the fix for poo#14936, bsc#1025415).
+
+-------------------------------------------------------------------
 Fri Mar 17 16:59:04 CET 2017 - schubi@suse.de
 
 - storage-ng: Taking /usr/bin/findmnt in order to evaluate mount
   points. Floppy support removed.
```
jreidinger: well, I am, not sure if this is good approach, what happen when it is merged to master? changelog will look very differently. but I will let it on you, just expect some troubles after merge.

ancorgs: Is not. We should have synced with master BEFORE introducing the findmnt change. As a rule we always sync before introducing storage-ng changes, but we forgot. :-(

PR #543, opened 2017-03-23
Add dashboard role handler

imobachgs:
Add a system role handler for the dashboard role. Moreover, does some refactoring to the system roles handlers mechanism.

See https://trello.com/c/PJqM8x0T for details.


imobachgs:
Thanks for the review! :)

src/lib/installation/system_role_handlers_runner.rb
```diff
@@ -0,0 +1,61 @@
...
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, contact SUSE.
+#
+# To contact SUSE about this file by physical or electronic mail, you may find
+# current contact information at www.suse.com.
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "installation/cfa/salt"
```
jreidinger: why it is needed here??? It is for worker role only, so it should be in that file.

imobachgs: You're right. I'll fix it (without version bumping) and anyway I'll remove it when merging into master.

kobliha:
@imobachgs Just a side note: This really looks like something not generic enough, something that probably belongs to a CaaSP-specific package. But it's just an opinion to consider...

imobachgs:
@kobliha Yes, it sounds good: we could move those roles handlers to a `yast-caasp` package. And we should think also about more code that could be included in that specific package.

PR #544, opened 2017-03-24
Move code to yast-caasp

imobachgs:
This code removes some code that will be moved to the new https://github.com/yast/yast-caasp repository. See https://github.com/yast/yast-caasp/pull/1.


src/lib/installation/system_role_handlers_runner.rb
```diff
@@ -51,7 +50,7 @@ def finish(role_id)
     #
     # @return [Boolean] True if the file was loaded; false otherwise.
     def require_handler(role_id)
-      filename = "installation/system_role_handlers/#{role_id}_finish"
+      filename = "y2system_role_handlers/#{role_id}_finish"
       require filename
       true
     rescue LoadError
```
jreidinger: in fact this false branch do not return false, but nil

doc/system-role-handlers.md
```diff
@@ -0,0 +1,40 @@
...
+    def run
+      # Handler implementation goes here
+    end
+  end
+end
+
+```
+
+The `run` method won't receive any argument and is also not expected to return
+any special value.
```
lslezak: And what about the exceptions? Are they caught (and ignored) or do they buble up and trigger the general YaST exception handler?

lslezak: Just document the behavior...

imobachgs: They just bubble up and get captured by YaST.

jreidinger:
@imobachgs lgtm from my side

imobachgs:
Thanks!

PR #545, opened 2017-03-27
Use the shared package download and extraction from FATE#320772

lslezak:
Use the shared functionality also in the self-update code.

- 3.2.33

### DIFF
--- a/test/update_repository_test.rb
+++ b/test/update_repository_test.rb
@@ -110,8 +110,8 @@ describe Installation::UpdateRepository do
     let(:libzypp_package_path) { "/var/adm/tmp/pkg1-3.1.x86_64.rpm" }
     let(:package_path) { "/var/adm/tmp/pkg1-3.1.x86_64.rpm" }
     let(:tempfile) { double("tempfile", close: true, path: package_path, unlink: true) }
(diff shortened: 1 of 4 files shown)


test/update_repository_test.rb
```diff
@@ -121,14 +121,13 @@
       allow(Tempfile).to receive(:new).and_return(tempfile)
 
       # Download
-      expect(Yast::Pkg).to receive(:ProvidePackage)
-        .with(repo_id, package["name"], tempfile.path.to_s)
-        .and_return(true)
+      expect_any_instance_of(Packages::PackageDownloader).to receive(:download)
```
imobachgs: I'm not a big fan of `{expect,allow}_any_instance_of` as they're intended to work be used [when working with legacy code](https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance) (check the title). I would rather prefer:

* Using dependency injection or...
* ... just simple mock the `allow` on the class constructor to return a double you can verify.

Something like:

```ruby
allow(Packages::PackageExtractor).to receive(:new).and_return(your_double)
```

In this case, I guess that the second option is good enough a less disruptive.

lslezak:
Ugh, the improved test caught a bug in the code... :flushed:

PR #547, opened 2017-03-30
document new roles features

jreidinger:


doc/control-file.md
```diff
@@ -798,13 +798,18 @@ Section *supported\_desktops* contains list of one or more
...
+It also allows to modify configuration of Systemd Services.
 
 A role can also define additional dialogs that are shown when a given role is
 selected. It is a common installation dialog with *Abort*, *Cancel* and *Next*
 buttons. It supports and uses all parameters from the **GetInstArgs** module.
 When going back, it will first show the last additional dialog and when going
 back through all additional dialogs, it will show again the roles selection.
 
+System roles by default preselect first defined entry unless user set it.
+If the first role has an attribute *no_default* then no role will be
```
kobliha: no_default = true; I guess? It's boolean (?)

doc/control-file.md
```diff
@@ -829,6 +834,11 @@ Example:
           <default_patterns>base Minimal kvm_server</default_patterns>
         </software>
         <additional_dialogs>kvm_setup,virt_manager_setup </additional_dialogs>
+        <services config:type="list">
+          <service>
+            <name>salt-minion</name>
```
kobliha: Please use more services, to make it clear

kobliha:
I'm missing an info about "services" in general, such as that roles can also define services, that are enabled at the end of installation

doc/control-file.md
jreidinger: yes, `<no_default  config:type="boolean"> true </no_default>` is the way to specify it. same as other booleans

doc/control-file.md
```diff
@@ -829,6 +834,11 @@ Example:
           <default_patterns>base Minimal kvm_server</default_patterns>
         </software>
         <additional_dialogs>kvm_setup,virt_manager_setup </additional_dialogs>
+        <services config:type="list">
```
kobliha: (?)

doc/control-file.md
```diff
@@ -798,13 +798,18 @@ Section *supported\_desktops* contains list of one or more
 System Roles, if defined in the control file, are presented during
 the first stage of the installation. The user will select one of them
 and they will affect the proposed configuration of Partitioning and Software.
+It also allows to modify configuration of Systemd Services.
```
kobliha: Hmm, maybe `allows to enable systemd services`, because other configuration is not supported

jreidinger: here I want to be more generic, as maybe in future we extend it with some removal? So details are lower, but in general here it is about generic abilities.

kobliha: OK, that's possible. I usually don't offer something I can't deliver at the end :)

kobliha:
LGTM

PR #548, opened 2017-03-31
The installer extension documentation

lslezak:
## A New Documentation for the Installer Extension

*Note: review the [rendered document](https://github.com/yast/yast-installation/blob/installation_xml_doc/doc/installer_extension.md), it looks much better than a plain MD file...*

- The installer allows adding new steps
add_on_extension_new_step

- Or adding new roles
add_on_extension_new_role

# Blog

The YaST installer already allows extending the installation work flow by add-on extensions. However, this was only supported for the SUSE tag repositories (ISO installation media). That means that online repositories, which normally use RPM-MD data format, could not be used with feature.

This sprint we extended the support also for the other repository formats (basically any repository format supported by libzypp). The original limitation was caused by the fact that the other repository types do not support other files except the RPM packages.

To overcome this limitation we now support packaging the installer extension files into a standard RPM package which can be provided by any repository type.

The implementation is [documented](https://github.com/yast/yast-installation/blob/master/doc/installer_extension.md) and there is an example [YaST:extension](https://build.opensuse.org/project/show/YaST:extension) OBS project with an example extension.

add_on_extension

### DIFF
--- a/doc/installer_extension.md
+++ b/doc/installer_extension.md
@@ -1,0 +1,105 @@
+# Modifying the Installation by an Add-on
+
+The YaST installation workflow can be easily changed or extended by add-ons.
+
+
+## Configuring the Add-on Metadata
+
+The product package on the add-on medium (usually a `*-release` RPM package)
+should link to a package providing the installer extension by a
+`Provides: installerextension(<package_name>)` dependency.
+
+That referenced package should contain the `installation.xml` file and optionally
+the `y2update.tgz` archive.
+
+## Modifying the Installation Defaults
+
+### Adding a New System Role
+
+This `installation.xml` example adds a new role:
+
+```xml
+<?xml version="1.0"?>
+<!DOCTYPE productDefines SYSTEM "/usr/share/YaST2/control/control.rng">
+<productDefines xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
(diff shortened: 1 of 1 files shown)


doc/installer_extension.md
```diff
@@ -0,0 +1,177 @@
...
+## Configuring the Add-on Metadata
+
+The product package on the add-on medium (usually a `*-release` RPM package)
+should link to a package providing the installer extension by a
+`Provides: installerextension(<package_name>)` dependency.
+
+That referenced package should contain the `installation.xml` file and optionally
+the `y2update.tgz` archive in the root directory. The package should never
+be installed into system, it is used just to provide the temporary data for the
+installer. After the installation is finished the files are not need anymore.
```
imobachgs: needed

doc/installer_extension.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Modifying the Installation Workflow
+
+This `installation.xml` example adds a new step into installtion workflow and
```
imobachgs: typo: installation

doc/installer_extension.md
```diff
@@ -0,0 +1,177 @@
...
+also in add-on installation in an installed system:
```
imobachgs: It looks confusing to me. What about something like this? Just an idea:

```
The `installation.xml` example adds two new steps:

* one of them into the installation workflow
* the other one into the add-on installation for already installed systems
```

PR #549, opened 2017-04-04
Renamed Controller Node to Administration Node (bsc#1032057).

mvidner:
[bsc#1032057](https://bugzilla.suse.com/show_bug.cgi?id=1032057)


mvidner:
see also https://github.com/yast/skelcd-control-CASP/pull/32

PR #552, opened 2017-04-20
NTP server for dashboard role

imobachgs:
This PR implements support to specify NTP servers in the Allin1 dialog.

* Displays an input text to specify NTP servers.
* Fixes an usability issue when changing the role: if you click on the roles combo box, the value for the current setting (controller node or ntp servers) are lost, even if you close the combo box without changing the value.

About this PR: most of Allin1 dialog stuff should be moved to CaaSP. I plan to do it during this sprint, but I prefer to merge this code already so people can start working on integration tests.

Regarding the yast-slp dependency, it's already in the inst-sys (used for yast2-registration). That dependency should be also moved to yast-caasp.


src/Makefile.am
```diff
@@ -217,7 +217,8 @@ ylibdialog_DATA = \
 ylibwidgetdir = "${yast2dir}/lib/installation/widgets"
 ylibwidget_DATA = \
   lib/installation/widgets/overview.rb \
-  lib/installation/widgets/system_role.rb
+  lib/installation/widgets/system_role.rb \
+  lib/installation/widgets/ntp_server.rb
```
jreidinger: please use alphabetic sorting

src/lib/installation/clients/inst_casp_overview.rb
```diff
@@ -178,5 +183,22 @@ def content
     def separate_wizard_needed?
       Yast::Mode.normal
     end
+
+    # Regexp to extract the URL from a SLP URL
+    # For instance, match 1 for "service:ntp://ntp.suse.de:123,65535"
+    # will be "ntp://ntp.suse.de:123"
+    SERVICE_REGEXP = %r{\Aservice:(ntp://[^,]+)(,\d+)?}
```
jreidinger: why it matches second part, but not use it?

src/lib/installation/clients/inst_casp_overview.rb
```diff
@@ -178,5 +183,22 @@ def content
...
+    # SLP identifier for NTP
+    NTP_SLP_SERVICE = "ntp".freeze
+    # Discover NTP servers through SLP
+    #
+    # @return [Array<String>] NTP server names
+    def ntp_servers
+      Yast::SlpService.all(NTP_SLP_SERVICE).map do |service|
+        match = SERVICE_REGEXP.match(service.slp_url)
+        url = match[1]
```
jreidinger: what about `url = service.slp_url[SERVICE_REGEXP, 1]`

src/lib/installation/clients/inst_casp_overview.rb
```diff
@@ -178,5 +183,22 @@ def content
...
+        URI(url).host
```
jreidinger: can it happen that regexp do not return expected format? in such case it will raise exception here

src/lib/installation/widgets/ntp_server.rb
```diff
@@ -0,0 +1,163 @@
...
+Yast.import "CWM"
+Yast.import "Popup"
+Yast.import "Label"
+Yast.import "IP"
+Yast.import "Hostname"
+
+module Installation
+  module Widgets
+    # This widget is responsible of validating and storing the NTP server to use.
+    class NtpServer < CWM::InputField
```
jreidinger: I miss here `require "cwm/widget"` where is defined those base widgets.

src/lib/installation/widgets/ntp_server.rb
```diff
@@ -0,0 +1,163 @@
...
+      # @return [Array<String>] List of default servers
+      attr_reader :default_servers
+
+      # @return [String] Last known value (@see #remember!)
+      attr_accessor :last_value
+      private :last_value, :last_value=
```
jreidinger: you can simple write this accessor to private section of class.
```
jreidinger@linux-vvcf:~/prace/yast/bootloader> irb
irb(main):001:0> class A
irb(main):002:1> private
irb(main):003:1> attr_accessor :a
irb(main):004:1> end
=> nil
irb(main):005:0> A.new.a
NoMethodError: private method `a' called for #<A:0x00000001b9bf58>
	from (irb):5
	from /usr/bin/irb:11:in `<main>'
```

src/lib/installation/widgets/ntp_server.rb
```diff
@@ -0,0 +1,163 @@
...
+      #
+      # @return [Boolean] true if value is valid; false otherwise.
+      def validate
+        return skip_ntp_server? if servers.empty?
+        return true if servers.all? { |v| Yast::IP.Check(v) || Yast::Hostname.CheckFQ(v) }
+        Yast::Popup.Error(
+          # TRANSLATORS: error message for invalid administration node location
+          _("Not valid location for the NTP servers, " \
+            "please enter a valid IP or Hostname")
+        )
```
jreidinger: why translated here when label intentionally is not? and one suggestion. I really suggest to write down invalid servers.
so something like
```
invalid_servers = servers.reject { |v| Yast::IP.Check(v) || Yast::Hostname.CheckFQ(v) }
return true if invalid_servers.empty?
# use invalid servers in popup

File: src/lib/installation/widgets/ntp_server.rb [hunk 8]
```diff
@@ -0,0 +1,163 @@
...
+
+        false
+      end
+
+      # Remember the value when init is called
+      #
+      # @see #last_value
+      def remember!
+        self.last_value = value
+      end
```
jreidinger: do you test if this work? I think we use always new instances. So why it is needed?

src/lib/installation/widgets/ntp_server.rb
```diff
@@ -0,0 +1,163 @@
...
+      # @return [Array<String>] List of hostnames/addresses
+      def servers
+        value.tr(",", " ").split(" ")
+      end
+
+      # Check if the user wants to intentionally skip the NTP server configuration
+      #
+      # @return [Boolean] true if user wants to skip it; false otherwise.
+      def skip_ntp_server?
+        Yast::Popup.AnyQuestion(
```
jreidinger: again why translation when label is intentionally not translated?

src/lib/installation/widgets/ntp_server.rb
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    # NTP Server widget placeholder
+    class NtpServerPlace < CWM::ReplacePoint
+      # @return [NtpServer] NTP Server widget
+      attr_reader :ntp_server
+      # @return [Empty] Empty widget placeholder
+      attr_reader :empty
+
+      private :ntp_server, :empty
```
jreidinger: same here, can be in private section, so top of class is for public ones.

src/lib/installation/widgets/ntp_server.rb
```diff
@@ -0,0 +1,163 @@
...
+      end
+
+      # Show the NtpServer widget
+      def show
+        replace(ntp_server)
+      end
+
+      # Hide the NtpServer widget
+      def hide
+        ntp_server.remember!
```
jreidinger: ah, now I see why you want this method. In fact it can be done easier and holding it here. With
```
def hide
  @ntp_value = ntp_server.value
  replace(empty)
end

def show
  replace(ntp_server)
  ntp_server.value = @ntp_value if @ntp_value
end
```

src/lib/installation/widgets/system_role.rb
```diff
@@ -110,12 +139,10 @@ def init
...
-        if value == "worker_role"
-          @controller_node_widget.show
-        else
-          @controller_node_widget.hide
-        end
-
+        to_show = ROLE_WIDGETS.fetch(value, [])
+        to_hide = widgets_map.keys - to_show
+        to_hide.each { |w| widgets_map[w].hide }
+        to_show.each { |w| widgets_map[w].show }
```
jreidinger: :+1:

src/lib/installation/widgets/system_role.rb
```diff
@@ -34,6 +35,10 @@ module Widgets
     # which must be a valid IP or FQDN.
     # bsc#1032057: old name: Controller Node, new name: Administration Node.
     class ControllerNode < CWM::InputField
+      # @return [String] Last known value (@see #remember!)
+      attr_accessor :last_value
+      private :last_value, :last_value=
```
jreidinger: looks like your favorite pattern :) So let me explain why I do not like it. When I read code, it is like "on top is most important attribute, it is last value...and then I read that it is not important at all as it is just private one :)

src/lib/installation/widgets/ntp_server.rb
```diff
@@ -66,7 +66,7 @@ def init
         self.value = saved_servers.join(" ")
       end
 
-      NOT_VALID_SERVERS_MESSAGE = "Not valid location for the NTP servers:\n%<servers>" \
+      NOT_VALID_SERVERS_MESSAGE = "Not valid location for the NTP servers:\n%<servers>s" \
```
jreidinger: I do not get this change, why `s` is needed here?

src/Makefile.am
```diff
@@ -217,8 +217,8 @@ ylibdialog_DATA = \
 ylibwidgetdir = "${yast2dir}/lib/installation/widgets"
 ylibwidget_DATA = \
   lib/installation/widgets/overview.rb \
-  lib/installation/widgets/system_role.rb \
-  lib/installation/widgets/ntp_server.rb
+  lib/installation/widgets/ntp_server.rb \
```
jreidinger: sorry, but `n` is before `o`

src/lib/installation/widgets/ntp_server.rb
```diff
@@ -0,0 +1,115 @@
...
+        role["ntp_servers"] = servers
+      end
+
+      # Initializes the widget's value
+      def init
+        saved_servers = role["ntp_servers"] || default_servers
+        self.value = saved_servers.join(" ")
+      end
+
+      NOT_VALID_SERVERS_MESSAGE = "Not valid location for the NTP servers:\n%<servers>s" \
```
jreidinger: still I do not get why s is needed here

imobachgs: AFAIK, it won't work:

```
string = "hello %<name>s"
format(string, name: "world") # hello world
string = "hello %<name>"
format(string, name: "world") # hello %
```

jreidinger: for string I suggest to use `format("hello %{name}", name: "world")` as it is more convenient when you do not need conversion of input

PR #554, opened 2017-04-24
 Merge Sle 12 sp1 into Sle 12 sp2

gilsonsouza:


teclator:
Just for record, it basically merge fix added in #553 into SP2

PR #555, opened 2017-04-24
Merging SLE 12 SP2 into master

gilsonsouza:


teclator:
From my POV it should be merged in https://github.com/yast/yast-installation/tree/SLE-12-SP2-CASP and then from it into master.

@imobachgs could you check.

.travis.yml
```diff
@@ -4,7 +4,7 @@ services:
   - docker
 
 before_install:
-  - docker build -t yast-installation-image .
+    - docker build -t yast-installation-image .
```
imobachgs: Is this change intentional?

PR #562, opened 2017-05-08
Don't crash in case of an invalid linuxrc registration url is provided (bsc#1035908)

teclator:
An exception will be raised now when a wrong registration url is provided and an error will be reported:

skippingupdates

This is in case of not asking about a new one.

### DIFF
--- a/src/lib/installation/update_repositories_finder.rb
+++ b/src/lib/installation/update_repositories_finder.rb
@@ -124,7 +124,13 @@ module Installation
     #
     # @return [Array<URI>,false] self-update URLs or false in case of error
     def update_urls_from_connect
-      url = registration_url
+      begin
+        url = registration_url
+      rescue URI::InvalidURIError => e
+        log.error("Custom registration url is wrong, URI failed with: #{e.message}")
+        return []
+      end
+
       return [] if url == :cancel
 
       custom_regserver = url != :scc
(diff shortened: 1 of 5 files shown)


src/lib/installation/update_repositories_finder.rb
```diff
@@ -124,7 +124,13 @@ def update_url_from_control
     #
     # @return [Array<URI>,false] self-update URLs or false in case of error
     def update_urls_from_connect
-      url = registration_url
+      begin
+        url = registration_url
+      rescue URI::InvalidURIError => e
+        log.error("Custom registration url is wrong, URI failed with: #{e.message}")
```
lslezak: Why not using `Report.Error` as in https://github.com/yast/yast-registration/pull/319?

teclator: In case of self_update we fail silently

lslezak: Yes, but silent fail should happen only in the default case. IMHO if the user uses the boot command line then an error should be printed.

imobachgs: After discussing with @teclator I think that @lslezak is right and we should report the error. But instead of using `Report` there, I would raise an exception (maybe a custom one) which could be handled in the client to not pollute this class with user interaction stuff.

lslezak: Good idea, an exception is better at this place.

teclator:
@lslezak and also @imobachgs and I prefer the #563

teclator:
@lslezak also I prefer the #563 one,

PR #563, opened 2017-05-09
Early check of wrong urls allowing to fix them (bsc#1035908)

teclator:
The idea is to reuse the network 'inst_install_inf' not just writing the hostname and proxy configuration but also checking wrong urls provided in the command line.

The new `::Installation::URLDialog` should be moved to yast-yast2 using the `UI` namespace instead of `Installation`

The current approach allows the user to fix the url, or nullify it in case of not fixed (abort).

screenshot_sle12sp3_2017-05-10_09 52 09

In case of `AutoYast`, after discussed with @imobachgs seems that we should continue just reporting it to not block the installation..

skippingupdates

In the registration step it will handle it just as a connection error, reporting it so no change is needed there.

screenshot_sle12sp3_2017-05-08_14 50 16

### DIFF
--- a/src/lib/installation/clients/inst_install_inf.rb
+++ b/src/lib/installation/clients/inst_install_inf.rb
@@ -22,10 +22,7 @@ module Yast
 
       regurl = Linuxrc.InstallInf("regurl")
 
-      if need_fix?(regurl)
-        fix_regurl!(regurl)
-        Linuxrc.ResetInstallInf
-      end
+      fix_regurl!(regurl) if need_fix?(regurl)
 
       Yast::Wizard.CloseDialog if separate_wizard_needed?
 
@@ -34,12 +31,21 @@ module Yast
 
   private
 
+    # Checks if the given url is invalid and needs to be fixed.
+    #
+    # @params url [String]
+    # @return [Boolean] return true if not nil and invalid.
     def need_fix?(url)
       url && !valid_url?(url)
     end
 
+    # Shows a dialog allowing the user to modify the invalid URL, modifying the
+    # /etc/install.inf file with the new value. In case of cancelled or empty,
+    # the URL will be completely removed.
+    #
+    # @param regurl [String]
     def fix_regurl!(regurl)
-      while regurl && !valid_url?(regurl)
+      while need_fix?(regurl)
         new_url = ::Installation::RegistrationURLDialog.new(regurl).run
         case new_url
         when :cancel
@@ -55,7 +61,9 @@ module Yast
         end
       end
 
-      regurl ? replace_install_inf("regurl", regurl) : delete_install_inf("regurl")
+      SCR.Write(path(".etc.install_inf.regurl"), regurl)
+      SCR.Write(path(".etc.install_inf"), nil) # Flush the cache
+      Linuxrc.ResetInstallInf
     end
 
     # Check if the given URI is valid or not
@@ -71,24 +79,6 @@ module Yast
     # Returns whether we need/ed to create new UI Wizard
     def separate_wizard_needed?
       Yast::Mode.normal
-    end
-
-    def replace_install_inf(key, value)
-      # Updating new URI in /etc/install.inf (bnc#963487)
-      # SCR.Write does not work in inst-sys here.
-      WFM.Execute(
-        path(".local.bash"),
-        "sed -i \'/#{key}:/c\\#{key}: #{value}\' /etc/install.inf"
-      )
-    end
-
-    def delete_install_inf(key)
-      # Deleting URI in /etc/install.inf (bnc#963487)
-      # SCR.Write does not work in inst-sys here.
-      WFM.Execute(
-        path(".local.bash"),
-        "sed -i \'/#{key}:/d\' /etc/install.inf"
-      )
     end
   end
 end
--- a/src/lib/installation/clients/inst_install_inf.rb
+++ b/src/lib/installation/clients/inst_install_inf.rb
@@ -1,0 +1,90 @@
+require "yast"
+require "network/install_inf_convertor"
+require "installation/dialogs/registration_url_dialog"
+
+module Yast
+  class InstInstallInfClient < Client
+    VALID_URL_SCHEMES = ["http", "https"].freeze
+
+    include Yast::Logger
+    include Yast::I18n
+
+    Yast.import "Linuxrc"
+    Yast.import "Wizard"
+    Yast.import "WFM"
+
+    def main
+      textdomain "installation"
+
+      InstallInfConvertor.instance.write_netconfig unless Mode.auto
+
+      Yast::Wizard.CreateDialog if separate_wizard_needed?
+
+      regurl = Linuxrc.InstallInf("regurl")
+
+      fix_regurl(regurl) if need_fix?(regurl)
+
+      Yast::Wizard.CloseDialog if separate_wizard_needed?
+
+      :next
+    end
+
+  private
+
+    def need_fix?(url)
+      url && !valid_url?(url)
+    end
+
+    def fix_regurl(regurl)
+      while regurl && !valid_url?(regurl)
+        new_url = ::Installation::RegistrationURLDialog.new(regurl).run
+        case new_url
+        when :abort
+          if Popup.YesNo(_("If you decide to abort, the custom URL\n" \
+                         "will be completely removed.\n\n Really abort?\n"))
+            regurl = nil
+          end
+        when "",
+          regurl = nil
+        else
+          regurl = new_url
+        end
+      end
+
+      regurl ? replace_install_inf("regurl", regurl) : delete_install_inf("regurl")
+    end
+
+    # Check if the given URI is valid or not
+    #
+    # @params url [String]
+    # @return [Boolean]
+    def valid_url?(url)
+      VALID_URL_SCHEMES.include?(URI(url).scheme)
+    rescue URI::InvalidURIError
+      false
+    end
+
+    # Returns whether we need/ed to create new UI Wizard
+    def separate_wizard_needed?
+      Yast::Mode.normal
+    end
+
+    def replace_install_inf(key, value)
+      # Updating new URI in /etc/install.inf (bnc#963487)
+      # SCR.Write does not work in inst-sys here.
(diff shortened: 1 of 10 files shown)


src/lib/installation/clients/inst_install_inf.rb
```diff
@@ -0,0 +1,90 @@
...
+    rescue URI::InvalidURIError
+      false
+    end
+
+    # Returns whether we need/ed to create new UI Wizard
+    def separate_wizard_needed?
+      Yast::Mode.normal
+    end
+
+    def replace_install_inf(key, value)
```
imobachgs: As a PoC is ok, but for a final version maybe you would like to extend/use the [Linuxrc module](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Linuxrc.rb).

src/lib/installation/clients/inst_install_inf.rb
```diff
@@ -0,0 +1,90 @@
...
+      url && !valid_url?(url)
+    end
+
+    def fix_regurl(regurl)
+      while regurl && !valid_url?(regurl)
+        new_url = ::Installation::RegistrationURLDialog.new(regurl).run
+        case new_url
+        when :abort
+          if Popup.YesNo(_("If you decide to abort, the custom URL\n" \
+                         "will be completely removed.\n\n Really abort?\n"))
```
lslezak: `removed` sounds strange, `ignored` or `skipped` is better. Does it fallback to the default URL in this case? If so then that should be mentioned in the text.

src/lib/installation/clients/inst_install_inf.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    # Returns whether we need/ed to create new UI Wizard
+    def separate_wizard_needed?
+      Yast::Mode.normal
+    end
+
+    def replace_install_inf(key, value)
+      # Updating new URI in /etc/install.inf (bnc#963487)
+      # SCR.Write does not work in inst-sys here.
+      WFM.Execute(
```
lslezak: I'd prefer pure a Ruby `File.read`/`File.write` approach. Running `sed` in `bash` through WFM looks weird.

src/lib/installation/clients/inst_install_inf.rb
```diff
@@ -0,0 +1,90 @@
...
+        path(".local.bash"),
+        "sed -i \'/#{key}:/c\\#{key}: #{value}\' /etc/install.inf"
+      )
+    end
+
+    def delete_install_inf(key)
+      # Deleting URI in /etc/install.inf (bnc#963487)
+      # SCR.Write does not work in inst-sys here.
+      WFM.Execute(
```
lslezak: The same here...

src/lib/installation/dialogs/registration_url_dialog.rb
```diff
@@ -0,0 +1,42 @@
...
+# current contact information at www.suse.com.
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "installation/dialogs/url_dialog"
+
+module Installation
+  class RegistrationURLDialog < ::Installation::URLDialog
+    def help_text
+      _("<p>\n" \
```
lslezak: Please add `TRANSLATORS:` comments so they know the context of the messages.

src/lib/installation/dialogs/url_dialog.rb
```diff
@@ -0,0 +1,96 @@
...
+      @url = url
+    end
+
+    # Shows a dialog when the given url is wrong
+    #
+    # @param [String] original Original value
+    # @return [String] new value
+    def dialog_content
+      HBox(
+        HWeight(30, RichText(help_text)),
```
lslezak: The layout is too complicated IMHO. I'd just put the text into the label above the text field.

That squashed box on the left does not look nice.

src/lib/installation/clients/inst_install_inf.rb
teclator: It removes the entry from the instal.inf file, so like not providing any url in the command line. I added the comment about the default one, but after the registration review, decided to not mention it at all as it is not handled by this code.

src/lib/installation/clients/inst_install_inf.rb
teclator: I just used it from:

https://github.com/yast/yast-autoinstallation/blob/47c24fb98e074f5b6432f3a4f8b9421362ee29cc/src/clients/inst_autoinit.rb#L157

src/lib/installation/clients/inst_install_inf.rb
teclator: Of course, it is based in the previous one, so if moved to a Read/Write approach then will be also updated, and as Imo suggested, make sense to just handle it in `Linuxrc`

src/lib/installation/dialogs/registration_url_dialog.rb
teclator: SURE, will add all of them.

src/lib/installation/dialogs/url_dialog.rb
teclator: The dialog is exactly the same used for ProfileLocation just to make it consistent:

https://github.com/yast/yast-autoinstallation/blob/6ed6c508685dbe5fa358d97d54940d3c1cdd361f/src/include/autoinstall/autoinst_dialogs.rb#L21

I like your suggestion, so for the generic one could be used. Will modify it according to your comments.

teclator: Something like this one?:

alternativeurl

lslezak: Yes, that's much better!

src/lib/installation/dialogs/url_dialog.rb
```diff
@@ -40,45 +40,52 @@ def initialize(url)
...
+
+    # Help text that will be displayed above the url entry
     #
     # @return [String]
     def help_text
       ""
     end
 
+    def show_help?
+      return true if help_text && help_text != ""
```
lslezak: `!help_text.empty?`

lslezak:
From the code POV it LGTM.

The question is which solution should be preferred. This one or #562 ?

- This solution is better from UX POV, user has a chance to fix a typo in the URL without need for reboot
- On the other hand this is a bit inconsistent, e.g. if you enter an invalid static IP in the `netsetup` boot option then it is ignored, no popup to fix that. That means some invalid options can be fixed some not.

(And I'd prefer this one.)

teclator:
@lslezak and I have detected other scenarios when the url's are read from the profile instead of the command line. So I will also ask for URL fix in that cases but as you commented there are other scenarios that are not checked at all.

At the end, it is just about let the user fix wrong entries in a friendly way but in many cases it will need to fix it and reboot, or reboot fixing the given command line options.

teclator:
@lslezak @imobachgs I have merged #562 into this PR as the modifications are not exclusive. 

In **a manual installation** we try to fix INVALID urls as early as possible

In **autoinstallation** we also rescue exceptions that came from the profile in AutoYast reporting them.

Could you review it?

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -176,6 +176,10 @@ def update_repositories
       @update_repositories = update_repositories_finder.updates
       log.info("self-update repositories are #{@update_repositories.inspect}")
       @update_repositories
+    rescue ::Registration::InvalidURL
```
lslezak: :warning: Be careful here, `::Registration::InvalidURL` is NOT defined when `require_registration_libraries` fails! E.g. openSUSE does not include registration, but supports self-update.

You need to check whether the `::Registration::InvalidURL` name is defined or check the `@require_registration_libraries` value. (But I do not know how to do it nice in a `rescue` statement, maybe rescue all and check for the type and re-raise other exceptions than `::Registration::InvalidURL`... ???)

lslezak: Or you can use `URI::InvalidURIError` exception from base Ruby.

teclator:
```
    rescue Exception => e
      raise e unless e.class == "Registration::InvalidURL"
``

Comment (imobachgs) [inline src/lib/installation/clients/inst_update_installer.rb]: I would avoid the `Exception` form. I'm fine with any of these options:

* Defining the `::Registration::InvalidURL` to inherit from `URI::InvalidURIError`...
* ... or, as @lslezak suggested, use `URI::InvalidURIError` directly.

Comment (teclator) [inline src/lib/installation/clients/inst_update_installer.rb]: @imobachgs OK, I get it

Comment (teclator) [inline src/lib/installation/clients/inst_update_installer.rb]: @lslezak Updated

File: src/lib/installation/dialogs/url_dialog.rb [hunk 9]
```diff
@@ -0,0 +1,142 @@
...
+      false
+    end
+
+    # Determines wether the dialog title is shown
+    #
+    # @return [Boolean] true if the dialog title is not empty
+    #
+    # @see dialog_title
+    def show_heading?
+      return true if !dialog_title.empty?
```
lslezak: This whole method body can be simplified to `!dialog_title.empty?`.

test/inst_install_inf_test.rb
```diff
@@ -0,0 +1,46 @@
...
+  describe "#main" do
+    it "writes the network configuration given by linuxrc" do
+      expect(Yast::InstallInfConvertor.instance).to receive(:write_netconfig)
+
+      subject.main
+    end
+
+    context "when a regurl is provided by linuxrc" do
+      let(:invalid_url) { "http://wrong_url{}.com" }
+      let(:valid_url) { "http://scc.custom.com" }
```
lslezak: NP: the default URL scheme is **https**.

src/lib/installation/dialogs/url_dialog.rb
teclator: true :)

test/inst_install_inf_test.rb
teclator: but is also considered a valid URL, isn't it?

teclator:
@lslezak thnx!

PR #565, opened 2017-05-29
Add 'cwm/widget' requirement because of changes made for (boo#1039901)

teclator:
In this PR yast/yast-yast2#573, the requirement of "cwm/widget" was moved to a method probably because of some recursive dependency conflict. So I have replaced the import by a requirement.


teclator:
@mvidner thnx ;)

PR #566, opened 2017-05-30
Don't forget desktop selection (bsc#1033594)

teclator:
We have fixed two different issues,

First of all it fixes the lost of desktop selection when the "Online repositories" is clicked (bsc#1033594)

And secondly we now requires at least the selection of one desktop to continue with the installation fixing (bsc#1040884)

Trello card: https://trello.com/c/40ZZxO2D/

A new card has been created to take care or handle by default the events not only of the main widget but also the buttons below it https://trello.com/c/0gsQI7Gf

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,9 +1,9 @@
 -------------------------------------------------------------------
 Tue May 31 21:47:24 UTC 2017 - knut.anderssen@suse.com
 
-- Don't lose the last desktop selection when online repositories is
-  called. (bsc#1033594)
-- Don't allow to continue without desktop selection. (bsc#1040884)
+- Do not lose the desktop selection when the online repositories is
+  pressed. (bsc#1033594)
+- Do not allow to continue without desktop selection. (bsc#1040884)
 - 3.2.43
 
 -------------------------------------------------------------------
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue May 31 21:47:24 UTC 2017 - knut.anderssen@suse.com
+
+- Don't lose the last desktop selection when online repositories is
+  called. (bsc#1033594)
+- Don't allow to continue without desktop selection. (bsc#1040884)
+- 3.2.43
+
+-------------------------------------------------------------------
 Wed May 31 10:52:59 UTC 2017 - lslezak@suse.cz
 
 - Fixed path to the "adddir" command (it is /sbin instead of /etc)
--- a/src/lib/installation/widgets/system_roles_radio_buttons.rb
+++ b/src/lib/installation/widgets/system_roles_radio_buttons.rb
@@ -66,7 +66,7 @@ module Installation
       end
 
       def validate
-        return true if !value.to_s.empty?
+        return true if value
 
         # TRANSLATORS: Popup error requesting to choose some option.
         Yast::Popup.Error(_("You must choose some option before you continue."))
--- a/src/lib/installation/widgets/system_roles_radio_buttons.rb
+++ b/src/lib/installation/widgets/system_roles_radio_buttons.rb
@@ -39,8 +39,8 @@ module Installation
       alias_method :store_orig, :store
 
       def initialize
-        # We need to handle all the events because if not the current selection
-        # is lost when the widget is redrawn (bsc#1033594)
+        # We need to handle all the events because otherwise the current
+        # selection is lost when the widget is redrawn (bsc#1033594)
         self.handle_all_events = true
       end
 
@@ -69,7 +69,7 @@ module Installation
         return true if !value.to_s.empty?
 
         # TRANSLATORS: Popup error requesting to choose some option.
-        Yast::Popup.Error(_("You must choose some option before continue."))
+        Yast::Popup.Error(_("You must choose some option before you continue."))
 
         false
       end
--- a/src/lib/installation/widgets/system_roles_radio_buttons.rb
+++ b/src/lib/installation/widgets/system_roles_radio_buttons.rb
@@ -40,7 +40,8 @@ module Installation
 
       def initialize
         # We need to handle all the events because otherwise the current
-        # selection is lost when the widget is redrawn (bsc#1033594)
+        # selection is lost when the widget is redrawn.
+        # FIXME: It will not be needed once RadioButtons widget take care of it
         self.handle_all_events = true
       end
 
--- a/src/lib/installation/widgets/system_roles_radio_buttons.rb
+++ b/src/lib/installation/widgets/system_roles_radio_buttons.rb
@@ -37,6 +37,13 @@ module Installation
       include SystemRoleReader
 
       alias_method :store_orig, :store
+
+      def initialize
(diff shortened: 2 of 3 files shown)


package/yast2-installation.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue May 30 07:34:24 UTC 2017 - knut.anderssen@suse.com
+
+- Don't lose the last desktop selection when online repositories is
+  called. (bsc#1033594)
```
imobachgs: np: I would say "online repositories button is pressed" or something like that.

package/yast2-installation.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue May 30 07:34:24 UTC 2017 - knut.anderssen@suse.com
+
+- Don't lose the last desktop selection when online repositories is
```
imobachgs: Try to not use the short form. Prefer "Do not" to "Don't".

src/lib/installation/widgets/system_roles_radio_buttons.rb
```diff
@@ -37,6 +37,13 @@ class SystemRolesRadioButtons < CWM::RadioButtons
       include SystemRoleReader
 
       alias_method :store_orig, :store
+
+      def initialize
+        # We need to handle all the events because if not the current selection
```
imobachgs: np: *if not* -> *otherwise*

src/lib/installation/widgets/system_roles_radio_buttons.rb
```diff
@@ -52,6 +59,21 @@ def store
...
+        ::Installation::SystemRole.select(value)
+
+        nil
+      end
+
+      def validate
+        return true if !value.to_s.empty?
+
+        # TRANSLATORS: Popup error requesting to choose some option.
+        Yast::Popup.Error(_("You must choose some option before continue."))
```
imobachgs: I'm not sure about this: "before you continue"?

test/lib/widgets/system_roles_radio_buttons_test.rb
```diff
@@ -32,4 +33,75 @@
       end
     end
   end
+
+  describe "#handle" do
+    it "selects the system role according to the current value" do
+      expect(Installation::SystemRole).to receive(:select).with(default)
```
imobachgs: Is this call something that you want to test? If it is only a condition for the test, you should use `allow`.

test/lib/widgets/system_roles_radio_buttons_test.rb
```diff
@@ -32,4 +33,75 @@
...
+    it "returns nil" do
+      expect(Installation::SystemRole).to receive(:select).with(default)
+
+      expect(widget.handle).to eql(nil)
+    end
+  end
+
+  describe "#init" do
+    it "initializes the widget with the current system role" do
+      expect(Installation::SystemRole).to receive(:current).and_return("server")
```
imobachgs: `allow`?

test/lib/widgets/system_roles_radio_buttons_test.rb
```diff
@@ -32,4 +33,75 @@
...
+  end
+
+  describe "#validate" do
+    let(:value) { nil }
+
+    before do
+      allow(widget).to receive(:value).and_return(value)
+    end
+
+    context "when the value of the widget is nil" do
```
imobachgs: np: you could rephrase as `when no option is selected`.

test/lib/widgets/system_roles_radio_buttons_test.rb
```diff
@@ -32,4 +33,75 @@
...
+      it "opens an error popup" do
+        expect(Yast::Popup).to receive(:Error)
+
+        expect(widget.validate).to eql(false)
```
imobachgs: np: why are you setting an expectation here? Just calling `widget.validate` should be better. After all, you're testing the return value in the following test.

test/lib/widgets/system_roles_radio_buttons_test.rb
```diff
@@ -32,4 +33,75 @@
...
+      end
+    end
+
+    context "when the value of the widget is an empty string" do
+      let(:value) { "" }
+
+      it "opens an error popup" do
+        expect(Yast::Popup).to receive(:Error)
+
+        expect(widget.validate).to eql(false)
```
imobachgs: Same as previous comment: just `widget.validate`.

src/lib/installation/widgets/system_roles_radio_buttons.rb
```diff
@@ -37,6 +37,13 @@ class SystemRolesRadioButtons < CWM::RadioButtons
       include SystemRoleReader
 
       alias_method :store_orig, :store
+
+      def initialize
+        # We need to handle all the events because if not the current selection
+        # is lost when the widget is redrawn (bsc#1033594)
+        self.handle_all_events = true
```
jreidinger: maybe mention here, that it won't be no longer needed when RadioButtons widget will be fixed. So we can remove it in future.
I also think that it makes more sense to associate bsc with handle and that select, as it is important part that fixes issue, not this line.

src/lib/installation/widgets/system_roles_radio_buttons.rb
```diff
@@ -52,6 +59,21 @@ def store
...
       end
 
+      def handle
+        ::Installation::SystemRole.select(value)
+
+        nil
+      end
+
+      def validate
+        return true if !value.to_s.empty?
```
jreidinger: this looks a bit strange. What is in fact value when not selected? if nil then you can use `return true if value`.

test/lib/widgets/system_roles_radio_buttons_test.rb
jreidinger: actually it is what is whole this it statement about. Storing in handle role value.

test/lib/widgets/system_roles_radio_buttons_test.rb
```diff
@@ -32,4 +33,75 @@
...
+
+  describe "#handle" do
+    it "selects the system role according to the current value" do
+      expect(Installation::SystemRole).to receive(:select).with(default)
+
+      widget.handle
+    end
+
+    it "returns nil" do
+      expect(Installation::SystemRole).to receive(:select).with(default)
```
jreidinger: here it should be just allow.

test/lib/widgets/system_roles_radio_buttons_test.rb
```diff
@@ -32,4 +33,75 @@
...
+
+      expect(widget.handle).to eql(nil)
```
jreidinger: :+1: for this test. in past we forget about it and it have nasty consequences.

src/lib/installation/widgets/system_roles_radio_buttons.rb
teclator: I just added the FIXME and removed the bsc

teclator:
@imobachgs, @jreidinger thnx for reviewing

teclator:
@jreidinger I updated it with your suggestions, so maybe you want to do a final check before merge it.

PR #567, opened 2017-05-31
Fixed path to the "adddir" command (bsc#1041882)

lslezak:
- It is in `/sbin` instead of `/etc`

- It's related to https://github.com/openSUSE/installation-images/pull/185

- 3.2.42

- Additionally updated the `.gitignore` file (removed the obsolete autotools temporary files)


okurz:
I'm interested. Why couldn't the issue that is supposed to be fixed not be catched by tests before QA SLE validation?

lslezak:
@okurz Because the tests are unit tests only, they tested that `/etc/adddir` was called regardless it existed or not.

For better testing you need integration tests and that's actually done by openQA. And having yet another integration test *before* the openQA integration test does not make sense.

okurz:
ok, I still wonder why it has not been catched by the staging openQA tests on the SR. I guess self-update was disabled for staging tests

lslezak:
Yes, that could be the reason.

PR #568, opened 2017-06-01
[Review] Request from 'schubi2' @ 'yast/yast-installation/review_170601_custom_roles'

schubi2:
  * 68b9065 Initialize pattern selection in the custom role workflow with patterns defined in the control file.

This should be submitted together with
https://github.com/yast/skelcd-control-openSUSE/pull/80
But also runs without it.

### DIFF
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.2.42
+Version:        3.2.44
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/src/lib/installation/widgets/system_roles_radio_buttons.rb
+++ b/src/lib/installation/widgets/system_roles_radio_buttons.rb
@@ -57,8 +57,10 @@ module Installation
           Yast::DefaultDesktop.SetDesktop(nil)
           # But select patterns which have been defined e.g. in the
           # software/default_patterns section in the control.xml file.
-          # (Can be done in the general default_patterns section or in
-          #  the role specific default_patterns section)
+          # (This can be done in the general default_patterns section or in
+          # the role specific default_patterns section.)
+          # Another possibility to set it are the default pattern settings
+          # in each product description.
           Yast::Packages.SelectSystemPatterns(false)
           Yast::Pkg.PkgSolve(false)
         else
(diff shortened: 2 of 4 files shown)


jreidinger:
code wise it looks good. Do you test it manually?

schubi2:
Yes, tested very "intensively" :-)

package/yast2-installation.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.2.42
+Version:        3.2.44
```
lslezak: At first I thought that there is typo, but the version is correct actually, it was not increased in the previous change.

So it's OK...

src/lib/installation/widgets/system_roles_radio_buttons.rb
```diff
@@ -50,9 +51,16 @@ def store
...
         store_orig
 
+        Yast::Packages.Reset([])
         if value == "custom"
           # for custom role do not use any desktop
           Yast::DefaultDesktop.SetDesktop(nil)
+          # But select patterns which have been defined e.g. in the
+          # software/default_patterns section in the control.xml file.
+          # (Can be done in the general default_patterns section or in
+          #  the role specific default_patterns section)
```
lslezak: The `Packages.SelectSystemPatterns` also selects the patterns defined by the product `defaultpattern()` dependency.

(See https://github.com/yast/yast-packager/blob/master/src/modules/Packages.rb#L2557 and https://github.com/yast/yast-packager/blob/dff498280c3969db2a386e9face338af36e24a4b/src/lib/packager/product_patterns.rb#L56).

IMHO the other source of the default patterns should be mentioned here as well.

src/lib/installation/widgets/system_roles_radio_buttons.rb
```diff
@@ -57,8 +57,10 @@ def store
           Yast::DefaultDesktop.SetDesktop(nil)
           # But select patterns which have been defined e.g. in the
           # software/default_patterns section in the control.xml file.
-          # (Can be done in the general default_patterns section or in
-          #  the role specific default_patterns section)
+          # (This can be done in the general default_patterns section or in
+          # the role specific default_patterns section.)
+          # Another possibility to set it are the default pattern settings
+          # in each product description.
```
lslezak: Another improvement:
```
defined by the "defaultpattern()" "Provides" dependency (FATE#320199).
```

PR #570, opened 2017-06-14
adapted re-probing to storage-ng

aschnell:
For https://trello.com/c/nxgcriYy/612-2-storageng-fix-instdisksactivate-client.


ancorgs:
The code looks good. But I would update the dependency on yast-storage-ng to a version including the `StorageManager#probe` method. That is, to whatever version we have after merging https://github.com/yast/yast-storage-ng/pull/260

ancorgs:
LGTM

PR #574, opened 2017-06-20
use only y2storage classes and methods

wfeldt:


ancorgs:
LGTM. Thanks for adding the comment

src/lib/installation/prep_shrink.rb
```diff
@@ -48,6 +56,9 @@ def run(*args)
 
     MAXIMAL_SIZE_KB = 8192
     def shrink_partitions
+# storage-ng
+# rubocop:disable Style/BlockComments
+=begin
```
ancorgs: Is this already documented in installer_hacks?

src/lib/installation/prep_shrink.rb
```diff
@@ -58,6 +69,7 @@ def shrink_partitions
           Yast::SCR.Execute(YAST_BASH_PATH, cmd)
         end
       end
+=end
```
ancorgs: Nitpicking: we should probably re-enable rubocop

src/lib/installation/prep_shrink.rb
wfeldt: https://github.com/yast/yast-storage-ng/pull/267

src/lib/installation/prep_shrink.rb
wfeldt: It will be gone anyway when the block is removed.

ancorgs:
LGTM again

PR #579, opened 2017-06-28
skip writing of desktop for opensuse (bsc#1030873)

jreidinger:
tested manually and it do what expected ( keep /etc/sysconfig/windowmanager ) untouched, but it still do not work, so more discussion in bnc is needed.

trello: https://trello.com/c/Nesqzfs9/1053-3-dont-write-default-desktop-if-none-selected-opensuse-only
bsc: https://bugzilla.opensuse.org/show_bug.cgi?id=1030873

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -2,7 +2,7 @@
 Thu Jun 29 08:40:43 UTC 2017 - jreidinger@suse.com
 
 - openSUSE only: when custom role is selected in desktop selection
-  keep default value in /etc/sysconfig/windowmanager
+  keep default value in /etc/sysconfig/windowmanager (bsc#1030873)
 - 3.2.46
 
 -------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jreidinger:
in the end decided to use this approach and fix aaa_base to have by default windowmanager "default" which works as expected instead of "kde-plasma".

package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jun 29 08:40:43 UTC 2017 - jreidinger@suse.com
+
+- openSUSE only: when custom role is selected in desktop selection
+  keep default value in /etc/sysconfig/windowmanager
+- 3.2.46
+
```
joseivanlopez: Is it not necessary to add the bug id here?

jreidinger: yes, it is required, thanks for notice

PR #580, opened 2017-06-29
Desktop fix master

jreidinger:
more then just a merge. Code improved and tests added.

### DIFF
--- a/src/clients/desktop_finish.rb
+++ b/src/clients/desktop_finish.rb
@@ -1,2 +1,2 @@
 require "installation/clients/desktop_finish"
-Yast::DesktopFinishClient.new.main
+Yast::DesktopFinishClient.run
--- a/src/lib/installation/clients/desktop_finish.rb
+++ b/src/lib/installation/clients/desktop_finish.rb
@@ -62,6 +62,7 @@ module Yast
       end
 
       desktop_map = DefaultDesktop.GetAllDesktopsMap[selected_desktop]
+      raise "Selected desktop '#{selected_desktop}' missing in desktops map" unless desktop_map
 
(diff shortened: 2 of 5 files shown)


src/lib/installation/clients/desktop_finish.rb
```diff
@@ -31,230 +31,97 @@
...
+        return nil
+      end
 
-          SCR.Write(path(".sysconfig.windowmanager.DEFAULT_WM"), @default_wm)
-          SCR.Write(
-            path(".sysconfig.windowmanager.X_MOUSE_CURSOR"),
-            @default_cursor
-          )
-          SCR.Write(path(".sysconfig.windowmanager"), nil)
+      desktop_map = DefaultDesktop.GetAllDesktopsMap[selected_desktop]
```
kobliha: Maybe raise an exception if there is no desktop_map for a selected_desktop?

src/clients/desktop_finish.rb
```diff
@@ -1,2 +1,2 @@
 require "installation/clients/desktop_finish"
-Yast::DesktopFinishClient.new.main
+Yast::DesktopFinishClient.run
```
lslezak: Any reason for this change? I mean I'm fine with a static class method, but we should use the same approach in all clients here. The other client files in the package use the `.new.main` way. It should be consistent...

jreidinger: reason is that previously it was simple Yast::Client with common method `main`. Now it is finish client, that have in its API method `run`.

lslezak: I see, it's a `FinishClient` now... then OK.

PR #584, opened 2017-07-06
Disable plymouth by a no affend method, it will not block tty device activation at first time user login(bsc#1042554).

qiangzhao:


package/yast2-installation.changes
```diff
@@ -1,4 +1,15 @@
 -------------------------------------------------------------------
+Thu Jul  6 13:12:12 UTC 2017 - qzhao@suse.com
+
+- Update YaST2-Firstboot.service: Depreciate plymouth --wait, add
+  conflict to plymouth start service.
+- Update YaST2-Second-Stage.service: Depreciate plymouth deactivate
+  command and add conflict to plymouth start service. For they will
```
imobachgs: "For they"? What about something like "For that" or just "TTY device activation will be blocked."

qiangzhao:
@wfeldt , Could you give me some advice about the latest modification ?

okurz:
@imobachgs @wfeldt you both approved the PR but it has not been merged. Now there's a conflict in changes files - as expected after some time - but is there a specific reason why the PR is not merged? That fix should have gone into SLE 12 SP3 but therefore failed to do so. I am wondering if there is something wrong in this PR or if it was just overlooked?

imobachgs:
Superseded by #642. It is the same, bug fixing a problem in the changes file. Thanks!

PR #585, opened 2017-07-17
initial version of product selection

jreidinger:
real example now with filtering using SLE 15 dist repo
leanos_install


src/clients/inst_product_selector.rb
```diff
@@ -0,0 +1,5 @@
+require "yast"
+
+require "installation/dialogs/product_selection"
```
lslezak: NP: We should probably unify the names, the client is called product **selector**, the dialog is called product **selection**.

src/lib/installation/product_reader.rb
```diff
@@ -0,0 +1,96 @@
...
+      products
+    end
+
+    def self.installation_package_mapping
+      installation_packages = Yast::Pkg.PkgQueryProvides("system-installation()")
+
+      mapping = {}
+      installation_packages.each do |list|
+        pkg_name = list.first
+        # FIXME: can be there more result then one?
```
lslezak: Yes, there can be more instances of the same package. One from the installation medium, the other from -Pool and -Updates repositories added by registration later.

To make it simple we could just take the first found, if we want to implement it better we can  improve it later (just put a comment here).

src/lib/installation/dialogs/product_selection.rb
```diff
@@ -0,0 +1,59 @@
...
+require "cwm/dialog"
+require "installation/widgets/product_selector"
+require "installation/product_reader"
+
+Yast.import "ProductControl"
+Yast.import "WorkflowManager"
+
+module Installation
+  module Dialogs
+    class ProductSelection < CWM::Dialog
```
imobachgs: What about adding some documentation? At least a brief description.

src/lib/installation/dialogs/product_selection.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      def contents
+        VBox(selector)
+      end
+
+      def run
+        res = super
+
+        if res == :next
```
imobachgs: What about using `return res unless res == :next` and avoiding the indentation?

src/lib/installation/product.rb
```diff
@@ -0,0 +1,49 @@
...
+
+require "yast"
+Yast.import "Pkg"
+
+module Installation
+  # Simple Libzypp Product wrapper
+  class Product
+    include Yast::Logger
+
+    attr_reader :name, :label, :order
```
imobachgs: NP: What about documenting each method? On the other hand, in this case, the very same values are passed to the constructor. So I can live with it as it is :)

src/lib/installation/product.rb
```diff
@@ -0,0 +1,49 @@
...
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ------------------------------------------------------------------------------
+
+require "yast"
+Yast.import "Pkg"
+
+module Installation
+  # Simple Libzypp Product wrapper
+  class Product
```
imobachgs: Having this wrapper looks like a good idea 👍

src/lib/installation/product_reader.rb
```diff
@@ -0,0 +1,96 @@
...
+
+Yast.import "Pkg"
+
+module Installation
+  # Read the product information from libzypp
+  class ProductReader
+    include Yast::Logger
+
+    # In installation Read the available libzypp base products for installation
+    # @return [Array<Installation::Product>] the found available base products
```
imobachgs: Does it work in a different way if it is not run in installation?

src/lib/installation/product_reader.rb
```diff
@@ -0,0 +1,96 @@
...
+    def self.available_base_products
+      products = base_products
+
+      installation_mapping = installation_package_mapping
+      result = products.map do |p|
```
imobachgs: What about using a more meaningul name as `prod`?

src/lib/installation/product_reader.rb
```diff
@@ -0,0 +1,96 @@
...
+    def self.product_package(name, repo_id)
+      return nil unless name
+      Yast::Pkg.ResolvableDependencies(name, :package, "").find do |p|
+        p["source"] == repo_id
+      end
+    end
+
+    def self.base_products
+      products = Yast::Pkg.ResolvableProperties("", :product, "").select do |p|
+        # during installation/upgrade the product["type"] is not valid yet yet
```
imobachgs: You could move this explanation to the method documentation. It is a implementation detail, but I think it is interesting enough.

src/lib/installation/product_reader.rb
imobachgs: I would add more details from @lslezak comment (like the Pool and Updates repositories detail).

src/lib/installation/product_reader.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    def self.installation_package_mapping
+      installation_packages = Yast::Pkg.PkgQueryProvides("system-installation()")
+
+      mapping = {}
+      installation_packages.each do |list|
+        pkg_name = list.first
+        # FIXME: can be there more result then one?
+        dependencies = Yast::Pkg.ResolvableDependencies(pkg_name, :package, "").first["deps"]
```
imobachgs: Can "deps" be nil?

src/lib/installation/product_reader.rb
```diff
@@ -0,0 +1,96 @@
...
+        install_provide = dependencies.find do |d|
+          d["provides"] && d["provides"].match(/system-installation\(\)/)
+        end
+
+        product_name = install_provide["provides"][/system-installation\(\)\s*=\s*(\S+)/, 1]
```
imobachgs: why do you need the first pair of parenthesis?

src/lib/installation/product_sorter.rb
```diff
@@ -0,0 +1,30 @@
...
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ------------------------------------------------------------------------------
+
+module Installation
+  # Sorter for sorting Products in required display order
+  # @param x [Installation::Product] the first item to compare
+  # @param y [Installation::Product] the second item to compare
+  PRODUCT_SORTER = proc do |x, y|
```
imobachgs: Why using a constant? Where is this sorter used?

src/lib/installation/widgets/product_selector.rb
```diff
@@ -0,0 +1,59 @@
...
+      def hspacing
+        1
+      end
+
+      def label
+        _("Product to Install")
+      end
+
+      def init
+        selected = products.find { |p| p.selected? }
```
imobachgs: Why not use `products.find(&:selected?)` ?

src/clients/inst_product_selector.rb
jreidinger: well, it was intention. Selector is widget, selection is action that do dialog. Dialog do selection using selector widget. ( it also avoids name collision :)

src/lib/installation/widgets/product_selector.rb
```diff
@@ -40,7 +48,12 @@ def store
         @product.select
       end
 
-      # TODO: validation, a product must be selected
+      def validation
+        return true if value
+
+        Yast::Popup.Error(_("Please select product to install."))
```
lslezak: NP: Please select **a** product to install.

src/clients/inst_product_selector.rb
lslezak: Then OK, I can live with that... :smile:

src/lib/installation/product_sorter.rb
lslezak: I used the [same approach as in the registration module](https://github.com/yast/yast-registration/blob/master/src/lib/registration/addon_sorter.rb). 

Oh, I used it in a dialog, either it was lost at some merge or I forgot to commit/push... I'll check

src/lib/installation/product_reader.rb
jreidinger: no, as I originally search for package that have in deps specific key "system-installation". So if it is empty, there is something going really bad.

src/lib/installation/product_reader.rb
jreidinger: you mean in regexp? it is escaped one as provides look like
```
"system-installation() = SLES"
```
Maybe I can add to comment example how provides looks like.

src/lib/installation/widgets/product_selector.rb
jreidinger: rubocop already complain, so updated ;)

src/lib/installation/product_sorter.rb
lslezak: Thanks for spotting this! I have definitely included the sorter in the workflow, but I did not notice it was lost in the final diff.

Added back again.

jreidinger:
manual testing proof that product sorting works.

src/lib/installation/product_reader.rb
jreidinger: well, actually this "hack" should be removed, as we no longer want to use src_id 0 only. We will filter it only by presence of corresponding system-installation package.

src/lib/installation/product_reader.rb
```diff
@@ -55,25 +55,23 @@ def self.available_base_products
...
-        # during installation, so the base product is from repo ID 0
-        p["source"] == 0
+      products = Yast::Pkg.ResolvableProperties("", :product, "").select do |prod|
+        prod["source"] == 0
       end
 
       # remove duplicates, there migth be different flavors ("DVD"/"POOL")
       # or archs (x86_64/i586)
-      products.uniq! { |p| p["name"] }
+      products.uniq! { |prod| prod["name"] }
```
imobachgs: I did not mean to change this one: the `p` was ok because you are user a one-liner (although it shadows the `#p` method -you could consider using `i` or whatever-). Just see our [style guide](https://github.com/SUSE/style-guides/blob/master/Ruby.md#parameters).

src/lib/installation/product_reader.rb
lslezak: This is designed to work only in installation, the running system is out of scope.

src/lib/installation/product_reader.rb
lslezak: OK, I was too eager at the rename... :wink:

PR #588, opened 2017-07-28
Adapt InstPreInstall to storage-ng

ancorgs:
This needs more manual testing (or even creating new automated tests) before being merged.

### DIFF
--- a/src/lib/installation/clients/inst_pre_install.rb
+++ b/src/lib/installation/clients/inst_pre_install.rb
@@ -228,8 +228,7 @@ module Yast
           count = 0
           disks = disks.select do |dist|
             next true unless dist.start_with?(s)
-            count += 1
-            count <= 8
+            (count += 1) <= 8
           end
 
           nil
@@ -249,7 +248,7 @@ module Yast
       device_names = restrict_disk_names.call(device_names)
       Builtins.foreach(device_names) do |device_name|
         device = Y2Storage::BlkDevice.find_by_name(probed, device_name)
-        filesystems = device.descendants.select {|i| i.is?(:blk_filesystem) }
+        filesystems = device.descendants.select { |i| i.is?(:blk_filesystem) }
         filesystems.each do |filesystem|
           device = filesystem.blk_devices.first
           if !filesystem.type.root_ok?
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/inst_pre_install.rb
```diff
@@ -225,13 +224,12 @@ def Initialize
...
         helper = lambda do |s|
           count = 0
-          disks = Builtins.filter(disks) do |dist|
-            next true if Builtins.search(dist, s) != 0
-            count = Ops.add(count, 1)
-            Ops.less_or_equal(count, 8)
+          disks = disks.select do |dist|
+            next true unless dist.start_with?(s)
+            count += 1
+            count <= 8
```
wfeldt: Not being a ruby expert this looks weird to me; what's wrong with simply `(count += 1) <= 8`?

src/lib/installation/clients/inst_pre_install.rb
```diff
@@ -225,13 +224,12 @@ def Initialize
       # mounting of all partitions (fate#305873, bnc#468922)
       # FIXME: copy-pasted from partitioner, just different number of disks and added /dev/dasd
       restrict_disk_names = lambda do |disks|
```
wfeldt: After thinking about this for a while I'm still not happy with this concept. In particular as we don't seem to care at all for updates here:

https://github.com/yast/yast-update/blob/master/src/modules/RootPart.rb#L2066-L2069

If at all this functionality belongs to yast-storage-ng.

Issues I have with this concept:
- does not care about other device classes (vd*, dm*, and all the modern stuff)
- it makes only sense to cut the device list if the number of file systems exceeds a certain limit (say, - completely random - 32)

wfeldt:
lgtm (but for the rubocop offense)

src/lib/installation/clients/inst_pre_install.rb
ancorgs: Both are perfectly valid and totally equivalent and none of them look weird to my eyes. :) I'm fine with changing it if you dislike the original for any reason.

wfeldt: no need, I lgtm'ed below; was just some thought

PR #593, opened 2017-08-16
Make fs type for home and root configurable

shundhammer:
https://trello.com/c/SQQ1x4Bg/691-5-suse-caasp-10-p2-1051762-docker-on-xfs-by-default

This makes the filesystem type for both home and root in the storage proposal configurable: When home_fs is configurable, root_fs is only a heartbeat away, and it's trivial to add it at the same time.

Test against SLE-12 SP3:

home_fs_sp3

### DIFF
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -1054,6 +1054,12 @@ property should be set to _true_. This works only for Btrfs root
 filesystems. If another root filesystem type is chosen, this might fail
 silently.
 
+*root_fs* (string, default _btrfs_) is the filesystem type for the root
+ partition.
+
+*home_fs* (string, default _xfs_) is the filesystem type for the home
+ partition.
+
 *home_path* (string) is the path (mount point) for the home
 partition or volume, if any is created (depending on *try_separate_home*,
 *limit_try_home* and available disk space).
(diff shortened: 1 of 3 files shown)


doc/control-file.md
```diff
@@ -1054,6 +1054,12 @@ property should be set to _true_. This works only for Btrfs root
 filesystems. If another root filesystem type is chosen, this might fail
 silently.
 
+*root_fs* (string, default _btrfs_) is the filesystem type for the root
+ partition.
```
ancorgs: We should document that, is not specified, the values from `/etc/sysconfig` would be used. So strictly speaking, the value is not `btrfs`.

doc/control-file.md
```diff
@@ -1054,6 +1054,12 @@ property should be set to _true_. This works only for Btrfs root
...
+
+*home_fs* (string, default _xfs_) is the filesystem type for the home
+ partition.
+
```
ancorgs: Same than above, the default is not exactly `xfs`, but whatever is specified in `/etc/sysconfig`.

doc/control-file.md
shundhammer: This is only for installation (as control.xml always is), so that's irrelevant.

doc/control-file.md
shundhammer: Same as above.

shundhammer: Not quite: There is no corresponding value for `home_fs` in /etc/sysconfig/storage.

PR #595, opened 2017-08-18
Merge product class

imobachgs:
Move `Packages::Product`, `Packages::ProductReader` and `Packages::ProductSorter` classes.(FATE#322276). See yast/yast-packager#263 for more details.

PR #597, opened 2017-08-24
Switch to yastdevel/ruby Docker image

imobachgs:


imobachgs:
Closing because I've included this change as part of #595.

PR #606, opened 2017-09-05
Download release notes only for the selected products (bsc#1056872)

lslezak:
- If any product is already selected then download the release notes only for the selected product (LeanOS). If nothing is selected download release notes for all available products - for backward compatibility with a single product medium (openSUSE).

- This is similar to the fix in registration: https://github.com/yast/yast-registration/pull/327

- 4.0.0

### DIFF
--- a/src/lib/installation/clients/inst_download_release_notes.rb
+++ b/src/lib/installation/clients/inst_download_release_notes.rb
@@ -130,9 +130,8 @@ module Yast
       # Get proxy settings (if any)
       proxy = curl_proxy_args
 
-      # installed may mean old (before upgrade) in initial stage
-      # product may not yet be selected although repo is already added
-      required_product_statuses = Stage.initial ? [:selected, :available] : [:selected, :installed]
+      required_product_statuses = check_product_states
+      log.info("Checking products in state: #{required_product_statuses}")
       products = Pkg.ResolvableProperties("", :product, "").select do |product|
         required_product_statuses.include? product["status"]
       end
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/inst_download_release_notes.rb
```diff
@@ -241,5 +240,24 @@ def main
...
+      # product may not yet be selected although repo is already added
+      return [:selected, :installed] unless Stage.initial
+
+      # if a product is already selected then use the selected products
+      # otherwise use the available one(s)
+      product_selected = Pkg.ResolvableProperties("", :product, "").any? do |p|
+        p["status"] == :selected
+      end
+
+      product_selected ? [:selected] : [:available]
```
jreidinger: I thought that we want to change also opensuse, so always product is selected, not?

lslezak: @imobachgs how the new welcome screen (with product selection) behaves when there is only one product? How does it behave in openSUSE?

imobachgs: If there is only 1 product in the media, it will be selected by default. You can find more info here: https://github.com/yast/yast-packager/blob/master/src/lib/y2packager/clients/inst_repositories_initialization.rb#L80

lslezak: In that case it will use the selected product which is correct.

The `:available` flag is there rather as a fallback to work correctly in the backward compatible way.

So to me it seems OK.

PR #609, opened 2017-09-18
Release notes from packages

imobachgs:
Adapt `inst_download_release_notes` client to the new API. Depends on https://github.com/yast/yast-packager/pull/279. Travis is failing because `yast2-packager 4.0.5` is not available yet.


src/lib/installation/clients/inst_system_analysis.rb
```diff
@@ -346,7 +329,7 @@ def InitInstallationRepositories
         # initialize libzypp and get the base product name (intentionally not translated)
         UI.SetProductName(Product.name || "SUSE Linux")
 
-        download_and_show_release_notes
+        WFM.CallFunction("inst_download_release_notes")
```
lslezak: Does it work correctly when called later again? I mean here it is started at the system analysis step which happens **before** the installation product is selected and then it's run once more later [after the add-on step](https://github.com/yast/skelcd-control-SLES/blob/master/control/installation.SLES.xml#L227).

src/lib/installation/clients/inst_download_release_notes.rb
```diff
@@ -25,186 +25,44 @@
...
-    }.freeze
+require "y2packager/product"
 
-    # Download *url* to *filename*
-    # May set InstData.stop_relnotes_download on download failure.
-    #
-    # @return [Boolean,nil] true: success, false: failure, nil: failure+dont retry
-    def curl_download(url, filename, proxy_args:, max_time: 300)
-      cmd = Builtins.sformat(
-        "/usr/bin/curl --location --verbose --fail --max-time %6 --connect-timeout 15  %1 '%2' --output '%3' > '%4/%5' 2>&1",
```
lslezak: So we are dropping the `curl` support? Your code downloads the highest version available but that will work *only when the system is registered!*

If you skip registration you have no chance to read the latest RN... :worried: 

Maybe we should keep the `curl` way as a fallback when not registered? This was not specified in the PBI, but maybe needs clarification/approval from PO/PM... ? (Maybe for the next sprint.)

kobliha: Next sprint, please

src/lib/installation/clients/inst_system_analysis.rb
imobachgs: Yes. All the information is cached and, when calling it again, it will just try to download information for new selected products or newer release notes packages. Then, `InstData.release_notes` will be updated accordingly.

lslezak: I see, thanks for the info!

imobachgs:
The coverage is decreasing due to the removal of [inst_system_analysis_client](https://github.com/yast/yast-installation/blob/master/test/inst_system_analysis_test.rb#L10). Those tests were only checking the `download_and_show_release_notes` which does not exist anymore. But, by the way, the `InstSystemAnalysis#main` method was executed as a side effect (and now it is not).

imobachgs:
https://github.com/yast/yast-packager/pull/284 contains the code to download release notes using cURL if needed and now yast2-installation does nothing about those details.

PR #611, opened 2017-09-26
Roles ordering

jreidinger:


src/lib/installation/clients/inst_modules_extensions.rb
```diff
@@ -0,0 +1,42 @@
...
+        log.info "module extension packages #{extension_packages.inspect}"
+
+        extension_packages.map do |list|
+          pkg_name = list.first
+          dependencies = Yast::Pkg.ResolvableDependencies(pkg_name, :package, "").first["deps"]
+          extension_provide = dependencies.find do |d|
+            d["provides"] && d["provides"].match(/installer_module_extension\(\)/)
+          end
+
+          module_name = extension_provide["provides"][/installer_module_extension\(\)\s*=\s*(\S+)/, 1]
```
lslezak: The `installer_module_extension` string is used several times here, it's quite long and prone to typos. So I'd prefer a constant for that.

test/select_system_role_test.rb
```diff
@@ -42,8 +42,9 @@
     context "when some roles are defined" do
       let(:control_file_roles) do
         [
-          { "id" => "foo", "partitioning" => { "format" => true } },
-          { "id" => "bar", "software" => { "desktop" => "knome" }, "additional_dialogs" => "a,b" }
+          { "id" => "foo", "order" => "100", "partitioning" => { "format" => true } },
+          { "id" => "bar", "order" => "200",
+            "software" => { "desktop" => "knome" }, "additional_dialogs" => "a,b" }
```
lslezak: I like Knome :-) If you do not know whether to use KDE or Gnome...

test/system_role_test.rb
```diff
@@ -44,6 +46,10 @@
       expect(described_class.all.size).to eql(2)
       expect(described_class.all.last.class).to eql(described_class)
     end
+
+    it "returns array sorted by order" do
+      expect(described_class.all.first.id).to eql("role_two")
```
lslezak: Then it should check all roles, not only the first one. We cannot be sure if it still works correctly after adding a third role.

PR #613, opened 2017-10-04
[Review] Request from 'schubi2' @ 'yast/yast-installation/review_171004_bnc_1055279'

schubi2:
  * 2fd602ba AY: warning user if an remote installation is not possible
  * bb3f25c9 AY: warning user if an remote installation is not possible


src/lib/installation/proposal_runner.rb
```diff
@@ -71,8 +73,11 @@ def initialize(store = ::Installation::ProposalStore)
     end
 
     def run
-      # skip if not interactive mode.
       if !Yast::AutoinstConfig.Confirm && (Yast::Mode.autoinst || Yast::Mode.autoupgrade)
+        # Checking if vnc, ssh,... is available
```
lslezak: Again, is it possible to add an unit test for this?

PR #614, opened 2017-10-10
Drop dasd reload

jreidinger:
https://trello.com/c/DCrlDL2j/1116-2-sle-15-remove-dasdreload-call-from-yast

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct 10 08:34:44 UTC 2017 - jreidinger@suse.com
+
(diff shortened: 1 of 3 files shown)


package/yast2-installation.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct 10 08:34:44 UTC 2017 - jreidinger@suse.com
+
+- drop call of dasd_reload as it cause renumbering of disks which
+  confuse users and hopefully less bug reports(FATE#318138)
```
mchf: NP: I'd use something like
"Dropped call of dasd_reload which caused renumbering of disks and user got confused."

PR #615, opened 2017-10-10
Do not show product selection during upgrade

teclator:
- [Trello Card](https://trello.com/c/BKCh35fR/1191-3-1059070-upgrade-case-let-user-continue-without-selecting-a-product-request-license-agreement-later)

This is an early release just to discuss the current dialog with UI/UX experts to check as soon as possible whether we should adapt it or not. 

upgrade_sle-12-sp2

upgrade_sle-12-sp2_textmode

screenshot_opensusetumbleweed_2017-10-16_00 34 28

### DIFF
--- a/src/lib/installation/dialogs/complex_welcome.rb
+++ b/src/lib/installation/dialogs/complex_welcome.rb
@@ -53,15 +53,9 @@ module Installation
       #
       # @return [String] Dialog's title
       def title
-        if show_product_selection?
-          if show_license?
-            _("Language, Keyboard and License Agreement")
-          else
-            _("Language, Keyboard and Product Selection")
-          end
-        else
-          _("Language and Keyboard Selection")
-        end
+        return product_title if show_product_selection?
+
+        _("Language and Keyboard Selection")
       end
 
       # Dialog content
@@ -140,10 +134,22 @@ module Installation
         products.size == 1
       end
 
+      # Determine whether the product selection must be shown
+      #
+      # The product selection will be shown only if there are at least some
+      # product available.
+      #
+      # @return [Boolean] true if the products list is not empty
       def show_product_selection?
         !products.empty?
       end
 
+      # Product content
+      #
+      # Shows the product selection if there is more than one product or the
+      # license agreement if there is only one.
+      #
+      # @return [Yast::Term] Product selection content; Empty() if no products
       def product_selection
         return Empty() unless show_product_selection?
 
@@ -156,6 +162,20 @@ module Installation
           Empty()
         else
           VWeight(1, VStretch())
+        end
+      end
+
+      # Title of the dialog in case there are some product available.
+      #
+      # The title can vary depending if the license agreement or the product
+      # selection is shown.
+      #
+      # @return [String] Dialog's title
+      def product_title
+        if show_license?
+          _("Language, Keyboard and License Agreement")
+        else
+          _("Language, Keyboard and Product Selection")
         end
       end
     end
--- a/src/lib/installation/dialogs/complex_welcome.rb
+++ b/src/lib/installation/dialogs/complex_welcome.rb
@@ -53,10 +53,14 @@ module Installation
       #
       # @return [String] Dialog's title
       def title
-        if show_license?
-          _("Language, Keyboard and License Agreement")
+        if show_product_selection?
+          if show_license?
+            _("Language, Keyboard and License Agreement")
+          else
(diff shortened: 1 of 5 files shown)


src/lib/installation/dialogs/complex_welcome.rb
```diff
@@ -53,10 +53,14 @@ def initialize(products, disable_buttons: [])
...
+        if show_product_selection?
+          if show_license?
+            _("Language, Keyboard and License Agreement")
+          else
+            _("Language, Keyboard and Product Selection")
+          end
         else
-          _("Language, Keyboard and Product Selection")
+          _("Language and Keyboard Selection")
         end
```
kobliha: This whole `if logic` makes sense now, but it will be weird later.
```
if show_product
  if show_license
    ...
  else
    ...
  else
    ...
  end
else
  ...
end
```

teclator: Yep, the code starts to smell, I was not sure by now which was the best way to handle the different titles and that is also why it is still in WIP ;)

As commented this is an early release to start discussions with UI/UX team ;), thnx for checking

src/lib/installation/dialogs/complex_welcome.rb
```diff
@@ -136,6 +134,28 @@ def show_license?
...
+      end
+
+      # Product content
+      #
+      # Shows the product selection if there is more than one product or the
+      # license agreement if there is only one.
+      #
+      # @return [Yast::Term] Product selection content; Empty() if no products
+      def product_selection
+        return Empty() unless show_product_selection?
```
imobachgs: The name of the method is confusing. Its name is `product_selection` although it shows the *product selector* or the *product license*. And also the `show_product_selection?` looks confusing to me. For me, the real condition is that there is no products to select.

test/inst_complex_welcome_test.rb
```diff
@@ -97,160 +97,181 @@
       subject.main
     end
 
-    context "when back is pressed" do
-      let(:dialog_results) { [:back] }
+    it "handles the dialog result" do
```
imobachgs: I am not sure if this test is ok: the fact that we call `handle_dialog_result` is an implementation detail. Actually, that method should be private (I know it was not part of this PR).

On the other hand, I would not remove the `when back is pressed it returns :back" because, IMHO, these are different things. For instance, if you break the `main` method and do no return `:back` anymore, the bug will be unnoticed.

test/inst_complex_welcome_test.rb
```diff
@@ -97,160 +97,181 @@
...
-        expect(subject).to receive(:setup_final_choice)
         subject.main
       end
 
-      it "returns nil" do
-        expect(subject.main).to be_nil
-      end
+      context "when no product is available for select" do
+        let(:dialog_results) { [:next] }
+        let(:productse) { [] }
```
imobachgs: typo?

test/inst_complex_welcome_test.rb
```diff
@@ -97,160 +97,181 @@
...
 
-      it "executes from next step" do
-        expect(Yast::ProductControl).to receive(:RunFrom)
-          .with(Yast::ProductControl.CurrentStep + 1, true)
-        subject.main
+        it "returns the result" do
+          expect(subject).to receive(:handle_dialog_result)
```
imobachgs: As I mentioned above, this one is an implementation detail for me.

test/inst_complex_welcome_test.rb
```diff
@@ -97,160 +97,181 @@
...
         end
+      end
+    end
+  end
 
-        context "when more than 1 product exists (it should be accepted later)" do
-          let(:dialog_results) { [:next, :back] }
-          let(:license_needed?) { true }
-          let(:products) { [product, other_product] }
+  describe "#handling_dialog_result" do
```
imobachgs: I do not think this method should be public.

test/inst_complex_welcome_test.rb
```diff
@@ -97,160 +97,181 @@
...
+          expect(subject.handle_dialog_result(:next)).to eq(nil)
         end
       end
+    end
+  end
 
-      context "when running in config mode" do
-        let(:dialog_results) { [:next, :back] }
-        let(:config_mode) { true }
+  describe "#product_selection_finished?" do
```
imobachgs: I do not think this method should be public.

PR #621, opened 2017-10-30
Always show license agreement even on upgrade mode

imobachgs:
It should fix [bsc#1065172](https://bugzilla.suse.com/show_bug.cgi?id=1065172).


src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -138,15 +138,21 @@ def setup_final_choice
...
+    # (as you cannot change the product during upgrade).
     #
     # @return [Array<Y2Packager::Product>] List of available base products;
     # empty list in update mode.
     def products
       return @products if @products
 
-      @products = Mode.update ? [] : Y2Packager::Product.available_base_products
+      @products = Y2Packager::Product.available_base_products
+      @products = [] if Mode.update && @products.size > 1
```
kobliha: NP: In fact, the comment probably belongs here (?)

teclator:
For me looks good and it is the simplest solution and with the current method's comment it is now more clear for others

src/lib/installation/clients/inst_complex_welcome.rb
teclator: If you read it without expand the code in RDOC it is more clear it we put the comment where it is.

kobliha: Absolutely

PR #623, opened 2017-11-02
Calling SUSEConnect --cleanup at startup

kobliha:
- When you abort an installation after you have registered your product, you can't just start again as selecting any product is then blocked later
- This is not needed for a single-product add-ons, such as openSUSE, but the cleanup redirects everything to /dev/null as it complains at the first start even at SUSE Enterprise anyway
- bsc#1065167

### DIFF
--- a/startup/First-Stage/F09-start
+++ b/startup/First-Stage/F09-start
@@ -7,6 +7,16 @@ log "====================="
 #---------------------------------------------
 callHooks preFirstCall
 
+# 9.05) cleanup before (re)starting
+
+# bsc#1065167: Installer does not allow to change a base product when it's
+# already registered. When installation is aborted and then restarted,
+# the data of the previous registration was not automatically removed.
+#
+# log "Cleaning up previous registration attempts"
+# Using /dev/null - If there is nothing to do, let it fail silently
+SUSEConnect --cleanup 1>/dev/null 2>/dev/null
+
 #=============================================
 # 9.1) check for driver update mode
 #---------------------------------------------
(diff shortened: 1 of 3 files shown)


startup/First-Stage/F09-start
```diff
@@ -7,6 +7,16 @@ log "====================="
...
 
+# 9.05) cleanup before (re)starting
+
+# bsc#1065167: Installer does not allow to change a base product when it's
+# already registered. When installation is aborted and then restarted,
+# the data of the previous registration was not automatically removed.
+#
+# log "Cleaning up previous registration attempts"
+# Using /dev/null - If there is nothing to do, let it fail silently
+SUSEConnect --cleanup 1>/dev/null 2>/dev/null
```
lslezak: Maybe run the cleanup only if the registration file exists (yes, a hardcoded file, but on the other hand cannot make troubles in openSUSE or if not registered):

```bash
if [ -e /etc/zypp/credentials.d/SCCcredentials ]; then
  echo "Running registration cleanup..."
  SUSEConnect --cleanup  > /dev/null 2>&1
  echo "Done"
fi
```
Maybe I'd even try de-registering first (use `timeout` to not block e.g. when there are network issues):

```bash
if [ -e /etc/zypp/credentials.d/SCCcredentials ]; then
  echo "Running registration cleanup..."
  timeout 20 SUSEConnect --de-register > /dev/null 2>&1
  echo "Done"
fi
```

De-registering will free the consumed subscription, that might be a good idea if you can register only one system with your key.

kobliha: I've thought about similar solutions, but there are some reasons for the current solution

- As simple as possible, this is a side-case
- No `path` to credentials to be defined in the code (even YaST uses SUSEConnect variable)
- No de-registration as it can take too much time, even 20 seconds is too much

lslezak: OK, so let's use the simple solution.

Just use the simpler redirection with `> /dev/null 2>&1`.

kobliha: OK

PR #627, opened 2017-11-30
Fix Travis CI

imobachgs:
* Documents #626.


imobachgs:
It is still failing because the new yast-rake version is not on the docker image yet.

imobachgs:
Travis is not expected to work in this case. See discussion [in #626](https://github.com/yast/yast-storage/pull/293#issuecomment-348163761).

imobachgs:
Superseeded by #628.

PR #628, opened 2017-11-30
Set AutoYaST activation callbacks for libstorage-ng

imobachgs:


imobachgs:
I've added changes from #627.

package/yast2-installation.spec
```diff
@@ -55,8 +49,8 @@ BuildRequires:	yast2-packager >= 4.0.9
 BuildRequires: yast2-storage-ng >= 0.1.32
```
ancorgs: This also depends on the latest version of yast2-storage-ng that accepts arguments in the call to `activate`

PR #630, opened 2017-12-04
Break dependency cycle

imobachgs:
I guess that a better solution for [bsc#1070996](https://bugzilla.suse.com/show_bug.cgi?id=1070996) (breaking autoyast2 and yast2-installation dependency cycle) could be used, which probably would imply moving some code between modules. For the time being, this one should work.


imobachgs:
Thanks!

PR #632, opened 2017-12-08
Log the system name for easier debugging

lslezak:
- Logs this at the very beginning:
```
[Ruby] clients/installation.rb:113 System identification: "SUSE Linux Enterprise 15"
```
- The value is read from `/ets/os-release` (included in inst-sys)
- It would help me a lot in [this case](https://bugzilla.suse.com/show_bug.cgi?id=1068302#c12)

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Dec 12 08:21:36 UTC 2017 - lslezak@suse.cz
+
+- Log the system name at start to see which system is running
+  in the inst-sys (for easier debugging)
+
+-------------------------------------------------------------------
 Mon Dec 11 15:50:04 UTC 2017 - knut.anderssen@suse.com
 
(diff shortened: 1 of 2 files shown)


package/yast2-installation.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Dec 12 08:21:36 UTC 2017 - lslezak@suse.cz
+
+- Log the system name at start to see which system is running
+  in the inst-sys (for easier debugging)
+
```
joseivanlopez: Is it not necessary bug/fate number and increase version in this package?

lslezak: No, it's just for internal debugging. The package will be submitted later together with the next "real" fix/feature.

lslezak: To the code coverage: because the coverage is too low and it's hard to add a new test for a trivial code change then a coverage decrease is kind of expected. :worried: 

But IMHO it still makes sense, at least we can see how bad it is here.

PR #633, opened 2017-12-08
fate#324198: filter modules extensions based on product

gilsonsouza:


src/lib/installation/clients/inst_modules_extensions.rb
```diff
@@ -21,11 +22,29 @@ def run
...
+      PROVIDES_PRODUCT = "extension_for_product()".freeze
 
       def extension_packages
+        product = Y2Packager::Product.selected_base
         extension_packages = Yast::Pkg.PkgQueryProvides(PROVIDES_KEY)
         log.info "module extension packages #{extension_packages.inspect}"
 
+        extension_packages.select! do |list|
+          pkg_name = list.first
+          dependencies = Yast::Pkg.ResolvableDependencies(pkg_name, :package, "").first["deps"]
```
jreidinger: do you measure how long it took? we can it twice here, so we should check if it does not slow it down too much.

src/lib/installation/clients/inst_modules_extensions.rb
```diff
@@ -21,11 +22,29 @@ def run
...
+
+          product_provides = dependencies.find_all do |d|
+            d["provides"] && d["provides"].match(/#{Regexp.escape(PROVIDES_PRODUCT)}/)
+          end
```
jreidinger: please log here pkg_name and product_provides. It will make debugging much easier.

src/lib/installation/clients/inst_modules_extensions.rb
gilsonsouza: I added a hash to avoid to execute it twice.

PR #635, opened 2017-12-11
Replace Remote module by the new y2remote/remote class

teclator:
- It depends on yast/yast-network#599


package/yast2-installation.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Mon Dec 11 15:50:04 UTC 2017 - knut.anderssen@suse.com
+
+- Replaced Remode module by the new y2remote/remoce class
```
imobachgs: Remode -> Remote and remoce -> remote.

PR #642, opened 2018-01-18
Merge "disable plymouth..."

imobachgs:
Supersedes https://github.com/yast/yast-installation/pull/584/. It is basically the same but fixing the changes file.


imobachgs:
Thanks!

PR #643, opened 2018-01-18
Disable plymouth by a no affend method, it will not block tty device …

imobachgs:
See https://github.com/yast/yast-installation/pull/584 for more details. This PR merges this code into the `SLE-12-SP3` branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ services:
 
 before_install:
   - docker build -t yast-installation-image .
+  # list the installed packages (just for easier debugging)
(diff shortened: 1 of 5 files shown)


imobachgs:
The Travis problem should be fixed by https://github.com/yast/yast-rake/pull/48.

imobachgs:
Thanks for the review! Finally we needed this change in `SLE-12-SP2`, so I backported it and there is a new PR (#648) which just merges those changes into `SLE-12-SP3`. So this PR is somehow superseded.

PR #644, opened 2018-01-19
save keyboard settings before console settings (bsc#1076798)

jsrain:


jsrain:
This has been agreed with Frederic as a temporary solution for Beta5. Next snapshots are still questionable :-(

PR #645, opened 2018-01-22
Disable plymouth by a no affend method, it will not block tty device …

imobachgs:
Port #643 to `SLE-12-SP2`.


imobachgs:
Thanks!

PR #646, opened 2018-01-22
Replace SetOverlay by SetSection

imobachgs:
[bsc#1077038](https://bugzilla.suse.com/show_bug.cgi?id=1077038)

The proposed fix have already been tested using a Kubic image.

### DIFF
--- a/src/lib/installation/system_role.rb
+++ b/src/lib/installation/system_role.rb
@@ -208,7 +208,7 @@ module Installation
       features = self.class.raw_roles.find { |r| r["id"] == id }.dup
 
       NON_OVERLAY_ATTRIBUTES.each { |a| features.delete(a) }
-      Yast::ProductFeatures.SetOverlay(features)
+      features.each { |s, v| Yast::ProductFeatures.SetSection(s, v) }
     end
   end
 end
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,9 +1,15 @@
 -------------------------------------------------------------------
+Mon Jan 22 13:00:36 UTC 2018 - igonzalezsosa@suse.com
+
+- Use a proper method to set roles features (bsc#1077038)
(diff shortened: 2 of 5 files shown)


mvidner:
This is triggered by this recent change: https://github.com/yast/yast-yast2/pull/671

src/lib/installation/system_role.rb
```diff
@@ -208,7 +208,7 @@ def overlay_features
       features = self.class.raw_roles.find { |r| r["id"] == id }.dup
 
       NON_OVERLAY_ATTRIBUTES.each { |a| features.delete(a) }
-      Yast::ProductFeatures.SetOverlay(features)
```
mvidner: Well, this is the only place (outside tests) where `SetOverlay` is called. I believe it does work for the other (non-Kubic) use cases.
Importantly, it aims to work even if you go Next and Back in the installation workflow.
Please describe what is different about Kubic that makes this change sensible.

imobachgs: Now that you have mentioned it, maybe the problem is that we simple won't call `ClearOverlay` in Kubic/CaaSP. If that's the problem, maybe we should fix it in a different way. What's your opinion?

mvidner: Oh, right, I forgot that Kubic/CaaSP actually uses the All-In-One dialog.
src/lib/y2caasp/widgets/system_role.rb calls role.overlay_features in the `store` method but this can be called several times. There should be a matching `ClearOverlay` before each.

imobachgs: I've added a call to ClearOverlay in the yast2-caasp module. Please, see https://github.com/yast/yast-caasp/pull/26.

mvidner:
OK, now that we have https://github.com/yast/yast-caasp/pull/26, I believe this one can be dropped :)

PR #647, opened 2018-01-22
use root_subvolume_read_only from libstorage-ng location (boo#1077047)

sysrich:
Should be merged after https://github.com/yast/yast-installation/pull/646/files (only because I assumed so in the .changes)


src/lib/installation/clients/umount_finish.rb
```diff
@@ -410,7 +410,7 @@ def set_root_subvol_read_only
     # should be set to read-only.
     #
     def root_subvol_read_only_configured?
-      ProductFeatures.GetBooleanFeature("partitioning", "root_subvolume_read_only")
+      ProductFeatures.GetBooleanFeature("proposal", "root_subvolume_read_only")
```
ancorgs: AFAIK, the value was moved from `/partitioning` to `/partitioning/proposal`, not to `/proposal` as this code seems to suggest. From the top of my mind (I don't use the ProductFeatures API so often), it looks to me like the correct fix would be:
```ruby
ProductFeatures.GetBooleanFeature("partitioning", "proposal", "root_subvolume_read_only")
```

package/yast2-installation.changes
```diff
@@ -1,3 +1,16 @@
...
+
+- use root_subvolume_read_only from libstorage-ng location
+  (boo#1077047)
+- 4.0.26
+
+-------------------------------------------------------------------
+Mon Jan 22 13:00:36 UTC 2018 - igonzalezsosa@suse.com
+
+- Use a proper method to set roles features (bsc#1077038)
+- 4.0.25
```
ancorgs: Why is this pull request getting this changelog entry in with no associated code change? I'm puzzled

src/lib/installation/clients/umount_finish.rb
sysrich: I couldnt' find any example of how it would work so took a stab in the dark

Your suggestion makes sense, will correct.

package/yast2-installation.changes
sysrich: Well it seemed weird putting version 4.0.26 in the .changes within the corresponding 4.0.25 - how else would you handle it with how you track version numbers in .changes?

PR #649, opened 2018-01-22
Moved root_subvolume_read_only (boo#1077047)

ancorgs:
Replaces #647

### DIFF
--- a/src/lib/installation/clients/umount_finish.rb
+++ b/src/lib/installation/clients/umount_finish.rb
@@ -416,9 +416,7 @@ module Yast
       partitioning = ProductFeatures.GetSection("partitioning")
       proposal = partitioning.nil? ? nil : partitioning["proposal"]
       return false if proposal.nil?
-
-      value = proposal["root_subvolume_read_only"]
-      value == true || value.respond_to?(:casecmp?) && value.casecmp?("yes")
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/umount_finish.rb
```diff
@@ -410,7 +410,15 @@ def set_root_subvol_read_only
...
-      ProductFeatures.GetBooleanFeature("partitioning", "root_subvolume_read_only")
+      # FIXME: this whole method should rely on Y2Storage::ProposalSettings.
+      # But right now there is no #root_subvolume_read_only method there
+
+      partitioning = ProductFeatures.GetSection("partitioning")
+      proposal = partitioning.nil? ? nil : partitioning["proposal"]
+      return false if proposal.nil?
+
+      value = proposal["root_subvolume_read_only"]
+      value == true || value.respond_to?(:casecmp?) && value.casecmp?("yes")
```
wfeldt: The value can only be `true` or `false`, enforced by the rng:

https://github.com/yast/yast-installation-control/blob/master/control/control.rnc#L516

ancorgs: I added the extra handling because `ProductFeatures.GetBooleanFeature` have it (probably historic reasons). But I will drop it and simplify this.

PR #650, opened 2018-01-26
Better mem

jreidinger:


src/lib/installation/clients/inst_prepareprogress.rb
```diff
@@ -167,6 +167,10 @@ def main
 
       Builtins.y2milestone("END of inst_prepareprogress.ycp")
 
+      # clean as much memory as possible before doing real installation, because some packages
+      # can have memory demanding scripts
+      GC.start
```
lslezak: It would be nice to log some numbers about used memory (vm, rss) before and after calling the garbage collector to see if it can help and how much.

jreidinger: you mean something from GC.stat or system wide?

lslezak: `GC.stat` seems to be too detailed and hard to understand, I'd prefer some information from the system, e.g.

```ruby
File.read("/proc/#{Process.pid}/status")
```

that should be enough.

kobliha:
Nice :) Thx for adding logging!

PR #654, opened 2018-01-31
Do not enable xvnc.socket for second stage installation

schubi2:
  * 88dafa12 do not enable xvnc.socket

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,9 +1,16 @@
 -------------------------------------------------------------------
+Wed Jan 31 13:09:51 CET 2018 - schubi@suse.de
+
+- Do not enable xvnc.socket. xvnc will be started by
+  calling it in vnc.sh directly. (bnc#1077236)
+- 4.0.28
+
+-------------------------------------------------------------------
 Tue Jan 30 09:30:26 UTC 2018 - lslezak@suse.cz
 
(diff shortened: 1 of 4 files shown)


package/yast2-installation.changes
```diff
@@ -1,9 +1,16 @@
...
+- Do not enable xvnc.socket for second installation stage.
+  Xvnc will be started by a direct call in vnc.sh. (bnc#1077236)
+- 4.0.28
+
 -------------------------------------------------------------------
 Tue Jan 30 09:30:26 UTC 2018 - lslezak@suse.cz
 
 - Reimplemented finding package upgrades to require less memory
   (related to bsc#1076768)
-- 4.0.33
```
teclator: Cool, from where came that version number ? :P

src/lib/installation/remote_finish_client.rb
```diff
@@ -76,7 +76,8 @@ def modes
 
     # Enables remote access
     def enable_remote
-      remote.enable!
```
teclator: If we don't have to enable the service anymore, then we could just call remote.write directly and remove this method.

package/yast2-installation.changes
schubi2: I think it is a typo. The last checked in package has really the version 4.0.27

PR #656, opened 2018-02-02
Fix empty role

jreidinger:


src/lib/installation/select_system_role.rb
```diff
@@ -86,6 +86,13 @@ def create_dialog
     def next_handler
       role_id = Yast::UI.QueryWidget(Id(:roles), :CurrentButton)
 
+      if role_id.nil? # no role selected (bsc#1078809)
+        # An Error popup
+        msg = _("No role is selected. Cannot continue until a role is selected.")
```
ancorgs: I would change it to a more straightforward (and, thus, less risky). "Please, select a role." or "Select one of the available roles to continue."

jreidinger: ok, will change

PR #657, opened 2018-02-06
Bringing back sidebar

hellcp:
For Leap 15.0 I wanted to bring back sidebar, with logo and steps on the left like this:

Sidebar :o
(And surprisingly this patch does that)
To my knowledge neither SUSE nor Kubic want it, so it would be nice to integrate it in a way that doesn't enable it by default (otherwise I will not hear the end of it from @sysrich about it being terrible idea and I shouldn't have started it).

Even though Kubic uses all its branding from Tumbleweed it makes no sense for single screen setup to use steps. It would be nice to disable them system level and not stylesheet level. I mean, it doesn't really matter to me, I could make second stylesheet without it, but it does to certain person that really likes Kubic.

With SUSE, they successfully use their theme that doesn't have any sort of sidebar and it works (and I would like to have that working as long as they wish, as they don't seem willing to change it for now).


ancorgs:
Making that configurable in control.xml shouldn't be a big deal. If the openSUSE artwork team really want this in, it should be doable.

@kobliha What do you think? @jreidinger, @mvidner, @lslezak as you guys probably know `Yast::Wizard` better than anyone else. Do you see any potential problem?

ancorgs:
@LelCP What about textmode installation? How does it look there?

LelCP:
From what I can tell, it looks exactly the same as it normally would, without sidebar
screenshot from 2018-02-06 05-15-06

jreidinger:
@ancorgs well this patch do what is intended, but as we use same source we should somehow protect it. It is sad that it cannot be tuned on stylesheet level, but control file level should be possible ( with reasonable default ).

@LelCP I like screenshot, it looks very well. It should be only for TW or also for Leap15?

jreidinger:
@ancorgs potential issue:

- maybe in some screen it does not fit ( but as ncurses does not support side bar for this exact reason, I think not so big problem ).
- possible glitches for custom role that have additional dialog
- not tested for long time, so potential bugs can be discovered.

sysrich:
Reminder - it will need to work and look well at low resolutions and lousy
aspect ratios, such as 800.600 and 1024.768

Especially as that is what openQA mostly tests

LelCP:
@jreidinger There are two or three bugs I discovered.
- Sometimes background for sidebar dissapears, it happens mostly on screens where stylesheet editing is disabled (pattern selection, date and time), on those where stylesheet editing is enabled applying theme again fixes it (Repositories Initialization), sounds easy to fix from my very limited knowledge
- Whatever is happening on that screenshot with optional steps being shown and current step arrow being on multiple steps that are ahead (it's after going back multiple steps so it's hard to say if it's intended behavior)
- Release Notes button shows up on sidebar only during User Settings (it stays there is you go back from users settings though)

@sysrich Only thing that doesn't work in 800x600 is time showing up properly in date and time 
what's the time

but there are bigger problems with that screen atm ;)

it's bugs time

the rest is clear and readable

mvidner:
@LelCP Yay, a new look! The YaST packaging pipeline requires a Bugzilla or FATE ID in the .changes file (a GitHub item alone will not do). Do you have one?

LelCP:
@mvidner can create either, don't have one yet

teclator:
@LelCP what is the state of this PR? there is still no fate id or bugzilla which can be used for it?

LelCP:
Good thing you reminded me of it, boo#1088785

kobliha:
It's nice to have this kind of contribution, but I'm afraid that we can't merge it now. It puts other products into danger. The "only" one: time-zone is probably not the only one. We have quite some hidden dialogs that appear on some architectures only or are related to some hardware.

Even introducing a control file entry might break something. The code is shared and "just adding one more if" is usually more than that - its often just the start where things start to break. We also don't have a feature request so nobody was spending their time thinking about a safe solution first.

I as a YaST Team TL and PO do not recommend merging it now - it seems to be really too late.

lslezak:
- We would need a `control.xml` option for this (disabled by default if not set)
- I guess this would require quite a lot of debugging and testing. The number of steps might depend on the product (e.g. there is a registration step in SLE), the environment (`with_iscsi=1`) or architecture (there are some extra steps on S390).
- I'm afraid this can break openQA a lot
- There is an issue already in the initial screenshot, the "Add-On Products" step is for some reason marked as done although only the very first dialog is displayed.

So I think now it's too late to merge it into SLE15/Leap15, we could consider it for SLE15-SP1/Leap15.1.

LelCP:
> As someone who supported @LelCP as he went down this crazy route, I reluctantly agree. IIRC I did let LCP know this was always a risk..so hopefully he wont be too upset about the delay :-/

I expected it since day 1 of this :D I don't see why I would be upset, it's a feature, for some of them we need to wait

src/lib/installation/clients/installation.rb
```diff
@@ -41,15 +41,22 @@ def main
       Yast.import "Hooks"
       Yast.import "Linuxrc"
       Yast.import "OSRelease"
+      Yast.import "ProductFeatures"
+      Yast.import "ProductControl"
```
teclator: Is it really needed?

hellcp: Apparently, yes, you could test it with and without, it didn't work for me without adding those imports though.

PR #659, opened 2018-02-08
Remove xinetd from startup scripts

teclator:
- [Trello Card](https://trello.com/c/TXqZgQhM/1325-5-remove-yast-xinetd-agent-inetd-sockets)

### DIFF
--- a/startup/YaST2.call
+++ b/startup/YaST2.call
@@ -165,24 +165,8 @@ function prepare_for_vnc () {
 	if [ $VNCPASS_EXCEPTION = 0 ];then
 		disable_splash
 		displayVNCMessage
-		stop_xinetd
 		startVNCServer
 		wait_for_x11
-		if is_xinetd_active; then
-			# xinetd shouldn't be running since we stopped it right
-			# before starting VNC. But there is still one small
-			# chance that xinetd was started by the systemd boot
-			# sequence during VNC startup. If it's running we cannot
-			# be sure that our VNC got the ports, so retry.
-			#
-			# Using systemd dependencies to ensure the correct order
-			# looks nicer, but is fragile (bnc#931643, bnc#947521).
-			killall Xvnc >/dev/null 2>&1
-			stop_xinetd
-			startVNCServer
-			wait_for_x11
-		fi
-		restore_xinetd
 		if [ "$server_running" = 1 ];then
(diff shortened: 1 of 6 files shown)


startup/YaST2.call
```diff
@@ -165,24 +165,8 @@ function prepare_for_vnc () {
 	if [ $VNCPASS_EXCEPTION = 0 ];then
 		disable_splash
 		displayVNCMessage
-		stop_xinetd
```
teclator: I was discussing with @jreidinger and with @schubi2 that I'm not sure if we should do the same with sockets as we did with xinetd.

https://github.com/yast/yast-installation/pull/654 (https://bugzilla.suse.com/show_bug.cgi?id=1077236)

After last changes we are not enabling the socket at the end of the first stage so should not be needed by now .

kobliha: In general, VNC should be active if it's requested by the customer

1. In AutoYaST, there can be a demand for the service/socket running
2. During manual installation, installation via VNC also offers to open port in FW, IMO handled this in the past as if the customer wanted to enable VNC service too

In other words, we should keep the old behavior as much as possible. Just convert it to new schema xinetd --> socket, systemd --> service

teclator: And about the current implementation:

Basically the client responsible of write the remote configuration during the first stage (manual installation , autoyast) is the remote_finish client.

It is only called if the Linuxrc.vnc option is given 
```ruby
    def modes
      Yast::Linuxrc.vnc ? [:installation, :autoinst] : []
    end
```

So from my POV, the #654 is dropping that behavior which previously was not affecting the Second Stage because we were stopping xinetd, launching the vncserver and then restoring xinetd during the cleanup.

So from my POV we should continue enabling the socket in the `remote_finish_client` and then do the same we were doing with xinetd but with the xvnc.socket.

@schubi2 @kobliha

teclator: BTW I will try to check the result of installations and auto-installations with SLE-12-SP3 and SLE-15 and just compare results.

teclator: Related: 

https://bugzilla.suse.com/show_bug.cgi?id=893501
https://github.com/yast/yast-installation/commit/f0ecde998c68a79e04cc60a957563f4b25198c26

jreidinger:
replaced by https://github.com/yast/yast-installation/pull/661

PR #660, opened 2018-02-12
Replace xinetd by xvnc.socket in startup scripts

teclator:
Alternative PR see (#659), in this PR we bring back the `xvnc.socket` enablement during the first stage when the linuxrc `vnc` option is given, the startup scripts have been adapted to stop the socket and restore it as we were doing with xinetd not supposing a problem during the `Second Stage` anymore

## Bugs related: ##

https://bugzilla.suse.com/show_bug.cgi?id=893501 (bug where the VNC service was originilally enabled during the installation)

https://bugzilla.suse.com/show_bug.cgi?id=1077236 (where it was removed due to a problem trying to launch the VNC Server during the Second Stage)

https://bugzilla.suse.com/show_bug.cgi?id=1078785 (To take in account, there is no `java client` anymore, so we should modify the current text (`startup/common/vnc.sh`) and adapt the script to serve the novnc javascript client)

### DIFF
--- a/startup/common/misc.sh
+++ b/startup/common/misc.sh
@@ -186,39 +186,31 @@ function disable_splash () {
 	[ -f /proc/splash ] && echo "verbose" > /proc/splash
 }
 
-#----[ stop_xinetd ]-----#
-function stop_xinetd () {
+#----[ stop_xvnc]-----#
+function stop_xvnc () {
 #--------------------------------------------------
-# stop xinetd since its default configuration collides
+# stop xvnc since its default configuration collides
 # with the Xvnc server used for VNC installation
 # ---
-	systemctl stop xinetd.service >/dev/null 2>&1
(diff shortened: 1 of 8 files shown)


startup/common/misc.sh
```diff
@@ -186,39 +186,31 @@ function disable_splash () {
...
-function stop_xinetd () {
+#----[ stop_xvnc]-----#
+function stop_xvnc () {
 #--------------------------------------------------
-# stop xinetd since its default configuration collides
+# stop xvnc since its default configuration collides
 # with the Xvnc server used for VNC installation
 # ---
-	systemctl stop xinetd.service >/dev/null 2>&1
+	systemctl stop xvnc.socket >/dev/null 2>&1
```
jreidinger: this is not enough, as stopping socket does not stop service that is started with socket

PR #661, opened 2018-02-12
Vnc

jreidinger:


schubi2:
Are you sure, that we will not break the second stage of AY again ?
https://github.com/yast/yast-installation/pull/654/files

jreidinger:
@schubi2 well, we stop xvnc in startup script to allow start of our own xvnc, so it should not break second stage. I expect if we break it then openqa catch it, but we try to handle that case by explicit stop of xvnc service, so port should be free.

PR #662, opened 2018-02-13
Fix set overlay

jreidinger:


package/yast2-installation.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Feb 13 12:53:44 UTC 2018 - jreidinger@suse.com
+
+- Ensure previous role is clear when doing desktop selection
```
lslezak: I'd sugest "previous role **selection** is clear**ed**".

PR #665, opened 2018-02-27
Do not block autoinstallation when no disks/controllers are found

imobachgs:
Fixes [bsc#1082854](https://bugzilla.suse.com/show_bug.cgi?id=1082854#c3).

### DIFF
--- a/test/inst_system_analysis_test.rb
+++ b/test/inst_system_analysis_test.rb
@@ -104,7 +104,7 @@ describe Yast::InstSystemAnalysisClient do
           end
         end
 
-        context "during autoinstallation" do
+        context "during normal installation" do
           let(:auto) { false }
 
           it "reports an error" do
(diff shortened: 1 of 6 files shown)


test/inst_system_analysis_test.rb
```diff
@@ -34,27 +38,121 @@
...
+        context "during autoinstallation" do
+          let(:auto) { true }
+
+          it "reports a warning" do
+            expect(Yast::Report).to receive(:Warning)
+            client.ActionHDDProbe
+          end
+        end
+
+        context "during autoinstallation" do
```
teclator: The context seems to be wrong

test/inst_system_analysis_test.rb
```diff
@@ -34,27 +38,121 @@
...
```
teclator: Same here

test/inst_system_analysis_test.rb
imobachgs: Oh, `CopyAndPasteError`.

PR #668, opened 2018-03-12
Read-only volumes

imobachgs:
Set 'ro' property for every Btrfs filesystem which is mounted as read-only at the end of the installation.

### DIFF
--- a/test/lib/clients/umount_finish_test.rb
+++ b/test/lib/clients/umount_finish_test.rb
@@ -56,7 +56,7 @@ describe Yast::UmountFinishClient do
       end
     end
 
-    context "when no Btrfs fileystem is mounted as read-only" do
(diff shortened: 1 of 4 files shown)


test/lib/clients/umount_finish_test.rb
```diff
@@ -3,98 +3,66 @@
...
-    # Validation should protect us from this, but is not always checked
-    context "if root_subvolume_read_only has a non boolean value in /partitioning/proposal section" do
-      let(:features) do
-        {
-          "partitioning" => {
-            "proposal" => { "root_subvolume_read_only" => "not so sure" }
-          }
-        }
-      end
+    context "when no Btrfs fileystem is mounted as read-only" do
```
joseivanlopez: NP: the position of "no" was a bit confusing for me. Why not "When Btrfs filesystem is no mounted as read-only".

PR #669, opened 2018-03-13
[Review] Request from 'schubi2' @ 'yast/yast-installation/review_180313_translage_licenses'

schubi2:
  * 9072b59d packaging
  * d8400121 rebuild translation widget when the language has been changed

### DIFF
--- a/src/lib/installation/dialogs/complex_welcome.rb
+++ b/src/lib/installation/dialogs/complex_welcome.rb
@@ -63,7 +63,7 @@ module Installation
       #
       # @return [Yast::Term] Dialog's content
       def contents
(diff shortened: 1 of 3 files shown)


src/lib/installation/dialogs/complex_welcome.rb
```diff
@@ -63,7 +63,7 @@ def title
       #
       # @return [Yast::Term] Dialog's content
       def contents
-        @contents ||= VBox(
+        @contents = VBox(
```
imobachgs: Why to use a `@contents` instance variable? Why not just returning the value of VBox (I mean, removing the assignment).

PR #670, opened 2018-03-13
Handle corner cases when setting Btrfs 'ro'

imobachgs:


src/lib/installation/clients/umount_finish.rb
```diff
@@ -421,11 +421,13 @@ def default_subvolume_as_ro(path)
       )
 
       default_subvolume = output.strip.split.last
-      spec = Y2Storage::VolumeSpecification.for(path)
-      if spec.btrfs_default_subvolume
-        prefix = Pathname.new(spec.btrfs_default_subvolume)
-        default_subvolume = Pathname.new(default_subvolume).relative_path_from(prefix).to_s
+      # no btrfs_default_subvolume and no snapshots
+      default_subvolume = "" if default_subvolume == "(FS_TREE)"
```
joseivanlopez: NP: I would use a specific const for that "(FS_TREE)" and also I would use the const in the unit tests.

test/lib/clients/umount_finish_test.rb
```diff
@@ -25,24 +23,92 @@
     let(:root_fs) do
       instance_double(
         Y2Storage::Filesystems::Btrfs,
-        is?:           true,
-        mount_point:   mount_point,
-        mount_options: mount_options
+        is?:               true,
```
joseivanlopez: I am worried about the amount of mocking here. Maybe we need to start to think about how to use some storage-ng test helpers in other packages. In these tests it would be nice to have `fake_scenario(file)`, for example. For the time being, it LGTM.

imobachgs: Yes, I though the same. Actually I would be happier if the call to `btrfs_subvolume_mount_point` wan't mocked too. IMHO, we need to move those helpers to the library so you can do something like `require "y2storage/rspec"` and use the helpers.

PR #671, opened 2018-03-16
Enable many, many roles (with a RichText widget)

mvidner:
- https://trello.com/c/WmjvFAZP
- [bsc#1084674](https://bugzilla.suse.com/show_bug.cgi?id=1084674)

Quick testing:
```sh
rake run[inst_system_role]
```
or
```sh
DISPLAY= rake run[inst_system_role]
```

six-roles-gui

six-roles-tui

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Mar 20 16:10:24 UTC 2018 - mvidner@suse.com
+
+- Make many system roles fit in the dialog (using a RichText
+  widget, bsc#1084674)
+- 4.0.41
+
+-------------------------------------------------------------------
 Mon Mar 19 22:53:57 UTC 2018 - knut.anderssen@suse.com
 
 - Self-Update report errors when the self_update is enabled
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -72,19 +72,20 @@ module Installation
     end
 
     def dialog_content
-      HSquash(VBox(role_buttons))
+      @selected_role_id = self.class.original_role_id || (roles.first && roles.first.id)
+
+      HSquash(ReplacePoint(Id(:rp), role_buttons(selected_role_id: @selected_role_id)))
     end
 
     def create_dialog
       clear_role
       ok = super
-      role_id = self.class.original_role_id || (roles.first && roles.first.id)
-      Yast::UI.ChangeWidget(Id(:roles), :CurrentButton, role_id)
+      Yast::UI.SetFocus(Id(:roles_richtext)) if ok
       ok
     end
 
     def next_handler
-      role_id = Yast::UI.QueryWidget(Id(:roles), :CurrentButton)
+      role_id = @selected_role_id
 
       if role_id.nil? # no role selected (bsc#1078809)
         # An Error popup
@@ -114,6 +115,19 @@ module Installation
       else
         finish_dialog(result)
       end
+    end
+
+    # called if a specific FOO_handler is not defined
+    def handle_event(id)
+      role = SystemRole.find(id)
+      if role.nil?
+        log.info "Not a role: #{id.inspect}, skipping"
+        return
+      end
+
+      @selected_role_id = id
+      Yast::UI.ReplaceWidget(Id(:rp), role_buttons(selected_role_id: id))
+      Yast::UI.SetFocus(Id(:roles_richtext))
     end
 
   private
@@ -170,15 +184,6 @@ module Installation
       Yast::ProductFeatures.ClearOverlay
     end
 
-    # Returns the content for the role buttons
-    #
-    # @return [Yast::Term] Role buttons
-    #
-    # @see role_buttons_options
-    def role_buttons
-      role_buttons_content(role_buttons_options)
-    end
-
     # Applies given role to configuration
     def apply_role(role)
       log.info "Applying system role '#{role.id}'"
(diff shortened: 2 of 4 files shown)


test/select_system_role_test.rb
```diff
@@ -158,116 +148,6 @@
           expect(subject.run).to eq(:back)
         end
       end
-
-      context "when no roles is selected" do
```
mvidner: In the new code a role is always selected. I have also tested manually with the scenario described in the original repor, https://bugzilla.suse.com/show_bug.cgi?id=1078809

package/yast2-installation.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Mar 16 14:09:09 UTC 2018 - mvidner@suse.com
+
+- Make many system roles fit in the dialog (using a RichText widget,
```
jreidinger: comma exceed line lenght

src/lib/installation/select_system_role.rb
```diff
@@ -72,19 +72,20 @@ def help_text
     end
 
     def dialog_content
-      HSquash(VBox(role_buttons))
+      @selected_role_id = self.class.original_role_id || (roles.first && roles.first.id)
```
jreidinger: I think deserve to be moved to system_role class itself. Like `selected` https://github.com/yast/yast-installation/blob/master/src/lib/installation/system_role.rb#L97 so you do not forget this no_default feature.
And maybe a bit renaming so "current" will return what should be selected and "selected" will return what user really select.

src/lib/installation/select_system_role.rb
```diff
@@ -116,6 +117,16 @@ def next_handler
       end
     end
 
+    # called if a specific FOO_handler is not defined
+    def handle_event(id)
+      role = SystemRole.find(id)
+      return if role.nil?
```
jreidinger: logging please, so we know what is happen when event is ignored.

src/lib/installation/select_system_role.rb
```diff
@@ -116,6 +117,16 @@ def next_handler
...
+
+      @selected_role_id = id
+      Yast::UI.ReplaceWidget(Id(:rp), role_buttons(selected_role_id: id))
+      Yast::UI.SetFocus(Id(:rb))
```
jreidinger: is it type that `:rb` or you really have two widget ids rp and rb? I really dislike it. Please no accronym when you use it in quite distant places. ( or constant ideally, so tests can catch it ).

src/lib/installation/select_system_role.rb
```diff
@@ -210,114 +212,59 @@ def roles(refresh: false)
       SystemRole.all
     end
 
-    # Returns the content for the role buttons according to the available space
+    # Returns the content for the role buttons
     #
-    # @param separation  [Integer] Separation between roles
-    # @param indentation [Integer] Roles indentation
-    # @param margin      [Integer] Top marging
-    # @param description [Boolean] Indicates whether the description should be included
```
jreidinger: missing documentation for param

src/lib/installation/select_system_role.rb
```diff
@@ -210,114 +212,59 @@ def roles(refresh: false)
...
     # @return [Yast::Term] Role buttons
-    def role_buttons_content(margin: 2, indentation: 4, separation: 2, description: true, show_intro: true)
+    def role_buttons(selected_role_id: nil)
```
jreidinger: does it make sense to have default? I think it is always called with variable.

src/lib/installation/select_system_role.rb
```diff
@@ -210,114 +212,59 @@ def roles(refresh: false)
...
       box = []
-      box << Left(Label(Yast::ProductControl.GetTranslatedText("roles_text"))) if show_intro
-      box << VSpacing(margin) unless margin.zero?
+      box << Left(Label(Yast::ProductControl.GetTranslatedText("roles_text")))
```
jreidinger: why not `box = [Left(Label(Yast::ProductControl.GetTranslatedText("roles_text")))]`

src/lib/installation/select_system_role.rb
```diff
@@ -210,114 +212,59 @@ def roles(refresh: false)
...
-    # Title + Descriptions
-    #
-    # @return [Integer] Number of lines needed to display roles information
-    def needed_lines_for_roles
-      return @needed_lines_for_roles if @needed_lines_for_roles
-      texts = roles.map(&:description)
-      texts << intro_text unless intro_text.nil?
-      lines_qty = texts.compact.map(&:lines).map(&:size).reduce(:+)
-      @needed_lines_for_roles = roles.size + lines_qty
+    BUTTON_ON = "◉".freeze # U+25C9 Fisheye
```
lslezak: I'd rather use images instead of the Unicode symbols so the rich text radio buttons look exatly same as normal radio buttons:

 ([inst_radio-button-checked.png](https://github.com/yast/yast-theme/blob/master/SLE/wizard/inst_radio-button-checked.png))
 ([inst_radio-button-unchecked.png](https://github.com/yast/yast-theme/blob/master/SLE/wizard/inst_radio-button-unchecked.png))

It seems that there are no icons for the normal mode (see https://github.com/yast/yast-theme/tree/master/SLE/wizard) but that should not be a problem as this dialog is displayed only during installation. (We could use that Unicode trick as a fallback in normal mode.)

See the [image implementation](https://github.com/yast/yast-registration/blob/cc13812295556b2ec462fc68865d4fca287e1403/src/lib/registration/ui/addon_selection_base_dialog.rb#L181) in the registration dialog.

src/lib/installation/select_system_role.rb
```diff
@@ -210,114 +212,59 @@ def roles(refresh: false)
...
-      description: true
-    }.freeze
-
-    def intro_text
-      @intro_text ||= Yast::ProductControl.GetTranslatedText("roles_text")
+        rb = richtext_radiobutton(id:       role.id,
+                                  label:    role.label,
+                                  selected: role.id == selected_role_id)
+        description = description.lines.map { |l| l + "<br>" }.join "\n"
+        "<p></p>#{rb}<p></p><ul>#{description}</ul>"
```
jreidinger: why this empty paragraph? simple <br> does not look good? at least please comment it in code.

src/lib/installation/select_system_role.rb
```diff
@@ -210,114 +212,59 @@ def roles(refresh: false)
...
-      show_intro:  true,
-      description: true
-    }.freeze
-
-    def intro_text
-      @intro_text ||= Yast::ProductControl.GetTranslatedText("roles_text")
+        rb = richtext_radiobutton(id:       role.id,
+                                  label:    role.label,
+                                  selected: role.id == selected_role_id)
+        description = description.lines.map { |l| l + "<br>" }.join "\n"
```
jreidinger: this is a bit magic to understand? why not use simple `description.gsub("\n", "<br>\n")` or reason is to cut the final newline?

jreidinger: btw also do not forget that lines keep ending line, so result is in fact `\n<br>\n` which I think you do not want.

test/select_system_role_test.rb
mvidner: No, the code does not use that. Oh great, now I found the code which uses that, on openSUSE, and I have to adapt it too I guess.

mvidner:
Even after code-review fixes the manual test does pass, wow.

For the record, for now I am sticking with the Unicode radio buttons which look different than the images used elsewhere in the UI.

PR #672, opened 2018-03-19
Extend current self_update_id doc.

teclator:


doc/SELF_UPDATE.md
```diff
@@ -111,9 +111,15 @@ The URL of the update repository is evaluated in this order:
         required just like in manual installation.
    4. Default SUSE Customer Center API (`https://scc.suse.com/`).
 
-   Registration server is then asked for URL providing product on media.
-   For multi-product media it is possible to specify ID used instead of product
-   name in control.xml as globals->self_update_id.
+   The registration server is then asked for the list of update repositories.
+
+   For being able to determine the list, the registration server expect
```
imobachgs: What about "In order to determine" ?

imobachgs: I would dot not use a blank line and, instead, I would say something like: *In order to determine such a list, the product `name`, `version` and `arch` are used.*

doc/SELF_UPDATE.md
```diff
@@ -111,9 +111,15 @@ The URL of the update repository is evaluated in this order:
...
+   the product `name`, `version` and `arch` of the installer being updated.
+   **Ex:** `product: SLED`, `version: 15`, `arch: x86_64`
```
jreidinger: I prefer to not use accronyms, it is just confusing.

doc/SELF_UPDATE.md
```diff
@@ -111,9 +111,15 @@ The URL of the update repository is evaluated in this order:
...
+
+   In case of multi-product media, as the installer is the same for all the
```
imobachgs: what about *a multi-product installation medium*?

doc/SELF_UPDATE.md
```diff
@@ -111,9 +111,15 @@ The URL of the update repository is evaluated in this order:
...
+   products on the media, the product `name` can be hard-coded in the control
```
imobachgs: *for all the included products* (and you can get rid of *media*).

doc/SELF_UPDATE.md
```diff
@@ -111,9 +111,15 @@ The URL of the update repository is evaluated in this order:
...
+   file (control.xml) as `globals -> self_update_id`.
```
imobachgs: I would keep consistency and use "xpath" style: `/globals/self_update_id` (as we do for `/general/self-update-url`).

PR #673, opened 2018-03-19
Check if self_update is explicitly enabled

teclator:
- https://trello.com/c/syNlrXoP/1371-3-p1-1084820-handle-selfupdate-during-autoupgrade-correctly

With this PR we will show also errors getting the list of update repositories when self_update has been explicitly enabled.

Test units are missing, and if the approach is OK will also update the documentation of errors handling.

## Manual tests

For the manual tests a Report has been added but it is not part of the PR.

### No enabled explicitly

notshownerrors

### Enable with linuxrc `self_update=1`

errorsshown

### Enable in the AY profile `/general/self_update`

errorsshownay

### DIFF
--- a/src/modules/InstFunctions.rb
+++ b/src/modules/InstFunctions.rb
@@ -40,6 +40,7 @@ module Yast
       Yast.import "Mode"
       Yast.import "ProductControl"
       Yast.import "Profile"
+      Yast.import "SCR"
     end
 
     # Returns list of ignored features defined via Linuxrc commandline
@@ -141,15 +142,14 @@ module Yast
       # Linuxrc always export SelfUpdate with the default value even if not has
       # been set by the user. For that reason we need to check the cmdline for
       # knowing whether the user has requested the self update explicitly.
-      in_cmdline = Linuxrc.value_for("self_update")
-      if in_cmdline && in_cmdline != "0"
+      if self_update_in_cmdline?
         log.info("Self update was enabled explicitly by linuxrc cmdline")
         return true
       end
 
       return false unless Mode.auto
 
-      profile = Yast::Profile.current
+      profile = Profile.current
       in_profile = profile.fetch("general", {}).fetch("self_update", false)
       log.info("Self update was enabled explicitly by the AY profile") if in_profile
       in_profile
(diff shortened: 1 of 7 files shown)


src/modules/InstFunctions.rb
```diff
@@ -131,6 +133,28 @@ def second_stage_required?
...
     alias_method :second_stage_required, :second_stage_required?
 
+    # Determine whether the installer update (self_update=1) has been
+    # explicitly enabled by linuxrc or by the AY profile. It does not check if
+    # a custom url has been provided or not.
+    #
+    # return [Boolean] true if enabled explicitly; false otherwise
+    def self_update_explicitly_enabled?
+      cmdline = polish(Linuxrc.InstallInf("Cmdline") || "").split
+      cmdline_requested = cmdline.include?("selfupdate=1")
```
imobachgs: np: I would rename `cmdline_requested` to `requested_in_cmdline` or just `in_cmdline`. Same for `profile_requested`.

src/modules/InstFunctions.rb
```diff
@@ -131,6 +133,28 @@ def second_stage_required?
     end
     alias_method :second_stage_required, :second_stage_required?
 
+    # Determine whether the installer update (self_update=1) has been
+    # explicitly enabled by linuxrc or by the AY profile. It does not check if
+    # a custom url has been provided or not.
+    #
+    # return [Boolean] true if enabled explicitly; false otherwise
+    def self_update_explicitly_enabled?
+      cmdline = polish(Linuxrc.InstallInf("Cmdline") || "").split
```
imobachgs: I would comment why are we using `Cmdline` instead of just `SelfUpdate`...

teclator:
@imobachgs I had to modify the read of the cmdline as also `Linuxrc.value_for` relies in the `InstallInf`so had to read the `/proc/cmdline` explicitly

src/modules/InstFunctions.rb
```diff
@@ -161,6 +161,15 @@ def self_update_explicitly_enabled?
 
   private
 
+    # Return whether self_update has been explicitly enabled by linuxrc or not
+    #
+    # return [Boolean] true if enabled; false otherwise
+    def self_update_in_cmdline?
+      cmdline = polish(SCR.Read(path(".target.string"), "/proc/cmdline").to_s)
+
+      !!cmdline.split.find { |s| s.start_with?("selfupdate=") && s != "selfupdate=0" }
```
imobachgs: What about using `any?`? It should improve readability.

PR #674, opened 2018-03-22
bsc#1078785: Start webvnc during installation process

gilsonsouza:


mvidner:
I have superseded this in #677.

PR #675, opened 2018-03-23
System roles: make the radio buttons match other dialogs, using images

mvidner:
A fixup of #671
 https://trello.com/c/WmjvFAZP/1370-5-system-roles-dialog-with-8-roles

six-roles-gui-rb2
six-roles-tui-rb2

### DIFF
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -76,7 +76,7 @@ module Installation
       @selected_role_id = self.class.original_role_id
       @selected_role_id ||= roles.first && roles.first.id if SystemRole.default?
 
-      HSquash(ReplacePoint(Id(:rp), role_buttons(selected_role_id: @selected_role_id)))
+      HCenter(ReplacePoint(Id(:rp), role_buttons(selected_role_id: @selected_role_id)))
     end
 
     def create_dialog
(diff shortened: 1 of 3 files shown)


src/lib/installation/select_system_role.rb
```diff
@@ -257,14 +257,21 @@ def richtext_radiobutton_tui(id:, label:, selected:)
...
-      widget = "#{check} #{label}"
       # check for installation style, which is dark, FIXME: find better way
       installation = ENV["Y2STYLE"] == "installation.qss"
+      if installation
+        image = selected ? "inst_radio-button-checked.png" : "inst_radio-button-unchecked.png"
+        bullet = "<img src=\"#{IMAGE_DIR}/#{image}\"></img>"
+      else
+        bullet = selected ? BUTTON_ON : BUTTON_OFF
+      end
+      widget = "#{bullet} #{label}"
```
lslezak: I have overlooked that earlier, maybe we should escape the `label` value and others. This is not a web page so no JS inject attack possible, but the RichText might break if somebody uses `<`, `>` or `&` in the description or label.

So using `CGI.escapeHTML` here would be a good idea. See e.g. https://github.com/yast/yast-registration/blob/540b80931575e6b436b7e4bc7a426530e20cda12/src/clients/scc.rb#L81

lslezak: Or do we allow HTML content in the roles labels and descriptions?

PR #677, opened 2018-03-28
Start webvnc during the installation

mvidner:
- Supersedes #674 
- https://trello.com/c/rkYtkFT1
- [bsc#1078785](https://bugzilla.suse.com/show_bug.cgi?id=1078785)

### DIFF
--- a/startup/common/vnc.sh
+++ b/startup/common/vnc.sh
@@ -17,6 +17,8 @@
 #----------------
 
 . /etc/YaST2/XVersion
+# set python path for websockify
+. /root/.profile
 
 #----[ setupVNCAuthentication ]------#
 setupVNCAuthentication () {
@@ -92,9 +94,6 @@ startVNCServer () {
 	>/var/log/YaST2/vncserver.log 2>&1 &
 	xserver_pid=$!
 
-	# Python gets confused by the symlink games played in inst-sys
-	# so we must help it
-	export PYTHONPATH=`echo /usr/lib/python3*`/site-packages:$PYTHONPATH
 	websockify \
 		--web /usr/share/novnc \
 		5801 \
--- a/startup/common/vnc.sh
+++ b/startup/common/vnc.sh
@@ -65,6 +65,11 @@ displayVNCMessage () {
 	echo
 }
 
+websockify() {
+	# this shortcut avoids requiring the optional numpy dependency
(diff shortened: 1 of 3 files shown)


mvidner:
Testing this is cumbersome. For a start, https://build.suse.de/request/show/160504 has not landed in an ISO yet. So, quoting Gilson's instructions:

> $ osc -A https://api.suse.de getbinaries home:snwint:head/installation-images sle15 x86_64 -M SLES
> So, we can create an image by unpacking skelcd-installer-SLES.rpm, let's say in /tmp/foo, and then running:
> $ mksusecd -c new.iso sles15.iso /tmp/foo/usr/lib/skelcd/CD1

gets you an installation ISO, then supply a RPM built out of this PR as a DUD.

Anyway, I have tested this manually.

startup/common/vnc.sh
```diff
@@ -82,12 +87,20 @@ startVNCServer () {
...
-		-httpd /usr/share/vnc/classes \
 		-rfbport 5901 \
-		-httpport 5801 \
 		-fp $Xfontdir/misc/,$Xfontdir/uni/,$Xfontdir/truetype/ \
 	>/var/log/YaST2/vncserver.log 2>&1 &
 	xserver_pid=$!
+
+	# Python gets confused by the symlink games played in inst-sys
+	# so we must help it
+	export PYTHONPATH=`echo /usr/lib/python3*`/site-packages:$PYTHONPATH
```
lslezak: Maybe we should do this trick always, even in the standard installation, not only in VNC installation. Although we do use any python tools yet in inst-sys this could possibly avoid the same issue elsewhere.

mvidner: Good point. Actually, PYTHONPATH **is** already patched similarly in the inst_sys `/root/.profile` which confused the hell out of me: YaST did not see that but the shell spawned by Ctrl-Alt-Shift-X did see it.

lslezak: Hmm, then we probably could simply source that profile file in the YaST starting scripts...

mvidner:
@lslezak updated (and retested)

startup/common/vnc.sh
```diff
@@ -17,6 +17,7 @@
 #----------------
 
 . /etc/YaST2/XVersion
+. /root/.profile
```
lslezak: I think a short comment *why* would make sense here...

PR #681, opened 2018-04-10
bnc 1086846 add-on product selecting again for installation

schubi2:


test/widgets/product_selector_test.rb
```diff
@@ -57,6 +57,8 @@
     before do
       allow(Yast::Pkg).to receive(:PkgApplReset)
       allow(Yast::Pkg).to receive(:PkgReset)
+      allow(Yast::AddOnProduct).to receive(:selected_installation_products)
```
imobachgs: This test only prevents the testsuite from failing, but it does not really test the products installation (the call to `Pkg.ResolvableInstall`), right? It would be nice to test it.

PR #685, opened 2018-04-19
Log more details when unmounting fails (to debug bsc#1090018)

lslezak:
- Unmounting `/mnt` fails with "umount: /mnt: target is busy" error
- See https://bugzilla.suse.com/show_bug.cgi?id=1090018
- I have no idea why it fails, likely some `%post` script started a process or maybe YaST keeps some file opened
- To help with debugging log the `fuser` output to see which process use that mount point
- 4.0.50

### DIFF
--- a/src/lib/installation/clients/umount_finish.rb
+++ b/src/lib/installation/clients/umount_finish.rb
@@ -159,13 +159,14 @@ module Yast
             WFM.Execute(path(".local.umount"), umount_this)
           )
           if umount_result != true
-            begin
-              # the details are printed on STDERR, redirect it
-              fuser = `LC_ALL=C fuser -v -m #{Shellwords.escape(umount_this)} 2>&1`
-              log.warn("Running processes using #{umount_this}: #{fuser}")
-            rescue Errno::ENOENT
-              log.warn("Calling fuser failed, probably not installed")
-            end
+            # run "fuser" to get the details about open files
+            # (the details are printed on STDERR, redirect it)
+            fuser = begin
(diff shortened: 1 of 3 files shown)


src/lib/installation/clients/umount_finish.rb
```diff
@@ -159,9 +159,13 @@ def main
             WFM.Execute(path(".local.umount"), umount_this)
           )
           if umount_result != true
-            # the details are printed on STDERR, redirect it
-            fuser = `LC_ALL=C fuser -v -m #{Shellwords.escape(umount_this)} 2>&1`
-            log.warn("Running processes using #{umount_this}: #{fuser}")
+            begin
+              # the details are printed on STDERR, redirect it
+              fuser = `LC_ALL=C fuser -v -m #{Shellwords.escape(umount_this)} 2>&1`
```
jreidinger:
```
  `non-exist` rescue "not installed"
```

jreidinger:
```
  `non-exist` rescue "failed: #{$!}"
```

src/lib/installation/clients/umount_finish.rb
```diff
@@ -158,6 +159,14 @@ def main
             WFM.Execute(path(".local.umount"), umount_this)
           )
           if umount_result != true
+            # run "fuser" to get the details about open files
+            # (the details are printed on STDERR, redirect it)
+            fuser = begin
+                      `LC_ALL=C fuser -v -m #{Shellwords.escape(umount_this)} 2>&1`
+                    rescue
```
jreidinger: `rescue => e` and then use it `#{e}`

PR #687, opened 2018-04-25
Fix "installation settings" support for RTL languages

imobachgs:
* Related to https://github.com/yast/yast-yast2/pull/730 and https://github.com/yast/yast-packager/pull/348.

## Before

qt-ar-installation-settings-pre

Main problems: the 'plus' signs in the list of patterns and the alignment of other elements like 'Kdump' title or 'SSH' settings.

## After

qt-ar-installation-settings

qt-en-installation-settings

qt-ar-system-role-selection

## SLE 12 (for comparison)

qt-ar-installation-settings-sle12

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Apr 25 06:03:53 UTC 2018 - igonzalezsosa@suse.com
+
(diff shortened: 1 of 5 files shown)


mvidner:
Are you sure this is for master and not SLE-15-GA?

imobachgs:
No, you are right, it is for `SLE-15-GA`. I've adjusted all the PRs.

PR #688, opened 2018-04-25
Fix no disk

jreidinger:


ancorgs:
I guess this should be closed as replaced by #689

PR #689, opened 2018-04-25
Fix no disks 15

jreidinger:


package/yast2-installation.spec
```diff
@@ -47,9 +47,9 @@ BuildRequires:  yast2 >= 4.0.72
 # Yast::Packages.check_remote_installation_packages
 BuildRequires:	yast2-packager >= 4.0.9
 
-# Y2Storage::StorageManager#activate and #probe as boolean
-BuildRequires: yast2-storage-ng >= 4.0.114
-Requires:      yast2-storage-ng >= 4.0.114
+# Y2Storage::StorageManager#light_probe
```
ancorgs: The new function is not StorageManager#light_probe but StorageManager#devices_for_installation?

PR #691, opened 2018-04-27
Do not warn too soon about missing disks

imobachgs:
Related to https://github.com/yast/yast-autoinstallation/pull/432.


imobachgs:
Thanks! :)

PR #692, opened 2018-04-27
Better handle the restore scripts (bsc#1089643)

lslezak:
- Better handle the restore scripts when YaST is aborted or crashes
  - Move the restore call later, after the X server is stopped so the console is visible. Otherwise the user would see an empty (black) X screen and could think YaST is frozen. The registration rollback migth take 2-3 minutes. See the screenshot, originally the rollback progress message was not visible on the terminal, there was only a black X screen.
  - Run the restore scripts also when YaST crashes
- Delete the .repo files accidentally saved into inst-sys (fixes some problems when restarting YaST after abort or crash)
- Added logging
- 4.0.53

## Screenshots

The console output is now immediately visible when YaST exists, originally a black X screen was shown while the rollback was in progress.
rollback_abort

### DIFF
--- a/startup/First-Stage/F09-start
+++ b/startup/First-Stage/F09-start
@@ -19,7 +19,7 @@ SUSEConnect --cleanup > /dev/null 2>&1
 
 # delete the repositories accidentally saved into inst-sys
 # they make troubles when restarting YaST
-rm -rfv /etc/zypp/repos.d/*
+rm -fv /etc/zypp/repos.d/*.repo
 
 #=============================================
 # 9.1) check for driver update mode


startup/First-Stage/F09-start
```diff
@@ -17,6 +17,10 @@ callHooks preFirstCall
 # Using /dev/null - If there is nothing to do, let it fail silently
 SUSEConnect --cleanup > /dev/null 2>&1
 
+# delete the repositories accidentally saved into inst-sys
+# they make troubles when restarting YaST
+rm -rfv /etc/zypp/repos.d/*
```
jreidinger: this looks too string for me. Why not less strict `rm -fv /etc/zypp/repos.d/*.repo` ? It should help if there is some backup files and so on.

PR #693, opened 2018-05-02
Copy new storage-inst/ subdir to target at the end of the installation

shundhammer:
See also https://github.com/yast/yast-storage-ng/pull/609

This copies `/var/log/YaST2/storage-inst` to the target at the end of the installation.

Any old directory in the target system with that name is removed before so we don't get confused by upgrade after upgrade after upgrade when users submit y2logs with a bug report.

This is also a minor cleanup of that code; it had become quite messy over time.
Now all shell commands are properly logged, and there is less duplication in the code.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,3 @@
-FROM yastdevel/ruby
+FROM yastdevel/ruby:sle15
 COPY . /usr/src/app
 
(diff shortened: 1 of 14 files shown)


shundhammer:
wrong target branch - closing

PR #695, opened 2018-05-03
Keep the selected product in the desktop selection dialog (bsc#1088660)

lslezak:
- Fixes the problem with unselected product when going back - the selected products must not be reset
- See https://bugzilla.suse.com/show_bug.cgi?id=1088660
- Less strict check for selecting the product - do not ask to select a product when there is only one product available, the selection widget is hidden in that case anyway
- Tested manually
- 4.0.56

### DIFF
--- a/src/lib/installation/clients/inst_complex_welcome.rb
+++ b/src/lib/installation/clients/inst_complex_welcome.rb
@@ -194,7 +194,7 @@ module Yast
     # @return [Boolean] true if a product has been selected and license
     # agreement confirmed when required; false otherwise
     def product_selection_finished?
-      if selected_product.nil?
+      if selected_product.nil? && products.size > 1
         Yast::Popup.Error(_("Please select a product to install."))
         return false
       elsif license_confirmation_required? && !selected_product.license_confirmed?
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -194,7 +194,7 @@ def license_confirmation_required?
     # @return [Boolean] true if a product has been selected and license
     # agreement confirmed when required; false otherwise
     def product_selection_finished?
-      if selected_product.nil?
+      if selected_product.nil? && products.size > 1
```
jreidinger: this is a bit error prone, as if `selected_product` is nil and products.size is 1, then in next if `selected_product.license_confirmed?` will crash, not?

lslezak: Ah, good point...

src/lib/installation/widgets/system_roles_radio_buttons.rb
```diff
@@ -52,7 +52,8 @@ def store
         CustomPatterns.show = value == "custom"
         store_orig
 
-        Yast::Packages.Reset([])
+        # keep the selected products
+        Yast::Packages.Reset([:product])
```
mvidner: @lslezak The API is very confusing because we specify the kinds to NOT reset. A better API would say `foo.reset(except: [:product])`

PR #697, opened 2018-05-08
- disable mdadm auto assembly for installation

aschnell:
For https://trello.com/c/rjLmkP8u/373-sles15-p2-1090690-storage-ng-error-encrypting-a-partition-that-was-part-of-an-md-raid.


aschnell:
The failures at Travis seems to be caused by not using the latest yast2-storage-ng package (from yesterday).

shundhammer:
LGTM

PR #705, opened 2018-05-30
Copy active devices

jreidinger:
goal is to have as minimal and non-intrusive change as possible.

### DIFF
--- a/src/lib/installation/clients/copy_files_finish.rb
+++ b/src/lib/installation/clients/copy_files_finish.rb
@@ -42,6 +42,7 @@ module Yast
       textdomain "installation"
 
       Yast.import "AddOnProduct"
+      Yast.import "Arch"
       Yast.import "Linuxrc"
       Yast.import "Installation"
       Yast.import "Directory"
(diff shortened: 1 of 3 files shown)


src/lib/installation/clients/copy_files_finish.rb
```diff
@@ -126,6 +126,18 @@ def main
           end
         end
 
+        # Copy cio_ignore whitelist (bsc#1095033)
+        # Only in install, as update should keep its old config
+        if Mode.installation
```
lslezak: NP: I'd make the condition more strict, like `Stage.initial && Mode.installation && Arch.s390` to make sure it is applied only in  installation and make clear that it's needed only for s390.

jreidinger: well, I copy it from multipath stuff and we check that file really exists there, so it should be fine. BTW this client does is run only in stage initial, otherwise all client does not make sense.

lslezak: Yes, there is `File.exist?` but if you are not familiar with bootloader configuration and s390 then you do not know that. Someone might be wondering why the file is missing on their system...

If you do not want to add the `Arch.s390` check then it would be nice to at least add a comment that this is S390 specific.

PR #708, opened 2018-05-30
Sp4 merge

jreidinger:
not ready for merge yet, as it has old storage stuff, but good to show what we missing in SLE15


jreidinger:
as SLE15 is already released no longer makes sense

PR #710, opened 2018-05-31
Sp3 merge

jreidinger:


jreidinger:
@mvidner please ignore for now travis failure. @lslezak working on it. Problem is that we do not have yet public version of yast-rake for SP4.

PR #712, opened 2018-05-31
Sle15 merge2

jreidinger:
merge fix for SLE15 GA about whitelist of active devices on s390 plus oppurtinity refactoring(!) so no functionality change should happen. Goal of refactoring is beside improving code itself also increase test coverage for touched file.


src/lib/installation/clients/copy_files_finish.rb
```diff
@@ -19,313 +19,159 @@
...
 module Yast
-  class CopyFilesFinishClient < Client
-    include Yast::Logger
-
-    def main
-      Yast.import "Pkg"
-      Yast.import "UI"
+  # Step of base installation finish including mainly files to copy from insts-sys to target system
+  # @note This client expect that SCR is not yet switched.
+  #   Expect the hell to be frozen if you call it with switched SCR.
```
mvidner: If we can test for that then we should `raise`.

src/lib/installation/clients/copy_files_finish.rb
```diff
@@ -19,313 +19,159 @@
...
-        WorkflowManager.GetAllUsedControlFiles
-      )
-      workflows_list = []
-
-      Builtins.foreach(WorkflowManager.GetAllUsedControlFiles) do |one_filename|
-        if Builtins.regexpmatch(one_filename, "/")
-          one_filename = Builtins.regexpsub(one_filename, "^.*/(.*)", "\\1")
-        end
-        workflows_list = Builtins.add(workflows_list, one_filename)
+      log.info "Coping additional control files #{control_files.inspect}"
```
imobachgs: np: "Coping" -> "Copying". I know, it is an old line, but fixing is rather easy :)

src/lib/installation/clients/copy_files_finish.rb
jreidinger: yes, we have method for it - https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/wfm.rb#L116

test/copy_files_finish_test.rb
```diff
@@ -0,0 +1,236 @@
...
+        expect(content).to match(/# Note: Entries added during installation\/update/)
+        expect(content).to match(/blacklist moduleA/)
+        expect(content).to match(/blacklist moduleB/)
+      end
+
+      subject.write
+    end
+
+    it "creates blacklist file if target system does not contain it" do
+      allow(Yast::Linuxrc).to receive(:InstallInf).with("BrokenModules").and_return("moduleA, moduleB")
```
imobachgs: You could move this line to the common `before` block.

test/copy_files_finish_test.rb
```diff
@@ -0,0 +1,236 @@
...
+      blacklist_file = "/mnt/etc/modprobe.d/50-blacklist.conf"
```
imobachgs: What about using a constant? (or, at least, a `let`).

test/copy_files_finish_test.rb
```diff
@@ -0,0 +1,236 @@
...
+        expect(content).to match(/# Note: Entries added during installation\/update/)
+        expect(content).to match(/blacklist moduleA/)
+        expect(content).to match(/blacklist moduleB/)
+      end
+
+      subject.write
+    end
+
+    it "copies information about hardware status" do
+      expect(::FileUtils).to receive(:mkdir_p).with("/mnt/var/lib")
```
imobachgs: In other tests... could happen that it tries to create that directory?

package/yast2-installation.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Thu May 31 10:17:19 UTC 2018 - jreidinger@suse.com
+
+- Fix crash caused by previous fix when multipath is not available
+  (bsc#1095323)
+- Copy active_devices.txt for s390 to prevent blocking of
+  important devices when cio_ignore is active (bsc#1095033)
+- 4.1.0
```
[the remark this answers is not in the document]
jreidinger: no, fix is already part of SLE15-GA branch and send there as it is P1 shipstopper.

test/copy_files_finish_test.rb
jreidinger: by default it returns nil, so it skip this part and test path without it and as bonus also speed up test slightly.

PR #716, opened 2018-06-11
Firstboot: shutdown on failure and hide abort when congratulating

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1095251


src/lib/installation/clients/inst_congratulate.rb
```diff
@@ -233,6 +233,7 @@ def main
 
       Wizard.SetNextButton(:next, Label.FinishButton)
       Wizard.RestoreAbortButton
+      Wizard.HideAbortButton if Stage.firstboot
```
kobliha: Imagine yourself in 2 years: why is it here? Why just firstboot? A comment would help

PR #718, opened 2018-06-15
Do not crash in case /etc/os-release is a directory

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1097661

### DIFF
--- a/src/lib/installation/ssh_config.rb
+++ b/src/lib/installation/ssh_config.rb
@@ -70,7 +70,8 @@ module Installation
           name += " #{os_release["VERSION"]}"
         end
         name
-      rescue Errno::ENOENT # No /etc/os-release found
+      # No /etc/os-release file found or it is a directory (bsc#1097661)
+      rescue Errno::ENOENT, Errno::EISDIR
         default_name
       end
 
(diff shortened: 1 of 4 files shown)


src/lib/installation/ssh_config.rb
```diff
@@ -70,7 +70,8 @@ def name_for(mount_point)
           name += " #{os_release["VERSION"]}"
         end
         name
-      rescue Errno::ENOENT # No /etc/os-release found
+      # No /etc/os-release file found or it is a directory (bsc#1097661)
+      rescue Errno::ENOENT, Errno::EISDIR
```
jreidinger: and what about other Errno like insuficient permission, disk read failure and so on? Why not use generic Errno?

teclator: So you prefer to rescue from any SystemCallError in this case? In most of the scenarios where a config is read we just rescue when it is not present. 

This was a particular scenario, that it is also strange. But of course the other scenarios you are proposing could happen ..

jreidinger: well, It is not critical information. So lets survive and just log it.

PR #738, opened 2018-09-04
Fix path to media/build info file

dgdavid:
From https://bugzilla.suse.com/show_bug.cgi?id=1062297

> after installation I could not find /etc/YaST2/build

> There is no more /build - it has been integrated into /media. And yast needs to copy it in place of /build

~~This RFC is to know your point of view about what to do with the destination of `/media.1/media` in the installed system:~~

~~* Keep as it is: `/build`~~
~~* Change to `/media`~~
~~* Copy to both: `/media` and `/build` (does have sense???)~~

This PR change the source from `/media.1/build` to `/media.1/media` and destination to `/media`, but also keeping `/build` for backward compatibility with other (not adapted yet) tools, as @lslezak suggested in #yast IRC channel.

---

Related to https://trello.com/c/O0F6RkaQ

### DIFF
--- a/src/lib/installation/clients/copy_files_finish.rb
+++ b/src/lib/installation/clients/copy_files_finish.rb
@@ -80,8 +80,10 @@ module Yast
       # FIXME: probably won't work as expected with new multimedia layout.
       #   Ideally final modified control.xml should be saved.
       copy_control_file
-      # Copy /media.1/build to the installed system (fate#311377)
-      copy_build_file
+
+      # Copy /media.1/media to the installed system (fate#311377)
+      # Formerly /media.1/build (bsc#1062297)
+      copy_media_file
       copy_product_profiles
 
       # List of files used as additional workflow definitions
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/copy_files_finish.rb
```diff
@@ -303,19 +305,24 @@ def copy_control_file
...
-      ::FileUtils.cp(build_file, destination)
-      ::FileUtils.chmod(0o644, destination)
+    def copy_media_file
+      media_file = Pkg.SourceProvideOptionalFile(Packages.GetBaseSourceID, 1, "/media.1/media")
+
+      if media_file
+        log.info("Copying /media.1/media file to /media")
+        media_destination = File.join(Installation.destdir, Directory.etcdir, "media")
+        ::FileUtils.cp(media_file, media_destination)
+        ::FileUtils.chmod(0o644, media_destination)
```
lslezak: The `o` flag is not needed, in Ruby numbers starting with 0 are treated as octal numbers. IIRC we do not use it in the YaST code.

```console
# irb
irb(main):001:0> 0o644
=> 420
irb(main):002:0> 0644
=> 420
```

src/lib/installation/clients/copy_files_finish.rb
```diff
@@ -303,19 +305,24 @@ def copy_control_file
...
+
+        # FIXME: `/build` is deprecated, only keeped for backward compatibility. It must be removed
```
lslezak: Grammar: `kept`

src/lib/installation/clients/copy_files_finish.rb
dgdavid: Good to know :) I'll replace all occurrences of `0o644` by `0644` in the project.

Thank you!

dgdavid: We changed them, but then we realize that it is an offense for the current Rubocop configuration :stuck_out_tongue_winking_eye:

lslezak: Ah, sorry for that, I thought the default is the opposite... :grimacing:

PR #740, opened 2018-09-12
Do not restore old backups when upgrade fails

dgdavid:
Related to  https://bugzilla.suse.com/show_bug.cgi?id=1097297

---

Se also https://github.com/yast/yast-update/pull/110

### DIFF
--- a/startup/YaST2.call
+++ b/startup/YaST2.call
@@ -148,7 +148,7 @@ function prepare_for_ssh () {
 # ---
 #
 	:
-        export Y2STYLE="installation_slim" 
+        export Y2STYLE="installation_slim"
 	set_inst_qt_env
 }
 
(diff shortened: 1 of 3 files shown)


startup/YaST2.call
```diff
@@ -285,14 +285,22 @@ function ssh_reboot_message()
...
-    if [ -d /mnt/var/adm/backup/system-upgrade ]; then
-	log "\tStarting restore scripts"
-        for i in /mnt/var/adm/backup/system-upgrade/restore-*.sh; do
-	    log "\tStarting $i"
-            sh $i /mnt
-        done
-    fi
+  # restores available backups created after boot (bsc#1097297)
+  if [ -d /mnt/var/adm/backup/system-upgrade ]; then
+    boottime=$(awk '/^btime/{print $2}' < /proc/stat)
```
lslezak: Like as in https://github.com/yast/yast-update/pull/110#discussion_r217700869, we should check how the boot time is set on RPi...

dgdavid:
Closed in favor of #742, which is using a different approach since there is devices without RTC (like RPi)

PR #742, opened 2018-09-18
Do not restore old backups when upgrade fails

dgdavid:
> The backup only will be restored when the ID and VERSION_ID matches in
both, the current and backed os-release file.

### DIFF
--- a/startup/YaST2.call
+++ b/startup/YaST2.call
@@ -282,17 +282,46 @@ function ssh_reboot_message()
 	fi
 }
 
+#---[ validate_backup ]---#
+function validate_backup () {
+  # Check that the ID and VERSION_ID in the backup os-release file
+  # matches with the values in the /etc/os-release to considered it
+  # as a valid backup to be restored (bsc#1097297)
+  #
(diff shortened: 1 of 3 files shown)


startup/YaST2.call
```diff
@@ -282,17 +282,45 @@ function ssh_reboot_message()
...
 }
 
+#---[ validate_backup ]---#
+function validate_backup () {
+  # Check that the ID and VERSION_ID in the backup os-release file
+  # matches with the values in the /etc/os-release to considered it
+  # as a valid backup to be restored (bsc#1097297)
+  #
+  # See also https://www.freedesktop.org/software/systemd/man/os-release.html
+  backup_id=$(grep "^ID=" /mnt/var/adm/backup/os-release | cut -d'=' -f2 | tr -d '"')
```
lslezak: I'd just define a prefix variable so if change it then we will need to fix just one place.

```
prefix="/mnt"
...
... grep $prefix/etc/os-release ...

```

dgdavid: Good catch :)

PR #747, opened 2018-09-25
Skip the roles' selection dialog

dgdavid:
When there is only one role available, directly applies it and skip the dialog.

Related to https://trello.com/c/uzNWUsjG/337-fate-324713-optimize-system-role-handling-dont-show-system-role-dialog-when-only-one-or-none-system-role-is-available

### DIFF
--- a/test/select_system_role_test.rb
+++ b/test/select_system_role_test.rb
@@ -25,7 +25,7 @@ describe ::Installation::SelectSystemRole do
       subject.class.original_role_id = nil
 
       allow(Yast::ProductFeatures).to receive(:ClearOverlay)
-      allow(Yast::ProductFeatures).to receive(:SetOverlay) # .with
+      allow(Yast::ProductFeatures).to receive(:SetOverlay)
       allow(Yast::Packages).to receive(:SelectSystemPatterns)
     end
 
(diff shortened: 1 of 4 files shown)


test/select_system_role_test.rb
```diff
@@ -41,6 +41,77 @@
...
+        expect(Yast::Wizard).to_not receive(:SetContents)
+        expect(Yast::UI).to_not receive(:UserInput)
+        expect(Yast::UI).to_not receive(:QueryWidget)
+
+        subject.run
+      end
+
+      context "and going forward" do
+        before do
+          allow(Yast::UI).to receive(:UserInput).and_return(:next)
```
lslezak: NP: personally I prefer `expect` for the calls which define the context. This ensures the context is really used and is valid. With weaker `allow` it might happen that the correct context is not set.

test/select_system_role_test.rb
```diff
@@ -41,6 +41,77 @@
...
+          expect(Yast::ProductFeatures).to receive(:SetOverlay)
+
+          subject.run
+        end
+      end
+
+      context "and going back" do
+        before do
+          subject.class.original_role_id = "bar"
+          allow(Yast::GetInstArgs).to receive(:going_back).and_return(true)
```
lslezak: The same here, really checking that `going_back` was called ensures the "and going back" context is really used.

test/select_system_role_test.rb
dgdavid: I see your point, but I disagree.

Let me explain. 

It's true that using `allow` we could introduce not needed/executed "conditions" without any warning about that. But this does not seem to me to be a strong reason to use _expectations_ like conditions.

Usually, in a `before` block I expect to see `allow` instead `expect` and I only use the `expect` to certainly test the result that I looking for a specific condition/situation, not to prepare the scenario itself.

I don't know if I explain myself. But, 

```ruby
context "when dgdavid tries to give him POV to lslezak" do
  let(:lslezak) { @lslezak }
  let(:dgdavid) { @dgdavid }
  let(:dgdavid_pov) { "See the comment" }

  before do
    allow(lslezak).to receive(:other_pov)
    allow(dgdavid).to receive(:what_do_you_think?).and_return(dgdavid_pov)
  end

  it "explains himself properly" do
    expect(dgdavid.what_do_you_think?).to be_well_expressed
  end
end
```

:wink:

PR #748, opened 2018-09-26
[WIP][RFC] Use fbiterm always, if there is enough memory

dgdavid:
NOTE: it was removed the logic related to fallback to English language,
which should be restored.


lslezak:
Using fbiterm has some drawbacks, see https://fate.suse.com/325746. Until somebody decides that we cannot continue here... So closing for now.

PR #750, opened 2018-10-22
Skip role selection dialog when there is only one role available (:back direction)

dgdavid:
> Related to fate#324713 and [bsc#1112564](https://bugzilla.suse.com/show_bug.cgi?id=1112564), avoid also to show the role
selection dialog user is going back and there is only one role
available.

### DIFF
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -62,8 +62,9 @@ module Installation
         # If coming back, we have to run the additional dialogs first...
         clients = additional_clients_for(previous_role_id)
         direction = run_clients(clients, going_back: true)
-        # ... and only run the main dialog (super) if we are *still* going back
-        return direction unless direction == :back
+        # ...and only run the main dialog (super) if there is more than one role (fate#324713) and we
+        # are *still* going back
+        return direction if single_role? || direction != :back
       end
 
       if single_role?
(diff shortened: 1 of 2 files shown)


src/lib/installation/select_system_role.rb
```diff
@@ -60,12 +60,15 @@ def run
 
       if Yast::GetInstArgs.going_back
         # If coming back, we have to run the additional dialogs first...
-        clients = additional_clients_for(self.class.original_role_id)
+        clients = additional_clients_for(previous_role_id)
         direction = run_clients(clients, going_back: true)
         # ... and only run the main dialog (super) if we are *still* going back
         return direction unless direction == :back
```
lslezak: Does it work as expected? If you are going back (`direction = :back`) it goes on...

src/lib/installation/select_system_role.rb
```diff
@@ -60,12 +60,15 @@ def run
...
-      elsif single_role?
+      end
+
+      if single_role?
         # Apply the role and skip the dialog when there is only one (fate#324713)
+        log.info "Only one role available, applying it and skipping the dialog"
         clear_role
         return select_role(roles.first.id)
```
lslezak: ... and for a single role we select it again here...

src/lib/installation/select_system_role.rb
```diff
@@ -137,23 +144,19 @@ def single_role?
...
+        # Changing the role, show a Continue-Cancel popup to user
         msg = _("Changing the system role may undo adjustments you may have done.")
         return :back unless Yast::Popup.ContinueCancel(msg)
       end
-      self.class.original_role_id = role_id
-
-      apply_role(SystemRole.find(role_id))
 
+      apply_role(role_id)
       run_clients(additional_clients_for(role_id))
```
lslezak: And we run `run_clients()` here, but *without* the `going_back: true` so the result is `:next`.

PR #752, opened 2018-10-25
Remember the selected role (related to FATE#325834), fixed testsuite

lslezak:
- The selected role was not remembered, `Installation::SystemRole.current` returned `nil` even after selecting a role.
- This makes troubles with the role finish clients, they are not executed because the finish client thinks nothing was selected:
  ```
  [Ruby] installation/system_role.rb:175 There is no role selected so nothing to do
  ```
- Fixed a problem with a missing mock, running `rake test:unit` executed the `/usr/lib/YaST2/bin/mask-systemd-units` command resulting in displaying this dialog (sorry, Czech only):
systemd_popup
- Fixed by mocking all `Yast::Execute.locally!` calls
- 4.1.21

### DIFF
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -217,6 +217,7 @@ module Installation
     def apply_role(role_id)
       log.info "Applying system role '#{role_id}'"
 
+      SystemRole.select(role_id)
       self.class.previous_role_id = role_id
       role = SystemRole.find(role_id)
       role.overlay_features
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -35,11 +35,6 @@ module Installation
   class SelectSystemRole < ::UI::InstallationDialog
     include UI::TextHelpers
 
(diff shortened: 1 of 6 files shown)


src/lib/installation/select_system_role.rb
```diff
@@ -217,6 +217,7 @@ def clear_role
     def apply_role(role_id)
       log.info "Applying system role '#{role_id}'"
 
+      SystemRole.select(role_id)
```
jreidinger: maybe you can do `role = SystemRole.select(role_id)` and remove line 222

src/lib/installation/select_system_role.rb
```diff
@@ -217,6 +217,7 @@ def clear_role
...
       self.class.previous_role_id = role_id
```
jreidinger: and when you select a role, maybe you can also remove this line and handle it only in select_system_role or is there still purpose for this?

jreidinger:
tests failing, maybe you need to update mocks, so select return really role?

PR #757, opened 2018-11-02
Force the installer to use fbiterm

imobachgs:
Force the installer to use `fbiterm`.

See also yast/yast-country#192 and yast/yast-autoinstallation#477.

## Screenshots

textmode-jp
textmode-en

### DIFF
--- a/startup/common/language.sh
+++ b/startup/common/language.sh
@@ -30,25 +30,22 @@ function check_run_fbiterm () {
 	if test "$TERM" = "linux" -a \
 		\( "$TTY" = /dev/console -o "$TTY" != "${TTY#/dev/tty[0-9]}" \);
 	then
-		case "$LANG" in
-		ja*.UTF-8|ko*.UTF-8|zh*.UTF-8)
-		# check whether fbiterm can run on console
-		if test -x /usr/bin/fbiterm && \
-			/usr/bin/fbiterm echo >/dev/null 2>&1;
-		then
-			RUN_FBITERM=1
-		else
-			# use english
-			export LANG=en_US.UTF-8
-			export LC_CTYPE=en_US.UTF-8
-		fi
-		;;
-		ja*|ko*|zh*)
+	    # check whether fbiterm can run on console
+	    if test -x /usr/bin/fbiterm && \
+		    /usr/bin/fbiterm echo >/dev/null 2>&1;
+	    then
+		RUN_FBITERM=1
+	    else
 		# use english
 		export LANG=en_US.UTF-8
 		export LC_CTYPE=en_US.UTF-8
-		;;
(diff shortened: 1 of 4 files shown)


imobachgs:
I've tested with different languages and it fails with this list: ar_EG, bn_BD, gu_IN, hi_IN, km_KH, mr_IN, pa_IN, ta_IN, th_TH. In these cases, we are switching back to en_US but we still launch `fbiterm` (unless it is not possible for any other reason).

I was thinking about to maintain the list of languages elsewhere because it will be used by yast2-country too (but I am not sure whether it is better to keep this simple solution by now).

startup/common/language.sh
```diff
@@ -30,25 +30,22 @@ function check_run_fbiterm () {
...
+	    else
 		# use english
 		export LANG=en_US.UTF-8
 		export LC_CTYPE=en_US.UTF-8
-		;;
-	esac
+	    fi
+
+	    case "$LANG" in
+		ar_EG*|bn_BD*|gu_IN*|hi_IN*|km_KH*|mr_IN*|pa_IN*|ta_IN*|th_TH*)
```
lslezak: Please add a comment why we handle these languages differently.

lslezak: And I think we should check only the language part of the locale, i.e. Arabic will be broken also for other countries, not only for Egypt.

startup/common/language.sh
```diff
@@ -30,25 +30,24 @@ function check_run_fbiterm () {
...
 		export LANG=en_US.UTF-8
 		export LC_CTYPE=en_US.UTF-8
-		;;
-	esac
+	    fi
+
+	    case "$LANG" in
+		# These languages are not properly supported by fbiterm causing YaST to crash
+		# (fate#325746).
+		ar*|bn*|gu*|hi*|km*|mr*|pa*|ta_*|th*)
```
lslezak: Just a question, why do we need to handle `ta_` differently? (with the underscore?)

imobachgs: just a typo, sorry.

startup/common/language.sh
```diff
@@ -30,25 +30,24 @@ function check_run_fbiterm () {
...
-		fi
-		;;
-		ja*|ko*|zh*)
+	    # check whether fbiterm can run on console
+	    if test -x /usr/bin/fbiterm && \
+		    /usr/bin/fbiterm echo >/dev/null 2>&1;
+	    then
+		RUN_FBITERM=1
+	    else
 		# use english
```
lslezak: I think we should at least log a warning when the fbiterm is not found (or does not work). This could later help with debugging.

imobachgs:
Thanks for the review!

PR #759, opened 2018-11-07
document rpm as control flow extension

jreidinger:


doc/control-file.md
```diff
@@ -1294,6 +1294,25 @@ product's CD:
     installation itself. Both for first stage installation and also for the
     running system.
 
+#### RPM
+
+It is possible to also modify behavior of installation by RPM.
+There is few requirements for that RPM. At first it needs to provide
```
lslezak: You can use a list instead of "at first... the second...", like:

----

There are few requirements for that RPM:
- It needs to provide the `installer_module_extension` symbol.
- It needs to specify for which￼ product is the extension defined. This is done by providing the `extension_for_product()` symbol.
- It needs an XML control file. The recommended location is ￼ `/usr/share/system-roles/<role_name>.xml`, alternatively you can use deprecated path `/installation.xml` (not recommended for new products).

PR #761, opened 2018-11-08
Demo: eager mode of zombie-killer 0.5

mvidner:
**Don't merge**, the eager mode of the zombie killer is stupid and its result does not handle edge cases like `nil`. It should be used only on code covered with tests. Which this is not.

Produced by [zombie-killer](https://rubygems.org/gems/zombie-killer)
```sh
zk -e .../yast-installation
```

### DIFF
--- a/control/complete_workflow.rb
+++ b/control/complete_workflow.rb
@@ -23,7 +23,7 @@ module Yast
 
       @control = ""
       @profile = ""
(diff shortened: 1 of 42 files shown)


jreidinger:
Do not show this in open pull request, lets just close it and link it from zombie killer if needed.

PR #762, opened 2018-11-15
Sle 15 ga bsc 1098571

schubi2:


src/lib/installation/dialogs/complex_welcome.rb
```diff
@@ -43,6 +43,9 @@ class ComplexWelcome < CWM::Dialog
       # @param disable_buttons [Array<Symbol>] List of buttons to disable
       def initialize(products, disable_buttons: [])
         textdomain "installation"
+
+        Yast.import "Wizard"
```
dgdavid: Is it possible to move this import at the beginning? Next to [`Yast.import "UI"`](https://github.com/yast/yast-installation/blob/3ba5c250a480d2db315fb73c0e745ebc918083f2/src/lib/installation/dialogs/complex_welcome.rb#L22)

dgdavid: Or maybe move the `Yast.import "UI"` here.

PR #765, opened 2018-11-24
Provide icon with module (boo#1109310)

hellcp:


src/desktop/installation.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Categories=Settings;System;Qt;
+
+Icon=yast-installation
+Exec=
```
lslezak: I'm not sure, maybe this violates the desktop file standard: *"The Exec key must contain a command line."*, see https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s07.html for more details.

Although it's not a real problem, some OBS checks could complain in the future. I'm not a desktop file expert, but maybe using a `/bin/true` dummy value here could be a good idea.

When I tried to build the package locally I got these errors:

```
ERROR: No sufficient Category definition: /home/abuild/rpmbuild/BUILDROOT/yast2-installation-4.1.30-1.x86_64//usr/share/applications/YaST2/installation.desktop 
Please refer to https://en.opensuse.org/openSUSE:Packaging_desktop_menu_categories
WARNING: Empty GenericName: /home/abuild/rpmbuild/BUILDROOT/yast2-installation-4.1.30-1.x86_64//usr/share/applications/YaST2/installation.desktop
ERROR: No sufficient Category definition: /home/abuild/rpmbuild/BUILDROOT/yast2-installation-4.1.30-1.x86_64//usr/share/applications/YaST2/upgrade.desktop 
Please refer to https://en.opensuse.org/openSUSE:Packaging_desktop_menu_categories
WARNING: Empty GenericName: /home/abuild/rpmbuild/BUILDROOT/yast2-installation-4.1.30-1.x86_64//usr/share/applications/YaST2/upgrade.desktop
Errors in installed desktop file detected. Please refer to http://en.opensuse.org/SUSE_Package_Conventions/RPM_Macros
```

For some reason this check is not executed at Travis so it does not fail, I'll check if that could be improved...

src/desktop/installation.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
```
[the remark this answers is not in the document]
hellcp: Did that, although I am still not happy about the requirement of having desktop files for icons. It's a neat feature, but it should be an addition to real icon loading.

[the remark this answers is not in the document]
lslezak: Yes, using icons should be possible even without a desktop file.

Please open a separate bug for that so we can improve that.

src/desktop/installation.desktop
```diff
@@ -0,0 +1,12 @@
...
+Type=Application
+Hidden=true
+Categories=Settings;System;Documentation;
+
+Icon=yast-installation
+Exec=/bin/true
+
+Name=Installation
+GenericName=System Upgrade
+Comment=Install system to the system
```
lslezak: Ah, something like "Install a new system" would be better here, but as this is just a dummy desktop file it does not matter much...

hellcp: Ah, miss on my part, I hope to remove those desktop files as soon as I patch YaST with a way to load icons directly.

PR #766, opened 2018-12-06
Improved saving y2logs during installation (bsc#1118643)

lslezak:
- Use the `/mnt/tmp` space instead of the RAM disk to avoid possible crash (out of memory)
- Tested manually in installed system with `TMPDIR=/tmp/tmp /usr/sbin/save_y2logs`, it used `/tmp/tmp` for saving the temporary data (like `dmesg` and jounal).
- It might not fix the YaST crash but let's try...
- 4.1.31

### DIFF
--- a/src/lib/installation/clients/save_y2logs.rb
+++ b/src/lib/installation/clients/save_y2logs.rb
@@ -17,8 +17,14 @@ module Yast
           Yast::Directory.logdir
         )
 
+        # use the target /tmp when available to save memory
+        target_tmp = File.join(Yast::Installation.destdir, "tmp")
+        tmpdir = File.exist?(target_tmp) ? target_tmp : "/tmp"
+
         WFM.Execute(Yast::Path.new(".local.bash"),
-          "/usr/sbin/save_y2logs '#{target_path}/yast-installation-logs.tar.xz'")
+          # set TMPDIR for `mktemp -d` so it uses the target disk, not inst-sys RAM disk
+          # https://github.com/yast/yast-yast2/blob/482eecb6064e2a904864fdab17e8c4bed41065ff/scripts/save_y2logs#L72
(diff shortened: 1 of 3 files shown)


src/lib/installation/clients/save_y2logs.rb
```diff
@@ -17,8 +17,14 @@ def main
...
+        # use the target /tmp when available to save memory
+        target_tmp = File.join(Yast::Installation.destdir, "tmp")
+        tmpdir = File.exist?(target_tmp) ? target_tmp : "/tmp"
+
         WFM.Execute(Yast::Path.new(".local.bash"),
-          "/usr/sbin/save_y2logs '#{target_path}/yast-installation-logs.tar.xz'")
+          # set TMPDIR for `mktemp -d` so it uses the target disk, not inst-sys RAM disk
+          # https://github.com/yast/yast-yast2/blob/482eecb6064e2a904864fdab17e8c4bed41065ff/scripts/save_y2logs#L72
+          "TMPDIR=#{tmpdir} /usr/sbin/save_y2logs '#{target_path}/yast-installation-logs.tar.xz'")
       end
```
jreidinger: looks good. BTW what about ruby own tmpdir creation which can be deleted afterwards, so we immediatelly also decrease memory usage? http://ruby-doc.org/stdlib-2.0.0/libdoc/tmpdir/rdoc/Dir.html#method-c-mktmpdir

lslezak: `$TMPDIR` defines the prefix for the `mktemp -d` call used in `save_y2logs`, see `man mktemp`.

We do not need to delete it, it's deleted by the script itself: https://github.com/yast/yast-yast2/blob/482eecb6064e2a904864fdab17e8c4bed41065ff/scripts/save_y2logs#L148

PR #768, opened 2018-12-10
More aggresive inst-sys cleaner (bsc#1118643)

lslezak:
- Clean the libzypp cache when running in graphical mode with less than 1GB RAM (in text mode keep the current limit 640MB), this avoids crashes on low memory systems.
- Updated the glob list, the `appdata-icons.tar.gz` file now also has a hash prefix, just like the other files. (See the [repo data](https://download.opensuse.org/distribution/leap/15.1/repo/oss/repodata/) for Leap 15.1.)

- 4.1.32

### DIFF
--- a/test/instsys_cleaner_test.rb
+++ b/test/instsys_cleaner_test.rb
@@ -4,9 +4,6 @@ require_relative "./test_helper"
 
 require "installation/instsys_cleaner"
 
-def stub_logging
-end
-
 describe Installation::InstsysCleaner do
   describe ".make_clean" do
     context "in the initial stage" do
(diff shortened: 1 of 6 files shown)


test/instsys_cleaner_test.rb
```diff
@@ -18,13 +15,14 @@ def stub_logging
...
       end
 
-      context "removes the libzypp cache if the memory is less than 640MB" do
+      context "a bit less than 640MB memory in text mode" do
         before do
           # 512MB - 1B
           expect(Yast2::HwDetection).to receive(:memory).and_return((512 << 20) - 1)
-          allow(Dir).to receive(:[]).and_return([])
           allow(described_class).to receive(:unmount_kernel_modules)
         end
```
dgdavid: There is a leftover below, isn't it?

Or maybe is the context "a bit less than 640MB memory in GUI mode" missing?

lslezak: Resolved on IRC: the context "a bit less than 640MB in GUI" would be the same as the existing "a bit less than 1GB in GUI", it would not provide any value.

PR #769, opened 2018-12-12
Hardening

jreidinger:


src/include/installation/inst_inc_first.rb
```diff
@@ -131,25 +134,23 @@ def SetInitialInstallation
...
-              WFM.Execute(path(".local.bash_output"), bashcmd)
-            )
-          end
+        ZYPP_DATA.each do |zypp_data_item|
+          next unless FileUtils.Exists(zypp_data_item)
+          Builtins.y2warning(
+            "Directory '%1' exists, removing...",
+            String.Quote(zypp_data_item)
+          )
+          bashcmd = "/usr/bin/rm -rf #{zypp_data_item.shellescape}"
```
shundhammer: isn't that overkill? That can only ever come from the constant string array create a few lines above, and we already know that it doesn't contain any weird characters.

It doesn't really hurt, but it was not insecure before.

But nevermind.

jreidinger: yep, I see that it was secure before. I just want to make it obvious for reader. So he do not need to check it himself that it is really safe.

src/include/installation/inst_inc_first.rb
```diff
@@ -131,25 +134,23 @@ def SetInitialInstallation
...
+          Builtins.y2milestone(
+            "Result: %1",
+            WFM.Execute(path(".local.bash_output"), bashcmd)
```
shundhammer: Calling an external shell command in a logging call scares me a lot more, though.

IMHO this should not be done in the logging; it should be done before, the result should be stored in a variable, and the variable should be logged. The key part, calling the external program, should be independent IMHO.

There are chances that someone will disable the logging, and then the operation will not happen. That's a real risk.

jreidinger: well, in ruby it does not happen. Even arguments for y2debug is evaluated, as y2debug is called and dropped, but arguments is evaluated in advance.
But I really dislike calling anything real in logs, so I am not against changing it.

src/lib/installation/clients/inst_finish.rb
```diff
@@ -161,7 +161,7 @@ def handle_kexec
       # run new kernel via kexec instead of reboot
 
       # command for reading kernel_params
-      cmd = "ls '#{String.Quote(Directory.vardir)}/kexec_done' |tr -d '\n'"
+      cmd = "/usr/bin/ls '#{String.Quote(Directory.vardir)}/kexec_done' | /usr/bin/tr -d '\n'"
```
shundhammer: Most likely this is a non-issue, but we are now using shellescape everywhere, so why not here? Putting the entire path into single quotes is probably not necessary. This might be much simpler as

`"/usr/bin/ls #{Directory.vardir.shellescape}/kexec_done | ..."`

Or am I wrong?

src/lib/installation/clients/inst_save_hardware_status.rb
```diff
@@ -36,14 +36,10 @@ def main
...
-          path(".target.bash_output"),
-          "\n" \
-            "hwinfo --pci --block --mouse --keyboard --isdn --save-config=all\n" \
-            "\n" \
-            "[ -d /var/lib/hardware/udi/org/freedesktop/Hal/devices ] && perl -pi -e \"s/hwinfo.configured = 'new'/hwinfo.configured = 'no'/\" /var/lib/hardware/udi/org/freedesktop/Hal/devices/*"
-        )
+      @out = SCR.Execute(
+        path(".target.bash_output"),
+        "/usr/sbin/hwinfo --pci --block --mouse --keyboard --isdn --save-config=all\n" \
+          "[ -d /var/lib/hardware/udi/org/freedesktop/Hal/devices ] && /usr/bin/perl -pi -e \"s/hwinfo.configured = 'new'/hwinfo.configured = 'no'/\" /var/lib/hardware/udi/org/freedesktop/Hal/devices/*"
```
shundhammer: `[` is just an alias to `test`, so maybe use `/usr/bin/test` instead? (and get rid of the closing `]`)

src/lib/installation/clients/inst_finish.rb
jreidinger: no, you are right, that it will be consistent with other modules. I use this just to not need write require :smile:

src/include/installation/inst_inc_first.rb
shundhammer: ok.

shundhammer:
Unconditional LGTM now.

PR #771, opened 2019-01-02
Cleanup

jreidinger:


mvidner:
Regarding the scenarios, they are something different because they are part of an interface with the product managers, and they are [documented here](https://github.com/yast/yast-installation/search?l=Markdown&q=scenarios). So we should remove also the docs, and also the [control file schema](https://github.com/yast/yast-installation-control/search?q=scenarios&unscoped_q=scenarios).

jreidinger:
@mvidner updated and https://github.com/yast/yast-installation-control/pull/77

PR #772, opened 2019-02-01
Set the right file permissions for the SSH deamon

schubi2:


src/lib/installation/ssh_key.rb
```diff
@@ -47,7 +47,16 @@ def write_files(dir)
...
         IO.write(path, file.content)
-        File.chmod(file.permissions, path)
+        if file.filename =~ /^ssh_host_.*(key)$/
+          # ssh deamon accepts only private keys with restricted
+          # file permissions.
+          log.info("Set permissions of #{file.filename} to 0o600")
+          File.chmod(0o600, path)
+        else
+          # Taking already given permissions
+          log.info("Set permissions of #{file.filename} to #{file.permissions.to_s(8)}")
```
lslezak: In the line above you yse the `0o` prefix for the octal number, I'd use it here as well:
```suggestion
          log.info("Set permissions of #{file.filename} to 0o#{file.permissions.to_s(8)}")
```

or use `format`

```suggestion
          log.info("Set permissions of #{file.filename} to #{format("0o%o", file.permissions)}"
```

PR #777, opened 2019-03-04
Unified "repositories" label (bsc#1123679)

lslezak:
- Unify the "repositories" and "add-on product" labels
- https://bugzilla.suse.com/show_bug.cgi?id=1123679
- 4.1.42

## Questions?

- The new text is too long for text mode, should we use a shorter text in text mode? The original `Previously Used Repositories` should be good enough. (Keep in mind that in some languages the translation might be even longer so a longer part might be clipped off.)

#### Graphical

Full label:

old_repos_at_upgrade

#### Text Mode

Shorter label to fit on the screen:

old_repos_at_upgrade_text

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Mar  7 11:55:36 UTC 2019 - Ladislav Slezák <lslezak@suse.cz>
+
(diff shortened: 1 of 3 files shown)


lslezak:
After the discussion it turned out the old label was good enough (https://bugzilla.suse.com/show_bug.cgi?id=1123679#c11)...

PR #778, opened 2019-03-05
Respect Textmode=1 when installing over SSH with a DISPLAY (bsc#1047470)

mvidner:
https://trello.com/c/XddgyRZS
https://bugzilla.suse.com/show_bug.cgi?id=1047470

The fix is: in `check_network`, which is called for SSH installations, only check `$DISPLAY` when `Textmode` is not set. 

Additional small fix: no need to check the qt plugin if we haven't selected it in the first place. Note that some logical combinations don't assign `Y2_MODE`. That is fine because for SSH, it is preset to ncurses before calling this function.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Mar  5 10:31:35 UTC 2019 - mvidner@suse.com
+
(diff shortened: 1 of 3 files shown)


mvidner:
Tested manually:
- Textmode=1 without this PR -> Qt
- Textmode=1 with this PR -> ncurses
- Textmode=0 with this PR -> Qt

PR #779, opened 2019-03-05
Respect Textmode=1 when installing over SSH with a DISPLAY (SP5)

mvidner:
A backport of #778

### DIFF
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.3.0
+Version:        3.3.0.1
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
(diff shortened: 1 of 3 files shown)


package/yast2-installation.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.3.0
+Version:        3.3.0.1
```
lslezak: SP4 uses 3.2.x (3.2.60), you can continue with 3.3.x here (3.3.1), no need for the fourth digit...

mvidner: Well, the previous PR, which used 3.3.0, was wrong. 3.3.0 is already used in master for something else, see PR #580.

lslezak: Ah, I overlooked that, thanks!

PR #780, opened 2019-03-05
Stop using obsolete XVersion; check textdomain presence

mvidner:
It was useful in 2006 when X.org 6 and X.org 7 had different places for files
- https://bugzilla.suse.com/show_bug.cgi?id=1144627

textdomain:
- https://bugzilla.suse.com/show_bug.cgi?id=1130822

### DIFF
--- a/startup/YaST2.call
+++ b/startup/YaST2.call
@@ -27,8 +27,6 @@
 #
 #set -x
 
-. /etc/YaST2/XVersion
-
 #=============================================
 # Functions...
 #---------------------------------------------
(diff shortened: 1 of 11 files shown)


startup/YaST2.call
```diff
@@ -460,7 +457,6 @@ set_syslog ; set_proxy
 #---------------------------------------------
 # 2.1) setup PATH
 PATH=$ybindir:$PATH
-[ -z Xstartpath ] || PATH=/var/X11R6/bin:$PATH
```
mvidner: BTW this line has prompted the cleanup: [ShellCheck](https://www.shellcheck.net/) correctly notes that we are testing a nonempty literal for emptiness (a `$` is missing)

PR #781, opened 2019-03-06
Fix perms

jreidinger:


package/yast2-installation.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Mar  6 13:01:03 UTC 2019 - jreidinger@suse.com
+
+- Do not change polkit default priviledges during upgrade
```
dgdavid: NP: I thought that "priviledges" had a typo, but it seems [that it is right (but obsolete)](https://en.wiktionary.org/wiki/priviledge#English)

src/lib/installation/clients/save_config_finish.rb
```diff
@@ -194,7 +194,8 @@ def main
             "globals",
             "polkit_default_privs"
           )
-          if !polkit_default_privs.nil? && polkit_default_privs != ""
+          # Do not write polkit privs during update (bsc#1120720)
```
dgdavid: BTW, quite long `main` method. I would have taken the opportunity to extract this to its own (private) method.

jreidinger: well, it is not so much tested and I do not want to spend here too much time, so lets do it in future (r)

PR #783, opened 2019-03-08
Fix packages and patterns selection

joseivanlopez:
## Problem

The installation summary shows a list of patterns to install, but if you go back until the Role Selection screen and then go forward again without changing the role, the list of patterns shown by the installation summary changes (there are much less patterns).

* https://bugzilla.suse.com/show_bug.cgi?id=1126517
* https://trello.com/c/OrkgoQYI/795-important-osdistribution-p2-1126517-patterns-are-not-selected-anymore-when-going-back-to-the-system-role-selection-screen

## Solution

The packages/patterns automatically selected by the solver should be deselected before calling the solver again. Otherwise, the solver does not select the same list of patterns (a bug with the solver?).

## Testing

* Tested manually.

### DIFF
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -212,17 +212,31 @@ module Installation
       role.overlay_features
       adapt_services(role)
 
-      # Reset pkg and pattern selection as many roles define own roles
-      # so ensure when going back that it will properly set (bsc#1088883)
-      reset_patterns
+      select_packages
     end
 
-    def reset_patterns
-      Yast::Pkg::ResolvableProperties("", :pattern, "").each do |pattern|
-        next if pattern["status"] != :selected
-        Yast::Pkg.ResolvableNeutral(pattern["name"], :pattern, false)
-      end
+    # Selects packages for the currently selected role
+    #
+    # Note that everything that was previously selected by the solver needs to be reset.
+    # Otherwise, the solver would not select the same list of patterns when the same role
+    # is selected again (bsc#1126517).
+    #
+    # Moreover, all patterns should be reset because many roles define their own roles, so
+    # the new patterns need to be properly set when going back (bsc#1088883). Only patterns
+    # selected by the user should be kept.
+    def select_packages
+      # By default, Packages.Reset resets a resolvable if the resolvable was not explicitly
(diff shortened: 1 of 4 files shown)


src/lib/installation/select_system_role.rb
```diff
@@ -212,17 +212,21 @@ def apply_role(role_id)
...
-        Yast::Pkg.ResolvableNeutral(pattern["name"], :pattern, false)
-      end
+    # Selects packages for the currently selected role
+    def select_packages
+      # Note that everything that was previously selected by the solver is reset (bsc#1126517).
+      #
+      # Moreover, all patterns are reset because many roles define their own roles, so the new
+      # patterns should be properly set when going back (bsc#1088883). Only patterns selected by
+      # the user are kept, see Packages.Reset.
+      Yast::Packages.Reset([:product, :patch, :packages, :language])
```
lslezak: I think we should also keep the selected product, resetting it might have a nasty consequences, esp. on SLE where are more base products available, we need to be sure that the original base product selected at the beginning stays untouched.

lslezak: Ah, the argument it what you want to *keep* (!), not *reset*!

joseivanlopez: Yes, reverse logic :)

src/lib/installation/select_system_role.rb
```diff
@@ -212,17 +212,31 @@ def apply_role(role_id)
...
+      # selected by the user. When a resolvable type is given in the parameter list, those
+      # resolvables are only reset when they were automatically selected by the solver.
+      #
+      # Products, patches, packages and languages are only reset if they were automatically
+      # selected by the solver. However, patterns are only kept if they were selected by the
+      # user (note that :pattern is not included in the Reset param list).
+      Yast::Packages.Reset([:product, :patch, :package, :language])
+
       Yast::Packages.SelectSystemPatterns(false)
+      Yast::Packages.SelectSystemPackages(false)
```
lslezak: What about the `:language` type, it does not seem to be selected by `SelectSystemPackages`. Could you check if the language packages are correctly selected again?

joseivanlopez: I have checked it, and the languages packages are correctly selected.

PR #785, opened 2019-03-14
ensure that installation with VNC + textmode still installs via VNC (bsc#1129073)

wfeldt:
### Remarks
- `check_network()` should not implicitly change `Y2_MODE`
- the comment in the code complaining about changing `Y2_MODE` in step 7 was right


jreidinger:
lookds good, just please add changelog entry and increase version

PR #790, opened 2019-03-19
Allow to preselect the product

dgdavid:
## Problem

Having multiple products in a single repository (i.e. openSUSE and Kubic), `system-installation()` could provide the "wrong" one during a network installation, ending with conflicts between product and patterns.

- https://bugzilla.suse.com/show_bug.cgi?id=1124590
- https://trello.com/c/RPnSpewA/808-3-tw-kubic-1124590-optionally-hardcode-the-product-name-in-the-controlxml

## Solution

Originally proposed by @lslezak, 

> to hardcode the product name in the control.xml 

So, the installer would select the right product **without asking to user**.

## Testing

- *Added a new unit test*

## Related PRs

- https://github.com/yast/yast-installation-control/pull/79

### DIFF
--- a/src/lib/installation/clients/inst_complex_welcome.rb
+++ b/src/lib/installation/clients/inst_complex_welcome.rb
@@ -35,6 +35,7 @@ Yast.import "Mode"
 Yast.import "Pkg"
 Yast.import "Popup"
 Yast.import "ProductControl"
+Yast.import "ProductFeatures"
 Yast.import "Stage"
 Yast.import "Timezone"
 Yast.import "Wizard"
@@ -145,14 +146,39 @@ module Yast
     # not know which product we are upgrading yet) nor the product selector
     # (as you cannot change the product during upgrade).
     #
+    # It could return a list with pre-selected product(s), @see #preselected_base_product.
+    #
     # @return [Array<Y2Packager::Product>] List of available base products;
     # empty list in update mode.
     def products
       return @products if @products
 
-      @products = Y2Packager::Product.available_base_products
+      @products = preselected_base_product || available_base_products
       @products = [] if Mode.update && @products.size > 1
       @products
+    end
+
+    # Returns, if any, the preselected base product (bsc#1124590)
+    #
+    # A product can be pre-selected using the `select_product` element in the software section
+    # in the control file.
+    #
+    # @return [nil, Array<Y2Packager::Product>] nil when no preselected product in control file,
+    #                                           a list containing the preselected base product, or
+    #                                           empty list if preselected product is not available
+    def preselected_base_product
+      preselected_product = ProductFeatures.GetStringFeature("software", "select_product")
+
+      return if preselected_product.empty?
+
+      available_base_products.select { |product| product.name == preselected_product }
+    end
+
+    # Returns all available base products
+    #
+    # @return [Array<Y2Packager::Product>] List of available base products
+    def available_base_products
+      @available_base_products ||= Y2Packager::Product.available_base_products
     end
 
     # Determine whether some product is available or not
--- a/src/lib/installation/clients/inst_complex_welcome.rb
+++ b/src/lib/installation/clients/inst_complex_welcome.rb
@@ -167,11 +167,22 @@ module Yast
     #                                           a list containing the preselected base product, or
     #                                           empty list if preselected product is not available
     def preselected_base_product
-      preselected_product = ProductFeatures.GetStringFeature("software", "select_product")
+      selected_product_name = ProductFeatures.GetStringFeature("software", "select_product")
 
-      return if preselected_product.empty?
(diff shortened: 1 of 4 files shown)


lslezak:
AFAIK, this is mainly for Tumbleweed and Kubic, do we need it in SLE15-SP1? I mean, should we rather merge it after SP1 is branched?

src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -145,16 +146,41 @@ def setup_final_choice
...
+    # A product can be pre-selected using the `select_product` element in the software section
+    # in the control file.
+    #
+    # @return [nil, Array<Y2Packager::Product>] nil when no preselected product in control file,
+    #                                           a list containing the preselected base product, or
+    #                                           empty list if preselected product is not available
+    def preselected_base_product
+      preselected_product = ProductFeatures.GetStringFeature("software", "select_product")
+
+      return if preselected_product.empty?
```
lslezak: I'm not sure, maybe explicit `return nil` here would be better readable for less experienced Ruby programmers...

What do you think?

src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -145,16 +146,41 @@ def setup_final_choice
...
+
```
lslezak: I'd add some more logging here, e.g. which product was specified in the `control.xml`, which product was found, etc... It will be hard to debug without any trace in the `y2log`.

src/lib/installation/clients/inst_complex_welcome.rb
dgdavid: To be honest, despite I do not have an strong preference about it,  I usually prefer/use `return if/unless` over `return nil if/unless`. 

For this particular case, I prefer to keep it as it is in order to be consistent with similar statements in the file (lines [92-95](https://github.com/yast/yast-installation/blob/c5a19d6619771ceb28f8d167fcc99f123f1d2159/src/lib/installation/clients/inst_complex_welcome.rb#L92-L95))

Did I convince you? :innocent:

lslezak: OK, :smiley: :+1:

src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -171,7 +171,18 @@ def preselected_base_product
 
       return if preselected_product.empty?
 
-      available_base_products.select { |product| product.name == preselected_product }
+      filtered_base_products = available_base_products.select do |product|
+        product.name == preselected_product
+      end
+      discarded_base_products = available_base_products - filtered_base_products
+
+      log.info(
```
dgdavid: @lslezak  I am not sure if this is enough (and "elegant"). If not, ideas are welcome :)

lslezak: Some notes:
- Prefer several short lines than one long, that's harder to read it in terminal (the y2log lines are already quite long)
- Logging is for developers, we do not need to care about nice formatting but rather prefer the exact details (e.g. distinguish between `""` and `nil`)

My suggestion:

```ruby
log.info("control.xml wants to preselect the #{preselected_product.inspect} product")

if !discarded_base_products.empty?
  log.info("Ignoring the other available products: #{discarded_base_products.map(&:name).inspect}")
end
```

dgdavid:
> In reply to https://github.com/yast/yast-installation/pull/790#discussion_r269994417

Thanks for the notes :) 

I'll use your suggestions instead, but without `.inpsect` since it seems to not be neccessary:

```irb
irb(main):056:0> "products: #{[OpenStruct.new(name: ""), OpenStruct.new(name: nil)].map(&:name)}"                  
=> "The following available products: [\"\", nil]"
irb(main):057:0> "available products: #{[OpenStruct.new(name: ""), OpenStruct.new(name: nil)].map(&:name).inspect}"          
=> "The following available products: [\"\", nil]"
irb(main):058:0>
```

src/lib/installation/clients/inst_complex_welcome.rb
dgdavid: > In reply to https://github.com/yast/yast-installation/pull/790#discussion_r269994417

Thanks for the notes :) 

I'll use your suggestions instead, but without `.inpsect` since it seems to not be neccessary:

```irb
irb(main):056:0> "products: #{[OpenStruct.new(name: ""), OpenStruct.new(name: nil)].map(&:name)}"                  
=> "The following available products: [\"\", nil]"
irb(main):057:0> "available products: #{[OpenStruct.new(name: ""), OpenStruct.new(name: nil)].map(&:name).inspect}"          
=> "The following available products: [\"\", nil]"
irb(main):058:0>
```

PR #791, opened 2019-03-29
Add a NtpSetup installation dialog

dgdavid:
## Problem

The screen for configuring NTP, currently available at `yast2-caasp`, should be able to be used for whatever role.

*Short description of the original problem.*

- http://bugzilla.suse.com/show_bug.cgi?id=1129095
- https://trello.com/c/8uqCahT2/875-3-make-yast2caasp-kubeadm-ntp-screen-more-genericly-available

## Solution

_Move_ the dialog to `yast2-installation`.

## Testing

- Also _moved_ the unit test available in `yast2-caasp`
- Tested manually in TW and Kubic

## Screenshots
 * TW
  
  | NTP Setup | Validation |
  |-----------|------------|
  |Screenshot_test_2019-03-28_22:52:27|Screenshot_test_2019-03-28_22:52:45|
   

  
  * KUBIC
  
  | NTP Setup | Validation | Help |
  |-----------|------------|-----|
  |Screenshot_test_2019-03-29_10:47:27|Screenshot_test_2019-03-29_10:48:05|Screenshot_test_2019-03-29_10:47:50|

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
 FROM yastdevel/ruby
+RUN zypper --non-interactive in --no-recommends yast2-ntp-client
 COPY . /usr/src/app
 
--- a/src/lib/installation/dialogs/ntp_setup.rb
+++ b/src/lib/installation/dialogs/ntp_setup.rb
@@ -28,7 +28,7 @@ module Installation
     # Simple dialog for settings the NTP server names
     class NtpSetup < CWM::Dialog
       def initialize
-        textdomain "caasp"
+        textdomain "installation"
 
         Yast.import "Lan"
         Yast.import "LanItems"
--- a/src/lib/installation/dialogs/ntp_setup.rb
+++ b/src/lib/installation/dialogs/ntp_setup.rb
@@ -1,0 +1,121 @@
+# encoding: utf-8
+
+# ------------------------------------------------------------------------------
+# Copyright (c) 2019 SUSE LLC
+#
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, contact SUSE.
+#
+# To contact SUSE about this file by physical or electronic mail, you may find
+# current contact information at www.suse.com.
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "cwm/dialog"
+require "installation/widgets/ntp_server"
+
+module Installation
+  module Dialogs
+    # Simple dialog for settings the NTP server names
+    class NtpSetup < CWM::Dialog
+      def initialize
+        textdomain "caasp"
+
+        Yast.import "Lan"
+        Yast.import "LanItems"
+        Yast.import "Product"
+        Yast.import "ProductFeatures"
+
+        super
+      end
+
+      # The dialog title
+      #
+      # @return [String] the title
+      def title
+        # TRANSLATORS: dialog title
+        _("NTP Setup")
+      end
+
+      def contents
+        return @content if @content
+
+        @content = HSquash(
+          MinWidth(
+            50,
+            # preselect the servers from the DHCP response
+            Widgets::NtpServer.new(ntp_servers)
+          )
+        )
+      end
+
+    private
+
+      # Propose the NTP servers from the DHCP response, fallback to a random
+      # machine from the ntp.org pool if enabled in control.xml.
+      #
+      # @return [Array<String>] proposed NTP servers, empty if nothing suitable found
+      def ntp_servers
+        # TODO: use Yast::NtpClient.ntp_conf if configured
+        # to better handle going back
+        servers = dhcp_ntp_servers
+        servers = [ntp_fallback] if servers.empty? && default_ntp_setup_enabled?
+
+        servers
+      end
+
+      # List of NTP servers from DHCP
+      #
+      # @return [Array<String>] List of servers (IP or host names), empty if not provided
+      def dhcp_ntp_servers
+        # When proposing NTP servers we need to know
+        #
+        #   1) list of (dhcp) interfaces
+        #   2) network service in use
+        #
+        # We can either use networking submodule for network service handling and get list of
+        # interfaces e.g. using a bash command or initialize whole networking module.
+        Yast::Lan.ReadWithCacheNoGUI
+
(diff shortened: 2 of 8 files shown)


src/lib/installation/dialogs/ntp_setup.rb
```diff
@@ -0,0 +1,121 @@
...
+require "yast"
+require "cwm/dialog"
+require "installation/widgets/ntp_server"
+
+module Installation
+  module Dialogs
+    # Simple dialog for settings the NTP server names
+    class NtpSetup < CWM::Dialog
+      def initialize
+        textdomain "caasp"
```
dgdavid: Regarding to `textdomain`, I guess it must be changed to _installation_, isn't it? Or is there any reason to keep its current value?

src/lib/installation/dialogs/ntp_setup.rb
```diff
@@ -0,0 +1,121 @@
...
+      end
+
+    private
+
+      # Propose the NTP servers from the DHCP response, fallback to a random
+      # machine from the ntp.org pool if enabled in control.xml.
+      #
+      # @return [Array<String>] proposed NTP servers, empty if nothing suitable found
+      def ntp_servers
+        # TODO: use Yast::NtpClient.ntp_conf if configured
```
dgdavid: I do not understand this `TODO` comment. I mean, it is not clear to me if we can do what it is stated there or simply keep the current behavior and also de `TODO` comment.

@lslezak it seems that you [wrote it originally](https://github.com/yast/yast-caasp/blob/28e3023ca1e58afb544a3e0d844435c0dfce8905/src/lib/y2caasp/clients/admin_role_dialog.rb#L71). Any suggestion?

src/lib/installation/dialogs/ntp_setup.rb
```diff
@@ -0,0 +1,121 @@
...
+      end
+
+      def ntp_host
+        # copied from timezone/dialogs.rb:
+        base_products = Yast::Product.FindBaseProducts
+
+        if base_products.any? { |p| p["name"] =~ /openSUSE/i }
+          "opensuse"
+        else
+          # TODO: use a SUSE server when available in the future
```
dgdavid: Any update about this? Do someone know if there is a `suse` ntp host already available?

src/lib/installation/dialogs/ntp_setup.rb
lslezak: For SP2 we should change it to "installation".

src/lib/installation/dialogs/ntp_setup.rb
lslezak: IIRC that value is set in the time zone configuration (if you go to the details and configure the NTP server there), so there a small duplication.

For now I'd just move it without further changes.

dgdavid: So, keep it as it is, right?

Dockerfile
```diff
@@ -1,3 +1,5 @@
 FROM yastdevel/ruby
+RUN zypper --non-interactive in --no-recommends \
+  yast2-ntp-client
```
jreidinger: if it is needed for build then please add build requires. For now it is dragged in by yast2-country, but it can change in future.

lslezak: Or we could mock the whole module to decrease the build chain...

src/lib/installation/widgets/ntp_server.rb
```diff
@@ -0,0 +1,143 @@
...
+
+      # Check if the user wants to intentionally skip the NTP server configuration
+      #
+      # @return [Boolean] true if user wants to skip it; false otherwise.
+      def skip_ntp_server?
+        Yast::Popup.AnyQuestion(
+          _("NTP Servers"),
+          # TRANSLATORS: error message for invalid ntp server name/address
+          _("You have not configured an NTP server. This may lead to\n" \
+          "your cluster not functioning properly or at all.\n" \
```
jreidinger: I am not sure about wording if it is appropriate for any possible opensuse use case.

Dockerfile
dgdavid: >  For now it is dragged in by yast2-country, but it can change in future

Sorry, just curious. What do you mean?

jreidinger: I mean that now yast2-country depends on yast2-ntp-client. So now it is there because build requires yast2-country. But if we need directly yast2-ntp-client we should not rely that yast2-country depends on yast2-ntp-client as it can change in future.

dgdavid: Thank you for the clarification, Josef.

My doubt came from the absence of `BuildRequires: yast2-ntp-client` in the `yast2-country`'s _.spec_ file. There I only could found a `Conflicts: yast2-ntp-client < 4.1.5`. Do the _Conflicts_ directive an implicite _BuildRequries_? I don not think so, but I did not find too much information.

jreidinger: @dgdavid no, it looks like yast2-country does not depend directly on yast2-ntp-client. That `Conflicts` is here just to ensure that if ntp is there, it is in proper version. So looks like country has some condition and can work without ntp-client installed. So question is now, when there is nothing that drags in ntp-client, does `rake osc:build` work in your branch?

dgdavid: > Conflicts is here just to ensure that if ntp is there, it is in proper version.

That is what I thought

>  does rake osc:build work in your branch?

Nice catch! No, it does not. 

```bash
An error occurred while loading ./test/dialogs/ntp_setup_test.rb.                                                                 
[   12s] Failure/Error: import_pure(mname)
[   12s]
[   12s] NameError:
[   12s]   component cannot import namespace 'NtpClient'
[   12s] # ./src/lib/installation/widgets/ntp_server.rb:31:in `<top (required)>'                                                           
[   12s] # ./src/lib/installation/dialogs/ntp_setup.rb:24:in `<top (required)>'                                                            
[   12s] # ./test/dialogs/ntp_setup_test.rb:6:in `<top (required)>'
```

So, I must to add it as a build requirement dependency, right?

src/lib/installation/widgets/ntp_server.rb
dgdavid: Since this is initially moved from `yast-caasp` to be used by Kubic, do you think we should take care of this right now or it could be postponed until it will be used in other use case? Just asking, I have not strong opinion about when we should address it.

Dockerfile
jreidinger: yes, add it as BuildRequire.

src/lib/installation/widgets/ntp_server.rb
jreidinger: well, add there at least TODO: so we are aware of it.

lslezak: Ah, good point, this dialog was originally designed for Kubic/CaaSP. For generic usage we need some more generic message.

lslezak: Something like 
```suggestion
          "your system not functioning properly.\n" \
```
???

PR #794, opened 2019-04-04
Removed BuildRequires: yast2-ntp-client (causing dependency cycle)

lslezak:
- `BuildRequires: yast2-ntp-client` caused a dependency cycle and must be removed.
- It was needed only for the unit tests, can be replaced by a stub module
- The disadvantage is that RSpec then cannot verify if the mocked methods really exist, but there is probably no better and simple solution


dgdavid:
Related to https://github.com/yast/yast-installation/pull/791

PR #796, opened 2019-05-08
BSC#1132915 - Bootparameter url file:///PATH will not be evaluated co…

varkoly:
…rrect.

Rewriting bash magic in ruby.


varkoly:
Solved in 798.

PR #797, opened 2019-05-10
Sle 15 sp1 bsc 1132915

schubi2:
Has also been tested by the reporter.

### DIFF
--- a/src/lib/transfer/file_from_url.rb
+++ b/src/lib/transfer/file_from_url.rb
@@ -219,17 +219,15 @@ module Yast::Transfer
 
           if Installation.boot == "cd" && !cdrom_device.empty?
             mtab =  File.read("/proc/mounts")
-            if mtab.match(Regexp.new("^#{cdrom_device}"))
+            m = mtab.match(/^#{cdrom_device}\s+(\S+)/)
+            if m
               Builtins.y2milestone(
                 "%1 is already mounted, trying to bind mount...",
                 cdrom_device
               )
-              cmd = "/bin/mount -v --bind "
-              cd_line = mtab.split("\n").find { |line| line.split[0] == cdrom_device }
-              cmd << cd_line.split[1] + " " + mount_point
-
               am1 = Convert.to_map(
-                SCR.Execute(path(".target.bash_output"), cmd)
+                SCR.Execute(path(".target.bash_output"),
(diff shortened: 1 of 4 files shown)


src/lib/transfer/file_from_url.rb
```diff
@@ -216,34 +216,18 @@ def get_file_from_url(scheme:, host:, urlpath:, localfile:,
...
-            )
 
-            if Ops.get_integer(am, "exit", -1) == 0 &&
-                Ops.greater_than(
-                  Builtins.size(Ops.get_string(am, "stdout", "")),
-                  0
-                )
+          if Installation.boot == "cd" && !cdrom_device.empty?
+            mtab =  File.read("/proc/mounts")
+            if mtab.include?(cdrom_device)
```
wfeldt: Wouldn't a regexp that at least guarantees that `mtab.split().find` below finds something be better? `include?` does
only a substring match and that could be anything.

src/lib/transfer/file_from_url.rb
```diff
@@ -216,34 +216,18 @@ def get_file_from_url(scheme:, host:, urlpath:, localfile:,
...
               Builtins.y2warning(
```
wfeldt: Not part of the patch, but anyway: why a warning? This is imho a normal situation - so `y2milestone`.

src/lib/transfer/file_from_url.rb
```diff
@@ -216,34 +216,18 @@ def get_file_from_url(scheme:, host:, urlpath:, localfile:,
...
 
-            if Ops.get_integer(am, "exit", -1) == 0 &&
-                Ops.greater_than(
-                  Builtins.size(Ops.get_string(am, "stdout", "")),
-                  0
-                )
-              Builtins.y2warning(
+          if Installation.boot == "cd" && !cdrom_device.empty?
+            mtab =  File.read("/proc/mounts")
+            if mtab.match(Regexp.new("^#{cdrom_device}"))
```
wfeldt: I'd add at least a space: `Regexp.new("^#{cdrom_device}\s")`. Why not
```
m = mtab.match(Regexp.new("^#{cdrom_device}\s+(\S+)"))
```
where `m[1]` then is the mountpoint and get rid of this split and find business?

wfeldt:
changelog?

PR #801, opened 2019-06-17
Copy Multipath config files

joseivanlopez:
## Problem

When using the multipath optioin *user_friendly_names* during installation, the following config files must also be copied to the installed system: */etc/multipath.conf* and */etc/multipath/bindings*. Otherwise, names persistence is not guarantee.

* https://www.suse.com/support/kb/doc/?id=7001133

Needed for:
* https://bugzilla.suse.com/show_bug.cgi?id=1133045
* https://trello.com/c/3re8LvvF/1049-autoyast-hw-probing-behavior-vs-scripts

## Solution

Client *inst_files_finish* has been adapted to copy multipath config files into the target system.

### DIFF
--- a/src/lib/installation/clients/copy_files_finish.rb
+++ b/src/lib/installation/clients/copy_files_finish.rb
@@ -251,20 +251,21 @@ module Yast
       )
     end
 
+    # See {https://www.suse.com/support/kb/doc/?id=7001133}
+    MULTIPATH_CONFIG_FILES = [
+      "/etc/multipath.conf",
+      "/etc/multipath/bindings",
+      "/etc/multipath/wwids"
+    ].freeze
+
+    private_constant :MULTIPATH_CONFIG_FILES
+
     def copy_multipath
       # Only in install, as update should keep its old config
       return unless Mode.installation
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/copy_files_finish.rb
```diff
@@ -252,16 +252,26 @@ def copy_vnc
...
-      multipath_config = "/etc/multipath/wwids"
-      if File.exist?(multipath_config)
-        log.info "Copying multipath blacklist '#{multipath_config}'"
-        target_path = File.join(Installation.destdir, multipath_config)
-        ::FileUtils.mkdir_p(File.dirname(target_path))
-        ::FileUtils.cp(multipath_config, target_path)
+      # Copy multipath stuff (bnc#885628, bsc#1133045)
+      #
+      # See {https://www.suse.com/support/kb/doc/?id=7001133}
+      multipath_config_files = [
```
lslezak: NP: This is a constant list so it should be a constant, i.e. `MULTIPATH_CONFIG_FILES`.

PR #807, opened 2019-08-02
 Fix the forced base product selection (3/3)

dgdavid:
## Problem

Recently, the `select_product` option was added to the control file to allow **forcing** the base product selection when having multiple products in a single repository (see https://github.com/yast/yast-installation/pull/790).

But seems that the feature was not fully implemented and the installer only avoids displaying the product selector, but not marking the forced product as selected, raising an internal error when trying to access to `#installation_package`.

* https://bugzilla.suse.com/show_bug.cgi?id=1143943
* https://trello.com/c/YuiJfYBt/1201-ostumbleweed-p1-1143943-build-20190801-undefined-method-installationpackage

## Solution

To select the **foced** product properly.

:warning: **Requires `yast2 4.2.17` - https://github.com/yast/yast-yast2/pull/954**

## Additional notes

See https://github.com/yast/yast-yast2/pull/954 for more information and screenshots.

### DIFF
--- a/src/lib/installation/clients/inst_complex_welcome.rb
+++ b/src/lib/installation/clients/inst_complex_welcome.rb
@@ -141,30 +141,20 @@ module Yast
 
     # Return the list of base products available
     #
-    # The list of base products could be
+    # When a base product is being forced, the list will contains only it.
     #
-    #   - Fixed to a single product: which will happen when the base product is
-    #     being forced through the control.xml. Then, only its license will be
-    #     show, nor the product selector.
-    #   - Empty: in update mode, when there are more than 1 product, this method
-    #     will return an empty list because the dialog will not show the license (we
-    #     do not know which product we are upgrading yet) nor the product selector
-    #     (as you cannot change the product during upgrade).
-    #   - Complete: containing all whe available base products.
+    # In update mode, when there are more than 1 product, this method will return an empty
+    # list because the dialog will not show the license (we do not know which product we are
+    # upgrading yet) nor the product selector (as you cannot change the product during upgrade).
     #
-    # @return [Array<Y2Packager::Product>] List of available base products;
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -141,48 +141,30 @@ def setup_final_choice
...
-      end
-      discarded_base_products = available_base_products - filtered_base_products
-
-      if !discarded_base_products.empty?
-        log.info("Ignoring the other available products: #{discarded_base_products.map(&:name)}")
-      end
-
-      filtered_base_products
+      @products =
+        if Y2Packager::Product.forced_base_product
```
lslezak: This changes the semantics a bit, originally the code returned empty list in update mode even when `preselected_base_product` was defined.

dgdavid: Good catch! Let me fix it!

dgdavid: Discussed in IRC:

> **\<dgdavid>** lslezak: but if there are more than one product but the control file is forcing to use only one... is almost the same that "there is only one product", isn't it?
**\<dgdavid>** lslezak: basically the same that is happening during the installation
<dgdavid> lslezak: unless we're only forcing the product for the installation, which seems to be what do you mean :)
**\<lslezak>** dgdavid: it depends on your expectations, we have the OSS repository providing both TW and Kubic products, if you install the system using the Kubic medium it will install Kubic from the repo (as defined in the control.xml),. Now if you boot the TW medium and select upgrade, what should it upgrade to?  To the newer Kubic (because Kubic was already installed) or switch to the TW (as specified in control.xml)?
**\<lslezak>** dgdavid: I'm not sure which is the correct option, but definitely we should not change the behavior unless there is a strong reason for that

PR #815, opened 2019-09-20
Support for the online installation medium (jsc#SLE-7214)

lslezak:
- Adjust the product selection in the welcome screen for the online medium installation
  - The products are read from the `control.xml` file
  - Updated selecting the product
- https://jira.suse.com/browse/SLE-7214
- Tested manually
- 4.2.14

### DIFF
--- a/src/lib/installation/clients/inst_complex_welcome.rb
+++ b/src/lib/installation/clients/inst_complex_welcome.rb
@@ -19,11 +19,15 @@
 # current contact information at www.novell.com.
 # ------------------------------------------------------------------------------
 
-require "yaml"
 require "fileutils"
+require "pp"
+require "yaml"
 require "yast"
+
 require "installation/dialogs/complex_welcome"
+require "y2packager/medium_type"
 require "y2packager/product"
+require "y2packager/product_control_product"
 
 Yast.import "Console"
 Yast.import "FileUtils"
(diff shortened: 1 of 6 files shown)


src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -19,11 +19,15 @@
 # current contact information at www.novell.com.
 # ------------------------------------------------------------------------------
 
-require "yaml"
 require "fileutils"
+require "pp"
```
jreidinger: why it is needed?

src/lib/installation/widgets/product_selector.rb
```diff
@@ -79,6 +86,10 @@ def registered?
       rescue LoadError
         false
       end
+
+      def product_id(prod)
```
jreidinger: hmm, can we extend prod to provide this method? For me it does not belong here, but to prod object itself.

src/lib/installation/clients/inst_complex_welcome.rb
lslezak: To better format the logged data:
```ruby
log.info "Found base products in the control.xml: #{@available_base_products.pretty_inspect}"
```

src/lib/installation/widgets/product_selector.rb
lslezak: It's actually a widget ID, it not related to the products at all, it's used for identifying the RadioButton widgets. So I'd leave it here. I just renamed it to `widget_id` to make it clear.

lslezak: ah, with "widget_id" it fails, that's already used by CWM, renaming to "item_id"...

PR #816, opened 2019-09-20
Added missing textdomains

mvidner:
- https://trello.com/c/mg7PFhw5


mvidner:
@lslezak Thanks! I have checked that master already declares the textdomains. The CI in master is taking care of that too.

PR #819, opened 2019-10-01
Support for the offline installation medium (jsc#SLE-7101)

lslezak:
- Adapt the product selection and initialization for the offline installation medium
- Do not allow changing the selected base product when going back (we would need to remove the previous base product repository which might be tricky if some addons have already been added)
- https://jira.suse.com/browse/SLE-7101
- 4.2.16

### DIFF
--- a/src/lib/installation/widgets/product_selector.rb
+++ b/src/lib/installation/widgets/product_selector.rb
@@ -1,10 +1,12 @@
 require "yast"
 
+require "y2packager/medium_type"
 require "y2packager/product_control_product"
 
 Yast.import "Pkg"
 Yast.import "Popup"
 Yast.import "AddOnProduct"
+Yast.import "WorkflowManager"
 
 require "cwm/common_widgets"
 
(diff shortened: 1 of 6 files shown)


src/lib/installation/widgets/product_selector.rb
```diff
@@ -48,8 +52,21 @@ def store
...
+        # online product from control.xml
         if @product.is_a?(Y2Packager::ProductControlProduct)
           Y2Packager::ProductControlProduct.selected = @product
+        # offline product from the medium repository
+        elsif @product.is_a?(Y2Packager::ProductLocation)
+          # in offline installation add the repository with the selected base product
+          show_popup = true
+          base_url = Yast::InstURL.installInf2Url("")
+          log_url = Yast::URL.HidePassword(base_url)
+          Yast::Packages.Initialize_StageInitial(show_popup, base_url, log_url, @product.dir)
```
jreidinger: ugh, that API that force you to do log url yourself looks horrible.

lslezak: Yes, ideally the `URL.HidePassword` call should be used inside...

PR #820, opened 2019-10-08
Adapted to new keyboard handling

schubi2:


src/lib/installation/clients/yast_inf_finish.rb
```diff
@@ -122,7 +121,6 @@ def main
         end
 
         Ops.set(@linuxrc, "Language", Language.language)
-        Ops.set(@linuxrc, "Keytable", Keyboard.keymap)
```
schubi2: Second stage will be used by AY only. So normally no input is needed anymore.

PR #821, opened 2019-10-10
Fix nil crash

jreidinger:
https://trello.com/c/oGDFwrPY/1171-2-tw-p2-1139518-tw20190623-internal-error-undefined-method-gsub-for-nilclass


teclator:
BTW, adding a description to the PR .. at least the link to the trello card or bug would be nice to have :)

PR #824, opened 2019-10-31
Decreasing required memory

schubi2:
Replacing old Pkg.ResolvableProperties()and Pkg.ResolvabeDependencies() calls by Y2Packager::Resolvable.any? and Y2Packager::Resolvable.find in order to save memory.

### DIFF
--- a/src/lib/installation/clients/inst_addon_update_sources.rb
+++ b/src/lib/installation/clients/inst_addon_update_sources.rb
@@ -27,6 +27,9 @@
 # Assumptions:
 # - the sources will be saved afterwards
 # (this means that running this client alone will not work)
+
+require "y2packager/resolvable"
+
 module Yast
   class InstAddonUpdateSourcesClient < Client
     def main
@@ -152,10 +155,10 @@ module Yast
     # @return the installation sources to be added
     def UpdateUrls
       urls = {}
-      products = Pkg.ResolvableProperties("", :product, "")
+      products = Y2Packager::Resolvable.find(kind: :product)
 
       Builtins.foreach(products) do |p|
-        Builtins.foreach(Ops.get_list(p, "update_urls", [])) do |u|
+        Builtins.foreach((p.update_urls || [])) do |u|
           # bnc #542792
           # Repository name must be generated from product details
           Ops.set(
@@ -163,16 +166,8 @@ module Yast
             u,
             Builtins.sformat(
               _("Updates for %1 %2"),
-              Ops.get_locale(
-                p,
-                "display_name",
-                Ops.get_locale(
-                  p,
-                  "name",
-                  Ops.get_locale(p, "summary", _("Unknown Product"))
-                )
-              ),
-              Ops.get_string(p, "version", "")
+              (p.display_name || p.name || p.summary || _("Unknown Product")),
+              p.version
             )
           )
           # alias should be simple (bnc#768624)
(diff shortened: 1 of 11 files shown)


src/modules/ImageInstallation.rb
```diff
@@ -1273,28 +1273,28 @@ def ProceedWithSelected(one_object, one_type)
...
-          resolvable_properties
+          resolvable_properties.map(&:name)
         )
       end
 
       # Leave only already installed (and matching the same architecture)
       resolvable_properties = Builtins.filter(resolvable_properties) do |one_resolvable|
-        Ops.get_symbol(one_resolvable, "status", :unknown) == :installed &&
-          Ops.get_string(one_resolvable, "arch", "") == arch
+        one_resolvable.status == :installed &&
```
lslezak: I'd move this filtering directly to the `Resolvable.find` arguments above.

src/modules/ImageInstallation.rb
```diff
@@ -1273,28 +1273,28 @@ def ProceedWithSelected(one_object, one_type)
...
+          one_resolvable.arch == arch
       end
 
       if ThisIsADebugMode()
-        Builtins.y2milestone("Resolvables installed: %1", resolvable_properties)
+        Builtins.y2milestone("Resolvables installed: %1", resolvable_properties.map(&:name))
```
lslezak: I'd completely remove `ThisIsADebugMode()` and use `log.debug` here.

src/modules/ImageInstallation.rb
```diff
@@ -1349,43 +1349,41 @@ def RestoreAllChanges
       @generic_set_progress.call(id, 0) if !@generic_set_progress.nil?
 
       Builtins.foreach(@all_supported_types) do |one_type|
```
lslezak: NP I was thinking about a complete refactoring of this but as we do not install images anymore we could probably drop it. And that's out of scope here...

src/lib/installation/clients/inst_addon_update_sources.rb
```diff
@@ -152,41 +155,32 @@ def KnownUrls
...
-      products = Pkg.ResolvableProperties("", :product, "")
+      products = Y2Packager::Resolvable.find(kind: :product)
 
       Builtins.foreach(products) do |p|
-        Builtins.foreach(Ops.get_list(p, "update_urls", [])) do |u|
+        Builtins.foreach(p.update_urls) do |u|
           # bnc #542792
           # Repository name must be generated from product details
+          p_name =
+            [p.display_name, p.name, p.summary, _("Unknown Product")].reject(&:empty?).first
```
imobachgs: I am starting to think that we should extend [Y2Packager::Resolvable](https://github.com/yast/yast-packager/blob/a58eb3ccda478a5908fb7d1e3bf4c65b64ce1bd6/src/lib/y2packager/resolvable.rb) with a `#human_name` or something similar.

src/lib/installation/clients/inst_addon_update_sources.rb
```diff
@@ -152,41 +155,32 @@ def KnownUrls
...
-                  Ops.get_locale(p, "summary", _("Unknown Product"))
-                )
-              ),
-              Ops.get_string(p, "version", "")
+              p_name,
+              p.version
             )
           )
           # alias should be simple (bnc#768624)
+          p_alias = [p.display_name, p.name, "repo"].reject(&:empty?).first
```
imobachgs: ... and an `#alias` method :)

PR #825, opened 2019-11-06
Save and restore properly the proposal vertical scroll

dgdavid:
## The problem

The proposal installation summary back to the top after performing any change, not preserving the scroll.

* Related to the behavior described at https://bugzilla.suse.com/show_bug.cgi?id=1143558
* https://trello.com/c/ZocPUtvz/1330-3-use-save-restore-scrolling-in-installation-proposal

## Why?

The proposal summary, which is being held by a `RichText` widget, is re-drawn after making a change. So, it's actually new content.

## The solution

To use [the new `:VScrollValue` property](https://github.com/libyui/libyui/pull/149) added to the  [libyui](https://github.com/libyui/libyui) RichText widget as part of the [YaST UI improvements](https://gist.github.com/shundhammer/1bc79af7d1a31892bd977678bc2af2c1). Taking into account that the content could be longer or shorter after each change, the solution is not completely perfect but it's perfect enough for most of the time.

## Test

Tested manually via driver update in both, Qt and ncurse.

## Screencasts

<details>
<summary>Before (toggle)</summary>

<hr/>

vscroll_not_restored

</details>

<details>
<summary>After (toggle)</summary>

<hr/>

vscroll_restored

</details>

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Nov 14 09:56:21 UTC 2019 - David Diaz <dgonzalez@suse.com>
+
+- Save and restore the proposal vertical scroll (related to a
+  problem reseting the scroll position reported in bsc#1143558)
+- 4.2.22
+
+-------------------------------------------------------------------
 Sun Nov 10 18:56:53 UTC 2019 - David Diaz <dgonzalez@suse.com>
 
 - Improve the role selection dialog using a new scrollable widget
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-installation
-Version:        4.2.21
+Version:        4.2.22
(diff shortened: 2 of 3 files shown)


shundhammer:
Oops; one thing left to check: Does yast-installation require the libyui version that introduces this property?

dgdavid:
> Oops; one thing left to check: Does yast-installation require the libyui version that introduces this property?

Yes, it does.

[There is a conflict with `libyui` < 3.8.2](https://github.com/yast/yast-installation/blob/61db415548ff950315c79907ac6582904b206019/package/yast2-installation.spec#L98) and this PR actually _requires_ a libyui > 3.7.0. In fact, [originally the conflict was declared](https://github.com/yast/yast-installation/compare/81c1181b1915831e142a8ae95a029594134170b0..a1b2da05cefb1cdb92c6e0ab23976047a6253219#diff-e8dd2765acbfec5b084347f6e1248121L99), but #827 introduced a _higher_ conflict. so...

PR #826, opened 2019-11-07
Implement upgrade for Full medium (jsc#SLE-7101)

lslezak:
- https://jira.suse.com/browse/SLE-7101
- https://trello.com/c/URPtTjGS/1365-5-system-upgrade-using-the-offline-media
- Skip the repository initialization at upgrade also for the offline medium
- 4.2.20

### DIFF
--- a/src/lib/installation/clients/inst_system_analysis.rb
+++ b/src/lib/installation/clients/inst_system_analysis.rb
@@ -108,7 +108,9 @@ module Yast
       actions_doing     << _("Searching for system files...")
       actions_functions << fun_ref(method(:FilesFromOlderSystems), "boolean ()")
 
-      skip_software = Mode.update && Y2Packager::MediumType.online?
+      # skip the repository initialization on both Online and Full medium,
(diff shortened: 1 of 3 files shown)


src/lib/installation/clients/inst_system_analysis.rb
```diff
@@ -108,7 +108,9 @@ def main
       actions_doing     << _("Searching for system files...")
       actions_functions << fun_ref(method(:FilesFromOlderSystems), "boolean ()")
 
-      skip_software = Mode.update && Y2Packager::MediumType.online?
+      # skip the repository initialization on both Online and Full medium,
+      # do it only on the standard medium layout (openSUSE Leap)
+      skip_software = Mode.update && !Y2Packager::MediumType.standard?
```
jreidinger: will it work for full common installation which now doing it here?

lslezak: There's that `Mode.update` condition...

PR #827, opened 2019-11-07
Improve the role selection dialog

dgdavid:
## Context

Some time ago, the list of available roles grown enough to start not fitting in the role selection dialog. To fix it, the list was embedded in a scrollable `RichText` widget since [libyui](https://github.com/libyui/libyui) didn't offer a better option (see https://github.com/yast/yast-installation/pull/671, [bsc#1084674](https://bugzilla.suse.com/show_bug.cgi?id=1084674), [bsc#1086187](https://bugzilla.suse.com/show_bug.cgi?id=1086187)).

## Current status

As part of a [YaST UI improvements](https://gist.github.com/shundhammer/1bc79af7d1a31892bd977678bc2af2c1) a new `SingleItemSelector` (see https://github.com/libyui/libyui/pull/148) has been added, which is a much better solution to display the roles list.

## Changes

To update the role selection dialog using the new widget and removing the old code.

## Tests

* Unit tests updated/adapted
* Also tested manually via driver update.

## Screenshots

<details>
<summary>Toggle</summary>

<hr />

<p align="center">Before</p>

| ncurses | Qt |
|---------|----|
| Current role selection dialog ncurse | Current role selection dialog |

| ncurses 80x25 | ncurses 80x25 scrolling |
|---------|----|
| ncurses 80x25 | ncurses 80x25 scrolling |

<hr />

<p align="center">After</p>

| ncurses | Qt |
|---------|----|
| Selection role dialog using the new widget ncurses| Selection role dialog using the new widget |

| ncurses 80x25 | ncurses 80x25 scrolling |
|---------|----|
| ncurses 80x25 | ncurses 80x25 scrolling|

| Qt 1920x1080 | Qt 2560x1600 |
|---------|----|
| Qt 1920x1080 | Qt 2560x1600 |

### DIFF
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -17,12 +17,10 @@
 #  To contact SUSE about this file by physical or electronic mail,
 #  you may find current contact information at www.suse.com
 
-require "cgi"
 require "yast"
 require "ui/installation_dialog"
 require "installation/services"
 require "installation/system_role"
-require "ui/text_helpers"
 
 Yast.import "GetInstArgs"
 Yast.import "Packages"
@@ -33,8 +31,6 @@ Yast.import "ProductFeatures"
 
 module Installation
   class SelectSystemRole < ::UI::InstallationDialog
-    include UI::TextHelpers
-
     NON_OVERLAY_ATTRIBUTES = [
       "additional_dialogs",
       "id",
@@ -113,19 +109,6 @@ module Installation
       else
         finish_dialog(result)
       end
-    end
-
-    # called if a specific FOO_handler is not defined
-    def handle_event(id)
-      role = SystemRole.find(id)
-      if role.nil?
-        log.info "Not a role: #{id.inspect}, skipping"
-        return
-      end
-
-      @selected_role_id = id
-      Yast::UI.ReplaceWidget(Id(:rp), role_buttons(selected_role_id: id))
-      Yast::UI.SetFocus(Id(:roles_richtext))
     end
 
   private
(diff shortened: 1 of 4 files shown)


src/lib/installation/select_system_role.rb
```diff
@@ -108,20 +111,30 @@ def next_handler
...
+    #
+    # @param preselected_role_id [String, nil] the id of the role that should be selected
+    # @return [Array<Item>] collection of role items
+    def roles_items(preselected_role_id)
+      roles.map do |role|
+        Item(
+          Id(role.id),
+          role.label,
+          # Keep the description's line lenght short enough to avoid horizontal scroll in 80x24
+          role.description.strip.scan(/.{1,70}\W/).join("\n"),
```
dgdavid: Despite the result looks good to me, I'm not completely sure about doing this or simply to rely in the description as it is provided (which currently produce horizontal scroll even in high resolutions).

What do you think?

src/lib/installation/select_system_role.rb
```diff
@@ -108,20 +111,30 @@ def next_handler
...
+          role.id == preselected_role_id
+        )
+      end
     end
 
-  private
+    # Return the curren selected role id
```
shundhammer: curren**t**

src/lib/installation/select_system_role.rb
shundhammer: IMHO it is perfectly suitable in the NCurses case (and quite a good idea, actually), but in Qt that forced line break looks odd to me, in particular in that large resolution in your screenshot.

shundhammer:
Technically, LGTM. Just the one typo in the comment.
We might think about the forced line break; it makes perfect sense in NCurses, but I am not so sure about Qt. But that's more the icing on the cake than a deal breaker.
I'd also like to have @lslezak look at this; he might have more input.

dgdavid:
Thanks so much for the review!

> Technically, LGTM. Just the one typo in the comment.
> We might think about the forced line break; it makes perfect sense in NCurses, but I am not so sure about Qt.

The thing is that some role description also force the horizontal scroll in Qt. I'll take an screenshot as soon as possible.

> But that's more the icing on the cake than a deal breaker.
I'd also like to have @lslezak look at this; he might have more input.

Perfect!

src/lib/installation/select_system_role.rb
dgdavid: [As I said](https://github.com/yast/yast-installation/pull/827#issuecomment-552563115), the horizontal scroll is also visible in Qt.

Screenshot_sle15-sp1_2019-11-12_09:36:08

Screenshot_sle15-sp1_2019-11-12_09:36:15


Maybe it should be _fixed_ in the role description itself, but I'm not sure.

dgdavid:
> I'll take an screenshot as soon as possible.

See [the comment above](https://github.com/yast/yast-installation/pull/827#discussion_r345100145)

src/lib/installation/select_system_role.rb
dgdavid: An here how it looks with the previous implementation (using the replaced RichText widget)

Screenshot_sle15-sp1_2019-11-12_09:50:51


Taking the @shundhammer's words borrowed from the IRC, this is because

> those description texts were quite obviously made with HTML (our RichText) in mind
> which does automatic line break

dgdavid: So, as agreed in a chat, the best solution, for now, is to set two different limits for the line break: the current one for ncurse and another longer for Qt.

src/lib/installation/select_system_role.rb
```diff
@@ -108,20 +111,39 @@ def next_handler
...
-      Yast::UI.ReplaceWidget(Id(:rp), role_buttons(selected_role_id: id))
-      Yast::UI.SetFocus(Id(:roles_richtext))
+    # Split the given text into several lines if it exceeds the max length set
+    #
+    # @param text [String]
+    # @return [String]
+    def adjust_text(text)
+      max_line_length = Yast::UI.TextMode ? TEXT_MODE_MAX_LINE_LENGTH : MAX_LINE_LENGTH
+
+      text.strip.scan(/(.{1,#{max_line_length}})(?:\s|$)/).join("\n")
```
dgdavid: JFYI, I also tried

```ruby
text.strip.gsub(/(.{1,#{max_line_length}})(?:\s|$)/, "\\1\n").chomp
```

but a benchmark reveal that _`scan` and `join`_ allocates less memory than _`gsub` and `chomp`_. Exactly `1.11x` less, although it will almost priceless for this use case...

dgdavid:
@shundhammer, @lslezak 

I re-tested it with the new changes (basically [different max line lengths for textmode](#discussion_r345142659)) and updated the most relevant screenshots in the PR description.

So, It's ready for a re-review, please.

shundhammer:
Still LGTM from my side. @lslezak ?

src/lib/installation/select_system_role.rb
jreidinger: hint, you can use existing code https://github.com/yast/yast-yast2/blob/66bc64d08c6ff017b88e3cff63fd02d78bbcf7f1/library/general/src/lib/ui/text_helpers.rb#L31

lslezak: I just noticed that this does not work well if a word is longer than the max. line length:

```ruby
max_line_length = 10
text = "123456789012 123456789012 123456789012 123456789012"
puts text.strip.scan(/(.{1,#{max_line_length}})(?:\s|$)/).join("\n")
```

results in 

```
3456789012
3456789012
3456789012
3456789012
```

i.e. some part is lost. But as using a word longer than at least 70 characters is not very likely we can live with that. But maybe it's worth adding a comment about this...

lslezak:
Oops, the screenshots are already there... :man_facepalming:

dgdavid:
> I just noticed that this does not work well if a word is longer than the max. line length:

Yep, I was trying to solve this problem in different ways, but as you said

> using a word longer than at least 70 characters is not very likely

reason why I decided to not invest much time on it. And, for the same reason, I don't want to solve this with a loop...

dgdavid:
> Yep, I was trying to solve this problem in different ways, but as you said
>
> > using a word longer than at least 70 characters is not very likely
> 
> reason why I decided to not invest much time on it. And, for the same reason, I don't want to solve this with a loop...

That said, I didn't know (my fault, for not asking or taken a look) about the helper pointed out by Josef (see [his comment](https://github.com/yast/yast-installation/pull/827#discussion_r346174567))

So, I'm going to update the code to use it.

src/lib/installation/select_system_role.rb
dgdavid: Ops! I didn't know about it. And... I didn't took a look around :sweat: 

Thanks!

dgdavid:
:tired_face: :tired_face: [`wrap_text`](https://github.com/yast/yast-installation/pull/827/commits/dca69228c45c2db9f096f42659aa4db4b0abd57b) helper don't respect the carriage returns inside the text :disappointed: 

using wrap_text

dgdavid:
> :tired_face: :tired_face: [`wrap_text`](https://github.com/yast/yast-installation/pull/827/commits/dca69228c45c2db9f096f42659aa4db4b0abd57b) helper don't respect the carriage returns inside the text :disappointed: 

Anyway, as we agreed in IRC, I'm goint to merge this, then continue with #825 and finally improve the helper to support this use case.

PR #829, opened 2020-01-14
Rc drop

jreidinger:
https://trello.com/c/Cx9X2roJ/1472-3-get-rid-of-etc-initd-and-sbin-rc-calls


jreidinger:
manual testing works fine ( of course not all possible scenario can be tested ), so merging.

PR #833, opened 2020-01-29
Old repositories

lslezak:
## Details

- Better work with the old repositories and services
- Related to
  - https://bugzilla.suse.com/show_bug.cgi?id=1159433
  - https://github.com/yast/yast-yast2/pull/1010
  - https://github.com/yast/yast-packager/pull/502
  - https://github.com/yast/yast-update/pull/147

## Testing

- Added unit tests
- Tested manually

## Notes

- The `clients/inst_upgrade_urls.rb` file has been completely reimplemented (1210 deletions :open_mouth:)
  - It is probably easier to review the complete new file instead of the diff ([see here](https://github.com/yast/yast-installation/blob/old_repositories/src/lib/installation/clients/inst_upgrade_urls.rb))
- The old client was able to configure the network if needed, but as we now try to configure the network automatically at the beginning this is not needed and that functionality has been removed (to keep the code simple).
- Besides that the UI should be the same as before, the change is only in the internal implementation (that means it should not break openQA, etc...).

### DIFF
--- a/src/lib/installation/clients/inst_upgrade_urls.rb
+++ b/src/lib/installation/clients/inst_upgrade_urls.rb
@@ -240,7 +240,7 @@ module Yast
 
           selected_repo = repo_manager.repositories.find { |r| r.repo_alias == current_item }
           if !selected_repo
-            log.error("Selected repository not found???")
+            log.error("Selected repository #{current_item.inspect} not found???")
             next
           end
 
--- a/src/lib/installation/clients/inst_upgrade_urls.rb
+++ b/src/lib/installation/clients/inst_upgrade_urls.rb
@@ -64,13 +64,7 @@ module Yast
         return ret
       end
 
-      # just for testing purposes
-      Wizard.CreateDialog if Mode.normal
-
       ret = run_dialog
-
-      # just for testing purpose
-      Wizard.CloseDialog if Mode.normal
 
       log.info("Returning #{ret.inspect}")
       ret
@@ -84,6 +78,9 @@ module Yast
     #
     # @return [Symbol] the UserInput symbol
     def run_dialog
+      # just for testing purposes
+      Wizard.CreateDialog if Mode.normal
+
       display_dialog
       refresh_dialog
 
@@ -92,6 +89,9 @@ module Yast
         repo_manager.activate_changes
         save_pkg_mgr
       end
+
+      # just for testing purposes
+      Wizard.CloseDialog if Mode.normal
 
       ret
     end
@@ -233,12 +233,16 @@ module Yast
         ret = UI.UserInput
         ret = :abort if ret == :cancel
 
-        if ret == :toggle || ret == "table_of_repos" || ret == :edit
+        case ret
+        when :toggle, :edit, "table_of_repos"
           current_item = UI.QueryWidget(Id("table_of_repos"), :CurrentItem)
           next unless current_item
 
           selected_repo = repo_manager.repositories.find { |r| r.repo_alias == current_item }
-          next unless selected_repo
+          if !selected_repo
+            log.error("Selected repository not found???")
+            next
+          end
 
           if ret == :edit
             edit_item(selected_repo)
@@ -247,8 +251,10 @@ module Yast
           end
 
           refresh_dialog
-        elsif ret == :next || ret == :back || (ret == :abort && Popup.ConfirmAbort(:painless))
+        when :next, :back
           break
+        when :abort
+          break if Popup.ConfirmAbort(:painless)
         else
           log.warn("Unknown input: #{ret}")
         end
--- a/src/lib/installation/clients/inst_upgrade_urls.rb
+++ b/src/lib/installation/clients/inst_upgrade_urls.rb
@@ -2,7 +2,7 @@
 
 # ------------------------------------------------------------------------------
 # Copyright (c) 2006-2012 Novell, Inc. All Rights Reserved.
-#
+# Copyright (c) 2020 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of version 2 of the GNU General Public License as published by the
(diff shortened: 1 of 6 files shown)


src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -19,208 +19,81 @@
...
-      @repos_to_add = []
-
-      @id_to_url = {}
-
-      # bnc #308763
-      @repos_to_add_disabled = []
-
-      @urls = []
+      # use ::Installation avoid collision with Yast::Installation
+      @repo_manager = ::Installation::UpgradeRepoManager.create_from_old_repositories
```
jreidinger: yeah, reason why we start using Y2 prefix.

jreidinger: BTW I would move this code to attr_reader.
so something like
```
def repo_manager
  @repo_manager ||= ::Installation::UpgradeRepoManager.create_from_old_repositories
end
```

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -19,208 +19,81 @@
...
-      if Mode.normal
-        Pkg.SourceSaveAll if @ret == :next
-        Wizard.CloseDialog
-      end
-
-      Builtins.y2milestone("Returning %1", @ret)
-      @ret
+      log.info("Returning #{ret.inspect}")
+      ret
     end
```
jreidinger: whole this main method is for complete rewrite too long. What about moving all dialog related stuff including that Open and Close dialog to own method like `run_dialog`

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -269,106 +142,53 @@ def CreateListTableUI
...
-      if currentitem.nil? || Ops.less_than(currentitem, 0)
-        Builtins.y2error("Cannot edit item: %1", currentitem)
-        return
-      end
-
-      url = Ops.get_string(@urls, [currentitem, "url"], "")
-      min_width = Builtins.size(url)
+    def edit_item(repo)
+      url = repo_manager.repo_url(repo)
+      # limit the maximum width
```
jreidinger: does it really do limit maximun width? It looks like minimal possible width to me.

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -387,979 +207,72 @@ def EditItem(currentitem)
...
-        end
-        deep_copy(one_url)
-      end
+      loop do
+        ret = UI.UserInput
+        ret = :abort if ret == :cancel
 
-      nil
-    end
+        if ret == :toggle || ret == "table_of_repos" || ret == :edit
```
jreidinger: I would use case here, as it will be more readable

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -387,979 +207,72 @@ def EditItem(currentitem)
...
+          current_item = UI.QueryWidget(Id("table_of_repos"), :CurrentItem)
+          next unless current_item
 
-    # Function removes repositories already registered by the installation
-    # from list of urls found on the system.
-    def RemoveInstallationReposFromUpgrededSystemOnes
-      # Works only for the very first registered (installation) repo
-      found = false
+          selected_repo = repo_manager.repositories.find { |r| r.repo_alias == current_item }
+          next unless selected_repo
```
jreidinger: this looks quite strange. I think we should use log.error here at least.

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -387,979 +207,72 @@ def EditItem(currentitem)
...
           break
-        elsif ui_ret == :edit
-          currentitem = Convert.to_integer(
-            UI.QueryWidget(Id("table_of_repos"), :CurrentItem)
-          )
-          EditItem(currentitem)
-          RedrawListTableUI()
         else
-          Builtins.y2error("Unknown UI input: %1", ui_ret)
+          log.warn("Unknown input: #{ret}")
```
jreidinger: why warn instead of error?

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -387,979 +207,72 @@ def EditItem(currentitem)
...
-      elsif Ops.is_boolean?(
-        Ops.get_boolean(repo, @REPO_ENABLED, false)
-      )
-        enabled = Ops.get_boolean(repo, @REPO_ENABLED, false)
-      end
+    # Save the changes to disk, reset the stored old repositories to not display
+    # this dialog again after going back.
+    def save_pkg_mgr
+      # do not save the changes in the test mode
+      Pkg.SourceSaveAll unless ENV["YAST_TEST"]
```
jreidinger: ok, second occurence. Maybe it deserve own method like `testing?`

src/lib/installation/clients/inst_upgrade_urls.rb
lslezak: If you press `Ctrl+Shift+Alt+D` to start the debugger `UI.UserInput` returns `:debugKey` (or something like that). Maybe there are some other cases...

src/lib/installation/upgrade_repo_manager.rb
```diff
@@ -0,0 +1,160 @@
...
+
+require "y2packager/original_repository_setup"
+require "y2packager/repository"
+require "y2packager/service"
+
+Yast.import "Pkg"
+
+module Installation
+  # This class takes care of managing the old repositories and services
+  # during upgrade.
```
jreidinger: I would extend this description to make clear what is different to new OriginalRepositorySetup class in yast2.

src/lib/installation/upgrade_repo_manager.rb
```diff
@@ -0,0 +1,160 @@
...
+      stored_repo_aliases = stored_repos.map(&:repo_alias)
+      old_repos = current_repos.select { |r| stored_repo_aliases.include?(r.repo_alias) }
+
+      current_services = Y2Packager::Service.all
+      stored_services = Y2Packager::OriginalRepositorySetup.instance.services
+      stored_service_aliases = stored_services.map(&:alias)
+      old_services = current_services.select { |s| stored_service_aliases.include?(s.alias) }
+
+      # sort the repositories by name
+      new(old_repos.sort_by(&:name), old_services)
```
jreidinger: why sort only repos and not services?

src/lib/installation/upgrade_repo_manager.rb
```diff
@@ -0,0 +1,160 @@
...
+    # @return [Symbol, nil] `:removed`, `:enabled` or `:disabled` symbol,
+    #    `nil` if the repository is not known
+    def repo_status(repo)
+      status_map[repo]
+    end
+
+    # Return the repository URL possibly new
+    #
+    # @param repo [Y2Packager::Repository] The queried repository
+    # @return [Symbol] <description>
```
jreidinger: ugh, what? I do not get return documentation

src/lib/installation/upgrade_repo_manager.rb
```diff
@@ -0,0 +1,160 @@
...
+    # @return [Symbol, nil] `:removed`, `:enabled` or `:disabled` symbol,
+    #    `nil` if the repository is not known
+    def toggle_repo_status(repo)
+      case repo_status(repo)
+      when :enabled
+        status_map[repo] = :disabled
+      when :disabled
+        status_map[repo] = :removed
+      when :removed
+        status_map[repo] = :enabled
```
jreidinger: another option, which I am not sure if it is more readable is to create transition map like
```
TOGGLE_STATUS = {
  enabled: :disabled,
  disabled: :removed,
  removed: :enabled
}
```
and then just call `status_map[repo] = TOGGLE_STATUS[repo_status(repo)]`

src/lib/installation/upgrade_repo_manager.rb
```diff
@@ -0,0 +1,160 @@
...
+      end
+    end
+
+    # Change the URL of a repository.
+    #
+    # @param repo [Y2Packager::Repository] The repository
+    # @param url [String] Its new URL
+    def change_url(repo, url)
+      new_urls[repo] = url if repo.raw_url.to_s != url
```
jreidinger: what happens if you change url twice and first time to new and the second time for old one?

src/lib/installation/upgrade_repo_manager.rb
```diff
@@ -0,0 +1,160 @@
...
+    end
+
+  private
+
+    # remove the old services
+    def remove_services
+      log.info("Old services to remove: #{services.map(&:alias).inspect}")
+      services.each do |s|
+        log.info("Removing old service #{s.alias}...")
+        Yast::Pkg.ServiceDelete(s.alias)
```
jreidinger: it is not method in service like in Repository class?

src/lib/installation/upgrade_repo_manager.rb
lslezak: We do not display the services at all, everything is silently removed. That's IMHO a bit strange and we should present them somehow in the dialog, but that's how it was done before...

src/lib/installation/upgrade_repo_manager.rb
lslezak: Oh, that's a template generated by VSCode, I forgot to replace that...

src/lib/installation/upgrade_repo_manager.rb
lslezak: I was also thinking about this solution but as there are only three states and it's needed only at a single place I decided this way. If it was needed also elsewhere then `TOGGLE_STATUS` would be better, but currently I do not see any advantage for that...

src/lib/installation/upgrade_repo_manager.rb
lslezak: Um, then it will try to change the URL to the same value which has no effect, it would just do one more useless Pkg call. And as this is quite a corner case I do not think it needs to be handled specially in the code.

src/lib/installation/upgrade_repo_manager.rb
lslezak: Yeah, noticed that too late, I'd need to touch yet another package again... :worried: 
As this is a quite urgent fix (a P1) I decided this simpler solution....

src/lib/installation/upgrade_repo_manager.rb
jreidinger: well, my suspicion from this condition is that it will still be updated to old url, even when user do not want to modify it due to condition here.

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -87,6 +80,22 @@ def main
 
     attr_reader :repo_manager
 
+    # Run the main dialog
+    #
+    # @return [Symbol] the UserInput symbol
+    def run_dialog
+      display_dialog
```
jreidinger: well, I actually expect also OpenDialog and CloseDialog calls inside this method as it looks related to me.

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -233,12 +233,16 @@ def handle_dialog
...
-        if ret == :toggle || ret == "table_of_repos" || ret == :edit
+        case ret
+        when :toggle, :edit, "table_of_repos"
           current_item = UI.QueryWidget(Id("table_of_repos"), :CurrentItem)
           next unless current_item
 
           selected_repo = repo_manager.repositories.find { |r| r.repo_alias == current_item }
-          next unless selected_repo
+          if !selected_repo
+            log.error("Selected repository not found???")
```
jreidinger: NP: what can help here is value of current item and list all repositories for debugging if it happens on customer side.

PR #835, opened 2020-02-14
Added "yupdate" script

lslezak:
### Summary

- Related to https://github.com/yast/yast-rake/pull/76
- https://bugzilla.suse.com/show_bug.cgi?id=1162826

### Problem

- Patching the installer is not easy, building DUD for several packages takes time and is not trivial

### Solution

- Let's have a simple script in the inst-sys called `yupdate` which can download the sources directly from GitHub and install them to the inst-sys
- To make it easier allow patching from a local Git checkout (without need to push every single change to GitHub)
- See more details in the `yupdate.md` documentation file

### Testing

- Tested manually
- Added some unit tests, but only small part is covered (BTW making the code coverage work was a bit tricky...)

### Notes

- As mentioned in the script, the reason why all classes are in a single file is that it will make installing or updating the script with curl/wget much easier.

### DIFF
--- a/bin/yupdate
+++ b/bin/yupdate
@@ -63,7 +63,9 @@ Commands:
                             https://github.com/yast/yast-rake/#server),
                             the default port is 8000
 
-    servers <host[:port]>   List the rake servers running on the remote machine
+    servers <host[:port]>   List the rake servers running on the remote machine,
+                            ask the specified server to report all `rake server`
+                            instances
 
     overlay create [<dir>]  Create a new writable overlay for directory <dir>,
                             if no directory is specified it creates overlays
@@ -188,6 +190,7 @@ HELP
     attr_reader :dir, :orig_dir
 
     # manage the OverlayFS for this directory
+    # @param directory [String] the directory
     def initialize(directory)
       @orig_dir = directory
       # expand symlinks
@@ -223,12 +226,12 @@ HELP
 
     # print the modified files in this directory
     def print_files
-      modified_files { |f, _modif, _orig| puts f }
+      iterate_files { |f, _modif, _orig| puts f }
     end
 
     # print the diff for the changed files in this directory
     def print_diff
-      modified_files do |f, _modif, orig|
+      iterate_files do |f, _modif, orig|
         next unless File.exist?(f) && File.exist?(orig)
         system("diff -u #{orig.shellescape} #{f.shellescape}")
       end
@@ -247,27 +250,36 @@ HELP
       yast_overlays.map { |o| new(o) }
     end
 
-  private
-
-    def dir_name
-      # escape (double) underscores for correct reverse conversion
-      dir.gsub("_", "__").tr("/", "_")
-    end
-
     def upperdir
-      File.join(OVERLAY_PREFIX, "upper", dir_name)
+      OverlayFS.escape_path("upper", dir)
     end
 
     def workdir
-      File.join(OVERLAY_PREFIX, "workdir", dir_name)
+      OverlayFS.escape_path("workdir", dir)
     end
 
     def origdir
-      File.join(OVERLAY_PREFIX, "original", dir_name)
+      OverlayFS.escape_path("original", dir)
     end
 
-    # find the modified files in this directory
-    def modified_files(&block)
+    def self.escape_path(subdir, path)
+      # escape (double) underscores for correct reverse conversion
+      File.join(OVERLAY_PREFIX, subdir, path.gsub("_", "__").tr("/", "_"))
+    end
+
+    def self.unescape_path(path)
+      Pathname.new(path).basename.to_s.gsub(/([^_])_([^_])/, "\\1/\\2")
+              .sub(/\A_/, "/").gsub("__", "_").gsub("//", "/")
+    end
+
+  private
+
+    # find the files in this directory
+    # the block is called with three parameters:
+    # - the path to the new file in the overlay directory
+    # - the original path in /
+    # - the path to the original file in the overlay directory
+    def iterate_files(&block)
       return unless block_given?
 
       Find.find(upperdir) do |f|
@@ -275,9 +287,8 @@ HELP
         upperdir_path = Pathname.new(upperdir)
         relative_path = Pathname.new(f).relative_path_from(upperdir_path)
         original_path = File.join(origdir, relative_path)
-        # unescape _
-        pth = upperdir_path.basename.to_s.gsub(/([^_])_([^_])/, "\\1/\\2")
-                           .sub(/\A_/, "/").gsub("__", "_").gsub("//", "/")
+        # unescape the path
+        pth = unescape_path(upperdir_path)
         block.call(File.join(pth, relative_path), f, original_path)
       end
     end
@@ -349,7 +360,7 @@ HELP
         # pipe the response body directly to the tar process
         IO.popen(["tar", "-C", dir, "--warning=no-timestamp", "-xz"], "wb") do |io|
           response.read_body do |chunk|
-            io.write chunk
+            io.write(chunk)
           end
         end
       end
@@ -563,9 +574,8 @@ HELP
     # the script might not work as expected in an installed system
     # and using OverlayFS is potentially dangerous
     def self.check!
-      mounts = `mount`
-      # the inst-sys uses tmpfs (RAM disk) for the root
-      return if mounts =~ /^tmpfs on \/ type tmpfs/
+      # the inst-sys contains the /.packages.initrd file with a list of packages
+      return if File.exist?("/.packages.initrd")
 
       # exit immediately if running in an installed system
       $stderr.puts "ERROR: This script can only work in the installation system (inst-sys)!"
@@ -612,8 +622,7 @@ HELP
 
     def run
       host = @argv.shift
-
-      return 1 unless host
+      raise "Missing server name argument" unless host
 
       servers = RemoteServer.find(host)
       servers.each do |s|
--- a/bin/yupdate
+++ b/bin/yupdate
@@ -1,0 +1,777 @@
+#!/usr/bin/env ruby
+
+# This script updates the YaST files in the inst-sys with the files from
+# a GitHub repository or a running tarball web server.
+#
+# See the help text below for the details or see the documentation in
+# the doc/yupdate.md file.
+#
+# Note: The reason why all classes are in a single file is that
+#       it allows to easy update the script to the latest version
+#       or add it to an older (not supported) installer with:
+#
+#   curl https://raw.githubusercontent.com/yast/yast-installation/master/bin/yupdate > /usr/bin/yupdate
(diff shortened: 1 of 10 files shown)


mvidner:
OK!
Naturally I compared this to the version in your scripts repo, https://github.com/lslezak/scripts/blob/yupdate_refactoring/yast/yupdate/yupdate

I wonder why you removed the `selfupdate` command? Is it because it is a natural byproduct of using yupdate on yast-installation?

bin/yupdate
```diff
@@ -0,0 +1,777 @@
...
+    PATCH = 0
+
+    STRING = "#{MAJOR}.#{MINOR}.#{PATCH}".freeze
+  end
+
+  # handle the "help" command line option
+  class HelpCommand
+    def run
+      name = File.basename(__FILE__)
+      puts <<-HELP
```
jreidinger: why not use built in ruby CLI parser? like https://ruby-doc.org/stdlib-2.7.0/libdoc/optparse/rdoc/OptionParser.html ? I use it for my log analyzer and works quite well for me.

bin/yupdate
```diff
@@ -0,0 +1,777 @@
...
+This is a helper script for updating the YaST installer in the installation system.
+
+Usage: #{name} <command> <options>
+
+Commands:
+
+    patch <github_repo> <branch>  Patch the installer with the sources from GitHub,
+                            <github_repo> is a repository name (including the organization
+                            or the user name, if missing "yast" is used by default),
+                            <branch> is the Git branch to use
```
jreidinger: is there default branch? like master?

bin/yupdate
```diff
@@ -0,0 +1,777 @@
...
+
+    patch <tarball_url>     Patch the installer with the sources provided by
+                            a generic HTTP server
```
jreidinger: only htto is allowed? not ftp?

bin/yupdate
```diff
@@ -0,0 +1,777 @@
...
+
+    patch <host[:port]>     Patch the installer with the sources provided by
+                            a "rake server" task (for details see
+                            https://github.com/yast/yast-rake/#server),
+                            the default port is 8000
+
+    servers <host[:port]>   List the rake servers running on the remote machine
```
jreidinger: this is a bit confusing terminology. Does it mean that more then one server can sit on same port?

mvidner:
Add `bin/yupdate` to `/.yardopts`?

bin/yupdate
```diff
@@ -0,0 +1,777 @@
...
+    def self.find_all
+      mounts = `mount`
+      mounts.lines.each_with_object([]) do |line, arr|
+        arr << new(Regexp.last_match[1]) if line =~ /^overlay on (.*) type overlay /
+      end
+    end
+
+    # return the default set of YaST overlays
+    def self.default_overlays
+      yast_overlays.map { |o| new(o) }
```
jreidinger: why not cache it? this looks like a lot of objects flying around.

bin/yupdate
```diff
@@ -0,0 +1,777 @@
...
+    def create
+      return if File.writable?(dir)
+      msg "Adding overlay for #{orig_dir}..."
+
+      FileUtils.mkdir_p(upperdir)
+      FileUtils.mkdir_p(workdir)
+      FileUtils.mkdir_p(origdir)
+
+      # make the original content available in a separate directory
+      system("mount --bind #{dir.shellescape} #{origdir.shellescape}")
```
jreidinger: btw cheetah can be used here and without shellescape and also with error handling.

bin/yupdate
```diff
@@ -0,0 +1,777 @@
...
+      return unless block_given?
+
+      Find.find(upperdir) do |f|
+        next unless File.file?(f)
+        upperdir_path = Pathname.new(upperdir)
+        relative_path = Pathname.new(f).relative_path_from(upperdir_path)
+        original_path = File.join(origdir, relative_path)
+        # unescape _
+        pth = upperdir_path.basename.to_s.gsub(/([^_])_([^_])/, "\\1/\\2")
+                           .sub(/\A_/, "/").gsub("__", "_").gsub("//", "/")
```
jreidinger: second sub look like potential issue for stuff like "__abc". It can be done previously as in previous gsub like
`gsub(/([^_]|\A)_([^_]|\z)/, "\\1/\\2")`

bin/yupdate
```diff
@@ -0,0 +1,777 @@
...
+      YAST_OVERLAYS + Dir["/usr/share/YaST2/*"].each_with_object([]) do |f, arr|
+        arr << f if File.directory?(f) && !File.writable?(f)
+      end
+    end
+
+    private_class_method :yast_overlays
+  end
+
+  # a generic HTTP downloader
+  class Downloader
```
jreidinger: hmm, why not use something like open uri? https://ruby-doc.org/stdlib-2.6.3/libdoc/open-uri/rdoc/OpenURI.html

bin/yupdate
```diff
@@ -0,0 +1,777 @@
...
+
+  # inst-sys test
+  class InstSys
+    # check if the script is running in the inst-sys,
+    # the script might not work as expected in an installed system
+    # and using OverlayFS is potentially dangerous
+    def self.check!
+      mounts = `mount`
+      # the inst-sys uses tmpfs (RAM disk) for the root
+      return if mounts =~ /^tmpfs on \/ type tmpfs/
```
jreidinger: how reliable it is ? what about testing instead e.g. `which extend` ?

bin/yupdate
```diff
@@ -0,0 +1,777 @@
...
+  # handle the "servers" command line option
+  class ServersCommand
+    def initialize(argv)
+      @argv = argv
+    end
+
+    def run
+      host = @argv.shift
+
+      return 1 unless host
```
jreidinger: hmm, what? I expect here something like printing help and `exit 1`

bin/yupdate
lslezak: No, I'd like to avoid using any default branch here. Using `master` is OK now, but later after the product gets into the maintenance and `master` becomes the next release it would contain incompatible changes.

Then the default should be something like `SLE-15-SP2`, but there is no way how the script could know that...

To be safe let's require that information from the user (developer).

doc/yupdate.md
```diff
@@ -0,0 +1,258 @@
...
+By default this will use port 8000, if the server uses another port just add
+`:` followed by the port number.
+
+*Note: Make sure the server port is open in the firewall configuration,
+see the [documentation](https://github.com/yast/yast-rake/#server) for
+more details.*
+
+#### Patching Multiple Packages
+
+The `yast patch` command installs the sources from all running `rake server`
```
jreidinger: you mean y2update here, right?

doc/yupdate.md
```diff
@@ -0,0 +1,258 @@
...
+  are also missing in the inst-sys)
+
+## Alternative
+
+1. For all repos, run `rake osc:build`
+2. Collect the resulting RPMs
+3. Run a server, eg. with `ruby -run -e httpd -- -p 8888 .`
+4. Type a loooong boot line to pass them all as DUD=http://....rpm
+   (or write that into a file and use the [info](
+    https://en.opensuse.org/SDB:Linuxrc#p_info) option)
```
jreidinger: I use easier alternative. 1,2,3 is same but as 4 I use mkdud and create single dud file.

lslezak:
> I wonder why you removed the `selfupdate` command? Is it because it is a natural byproduct of using yupdate on yast-installation?

Yes, it can be updated as a part of the yast2-installation package now. I was thinking about it and my conclusion was that yet another self-update way could be confusing so I have removed that.

Moreover it can updated manually with that documented `curl` command.

But if it later turns out that a separate self-update would make sense I can port it from my branch...

bin/yupdate
lslezak: Only http(s) is allowed. But you can run a simple webserver just by using a trivial `ruby -run -e httpd` command, configuring an FTP server is much more complicated. So IMHO it's good enough.

But if there is a need for that it could be added later...

bin/yupdate
lslezak: Um, the `OptionParser` works well if you have **optional** arguments, moreover it requires the (double) dashes. So instead of 

    yupdate patch yast-yast2 master

you would need to write

    yupdate --command patch --repo yast-yast2 --branch master

and we would still need to implement the logic that the "patch" command requires "--branch" option ourselves...

So it would help with the `--help` option but otherwise it would make it more complicated and less user friendly.

bin/yupdate
lslezak: There is no reason for caching, it should be always used only once during the execution. It's used from several places in the code, but that's for a different passed command.

bin/yupdate
lslezak: Because with this code we can pipe the HTTP stream directly to the `tar` process without saving the whole tarball file to the disk. We can unpack the data on the fly.

I'm not sure if open-uri can do this...

bin/yupdate
lslezak: @wfeldt told me that checking for the `/.packages.initrd` file is better, I'll update that...

bin/yupdate
lslezak: No, only one server per port. But you need to specify which of them will be used.

doc/yupdate.md
lslezak: OK, added a `mkdud` note...

bin/yupdate
lslezak: I have updated the description a bit.

lslezak:
> Add `bin/yupdate` to `/.yardopts`?

Well, the classes from `bin/yupdate` cannot be used easily from YaST and I do not want to pollute the YaST documentation with this a bit unrelated script.

bin/yupdate
mvidner: NP: In this case, Cheetah is overkill. The array form of `system` does not call the shell and needs no escaping.

bin/yupdate
mvidner: I would expect open-uri to work. This would be worth checking to simplify yupdate.

bin/yupdate
lslezak: I have added some unit test for escaping/unescaping and that works fine...

bin/yupdate
```diff
@@ -0,0 +1,771 @@
...
+    def iterate_files(&block)
+      return unless block_given?
+
+      Find.find(upperdir) do |f|
+        next unless File.file?(f)
+        upperdir_path = Pathname.new(upperdir)
+        relative_path = Pathname.new(f).relative_path_from(upperdir_path)
+        original_path = File.join(origdir, relative_path)
+        # unescape the path
+        pth = unescape_path(upperdir_path)
```
teclator: @lslezak 

```suggestion
        pth = OverlayFS.unescape_path(upperdir_path)
```

PR #838, opened 2020-02-19
[SLE-7188] Default to SUSE ntp servers

imobachgs:
Use the new `Y2Network::NtpServer` when proposing the default values for NTP servers.

Related PR: https://github.com/yast/yast-network/pull/1039

### DIFF
--- a/src/lib/installation/dialogs/ntp_setup.rb
+++ b/src/lib/installation/dialogs/ntp_setup.rb
@@ -22,6 +22,7 @@
 require "yast"
 require "cwm/dialog"
 require "installation/widgets/ntp_server"
+require "y2network/ntp_server"
 
 module Installation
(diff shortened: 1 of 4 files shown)


src/lib/installation/dialogs/ntp_setup.rb
```diff
@@ -68,7 +69,7 @@ def ntp_servers
         # TODO: use Yast::NtpClient.ntp_conf if configured
         # to better handle going back
         servers = dhcp_ntp_servers
```
teclator: As already commented in yast/yast-network#1039 I would like to move all this logic to a Y2Network::NtpServer#ntp_servers but let it for the future

teclator:
LGTM once rebased and fixed the conflicts

PR #839, opened 2020-02-20
Self-update control file

imobachgs:
Using the self-update is not possible to replace the `control.xml` file unless you craft a special package that contains that file in the root filesystem. This PRs extends the the self-update mechanism to update the `/control.xml` file with `/usr/lib/skelcd/CD1/control.xml`, which is where `skelcd-control-CAASP`, `skelcd-control-Kubic`, `skelcd-control-leanos`, `skelcd-control-MicroOS` and `skelcd-control-openSUSE` put the control file.

### DIFF
--- a/src/lib/installation/updates_manager.rb
+++ b/src/lib/installation/updates_manager.rb
@@ -55,6 +55,9 @@ module Installation
     # Repo is unreachable (name solving issues, etc.).
     class CouldNotProbeRepo < RepoError; end
 
+    # The control file was not updated due to some problem
+    class CouldNotUpdateControlFile < RepoError; end
+
     DRIVER_UPDATES_PATHS = [Pathname("/update"), Pathname("/download")].freeze
 
     # Constructor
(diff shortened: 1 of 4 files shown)


src/lib/installation/updates_manager.rb
```diff
@@ -110,14 +113,30 @@ def add_repository(uri)
...
+  private
+
+    NEW_CONTROL_FILE_PATH = "/usr/lib/skelcd/CD1/control.xml".freeze
+    APPLY_CMD = "/sbin/adddir %<source>s /".freeze
+
+    # Replaces the control file with the the updated one (if it exists)
+    def replace_control_file
+      return unless File.exist?(NEW_CONTROL_FILE_PATH)
+      cmd = format(APPLY_CMD, source: File.dirname(NEW_CONTROL_FILE_PATH))
+      out = Yast::SCR.Execute(Yast::Path.new(".target.bash_output"), cmd)
```
lslezak: I think using [`Yast::Execute.locally!`](https://github.com/yast/yast-yast2/blob/master/library/system/src/lib/yast2/execute.rb#L107) would be better here, `.target.bash_output` normally works in `/mnt`, luckily it's not mounted yet at this point. That call also raises an exception on failure, so no need for extra error handling.

src/lib/installation/updates_manager.rb
```diff
@@ -110,14 +113,30 @@ def add_repository(uri)
...
     def repositories?
       !repositories.empty?
     end
+
+  private
+
+    NEW_CONTROL_FILE_PATH = "/usr/lib/skelcd/CD1/control.xml".freeze
+    APPLY_CMD = "/sbin/adddir %<source>s /".freeze
+
+    # Replaces the control file with the the updated one (if it exists)
```
lslezak: Typo

```suggestion
    # Replaces the control file with the updated one (if it exists)
```

lslezak:
Thanks! Now it looks much better! :+1:

imobachgs:
Thanks for the review!

PR #840, opened 2020-02-27
Show Release Notes for Add-On Products

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1158287

## Trello

https://trello.com/c/vmm7flFk/

## Problem

The release notes for add-on products were not shown any more.

## Fix

Don't use `Product.with_status()` because that uses `all_products()` without the `force_repos` parameter.

In the case of online media, this means that it only uses products from the control file which means you will only get base products.

Now using `all_products()` directly, but _with_ the `force_repos` parameter, so we get all products from all repos; and now we also get SLE-HA, so we can check it for release notes.

## Screenshot

sle-ha-release-notes

### DIFF
--- a/src/lib/installation/clients/inst_download_release_notes.rb
+++ b/src/lib/installation/clients/inst_download_release_notes.rb
@@ -27,6 +27,7 @@
 # $Id$
 
 require "y2packager/product"
+require "y2packager/product_reader"
 
 Yast.import "InstData"
 Yast.import "Pkg"
@@ -100,10 +101,13 @@ module Yast
     def products
       # installed may mean old (before upgrade) in initial stage
       # product may not yet be selected although repo is already added
-      return Y2Packager::Product.with_status(:selected, :installed) unless Stage.initial
-      selected = Y2Packager::Product.with_status(:selected)
+      #
+      # Don't rely on Product.with_status() here, use force_repos (bsc#1158287)
+      all_products = Y2Packager::ProductReader.new.all_products(force_repos: true)
+      return all_products.select { |p| p.status?(:selected, :installed) } unless Stage.initial
+      selected = all_products.select { |p| p.selected? }
       return selected unless selected.empty?
-      Y2Packager::Product.with_status(:available)
+      all_products.select { |p| p.status?(:available) }
     end
 
     # Refresh release notes UI
--- a/src/lib/installation/clients/inst_download_release_notes.rb
+++ b/src/lib/installation/clients/inst_download_release_notes.rb
@@ -107,7 +107,7 @@ module Yast
       # of online media, and that means only base products, no add-ons.
       all_products = Y2Packager::ProductReader.new.all_products(force_repos: true)
       return all_products.select { |p| p.status?(:selected, :installed) } unless Stage.initial
-      selected = all_products.select { |p| p.selected? }
+      selected = all_products.select { |p| p.status?(:selected) }
       return selected unless selected.empty?
       all_products.select { |p| p.status?(:available) }
     end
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/inst_download_release_notes.rb
```diff
@@ -100,10 +101,13 @@ def main
     def products
       # installed may mean old (before upgrade) in initial stage
       # product may not yet be selected although repo is already added
-      return Y2Packager::Product.with_status(:selected, :installed) unless Stage.initial
-      selected = Y2Packager::Product.with_status(:selected)
+      #
+      # Don't rely on Product.with_status() here, use force_repos (bsc#1158287)
```
lslezak: I think we should also comment *why* we cannot use it here, something like "for the online medium it only reads the products from the control file (the base products, no addons)".

test/lib/clients/inst_download_release_notes_test.rb
```diff
@@ -19,13 +20,24 @@
...
+
     let(:textmode) { true }
     let(:packages_init_called) { true }
 
     before do
       allow(Yast::UI).to receive(:TextMode).and_return(textmode)
-      allow(Y2Packager::Product).to receive(:with_status).with(:selected)
-        .and_return([sles, sdk])
+      allow(Y2Packager::ProductReader).to receive(:new).and_return(prod_reader)
+      allow(Y2Packager::ProductReader).to receive(:all_products)
```
imobachgs: `ProductReader` does not implement `.all_products` (it is an instance method). Maybe you wanted to do `allow(prod_reader)...`.

test/lib/clients/inst_download_release_notes_test.rb
```diff
@@ -19,13 +20,24 @@
...
+        .with(force_repos: true).and_return(products))
+      allow(Y2Packager::Product).to receive(:status?).with(:selected)
```
imobachgs: `status?` is an instance method.

test/lib/clients/inst_download_release_notes_test.rb
```diff
@@ -19,13 +20,24 @@
...
     let(:textmode) { true }
     let(:packages_init_called) { true }
 
     before do
       allow(Yast::UI).to receive(:TextMode).and_return(textmode)
-      allow(Y2Packager::Product).to receive(:with_status).with(:selected)
-        .and_return([sles, sdk])
+      allow(Y2Packager::ProductReader).to receive(:new).and_return(prod_reader)
+      allow(Y2Packager::ProductReader).to receive(:all_products)
+        .with(force_repos: true).and_return(products))
```
imobachgs: RuboCop complains about a syntax error here (you just need to remove the last parenthesis).

src/lib/installation/clients/inst_download_release_notes.rb
```diff
@@ -107,7 +107,7 @@ def products
       # of online media, and that means only base products, no add-ons.
       all_products = Y2Packager::ProductReader.new.all_products(force_repos: true)
       return all_products.select { |p| p.status?(:selected, :installed) } unless Stage.initial
-      selected = all_products.select { |p| p.selected? }
+      selected = all_products.select { |p| p.status?(:selected) }
```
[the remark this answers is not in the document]
shundhammer: Yes, I know that rubocop meant that gibberish notation. That's what I was so upset about in IRC. That notation just obfuscates the code.

shundhammer:
Since the unit test relies very much on implementation details and merrily mocks them to fit, it failed at many places after the changes; that's why I had to adapt them in so many places.

PR #841, opened 2020-03-12
Stop plymouth splash to avoid serial terminal splutter.

qiangzhao:
Hi Yast-team:
I have a little modification to
Stop Plymouth animation splash before the installation program run, it will avoid mass up the serial screen's character garbled (bnc#1164123).
Thanks for the review.

### DIFF
--- a/startup/YaST2.First-Stage
+++ b/startup/YaST2.First-Stage
@@ -18,6 +18,11 @@
 #
 #set -x
 #=============================================
+# 0) Stop boot animation.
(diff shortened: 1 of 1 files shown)


wfeldt:
plymouth is already stopped before yast runs:

https://github.com/openSUSE/installation-images/blob/master/data/root/etc/inst_setup#L198

wfeldt:
As commented, this pr is not needed as plymouth gets already stopped.

PR #842, opened 2020-03-13
Added selfupdate verification (bsc#1163084)

lslezak:
## The Problem

- The inst-sys does not contain the RPM DB with the installed packages so the self-update code just blindly downloads and installs everything it finds without any dependency checks.
- YaST gets the self update repository URL either from the SCC or uses the default from the `control.xml`. This way ensures the self-update repository contains the right packages for the installer.
- However, users can use their custom self-update repository (e.g. SMT/RMT) and there is a possibility that by mistake the users could use a wrong repository, e.g. SP1 repository in SP2
- We got even an [L3 bug](https://bugzilla.suse.com/show_bug.cgi?id=1163084) with this mistake

## The Solution

- The inst-sys contains the list of the installed packages in the `/.packages.root` file
- To avoid false positives (as there might be a real package downgrade) we check only some YaST packages because their versions are under our control
- Before applying the self-update we check that the update versions are not older than the packages in the inst-sys
- It is possible to disable this sanity check by setting the environment variable `Y2_FORCE_SELF_UPDATE=1`, then the error is not displayed and the self-updates are applied. But this should be done only in special cases (testing) when you know what you are doing.

## Notes

Originally I also implemented a check for "too new" packages (e.g. using SP3 updates in the SP2 installer), but as this could potentially report some false positives I have removed that. Lets discuss that first...

## Testing 

- Tested manually (see the screenshot below)
- Added unit tests

## Screenshots

Using SP1 updates in the SP2 installer will result in this error message:

self_update_error

### DIFF
--- a/doc/SELF_UPDATE.md
+++ b/doc/SELF_UPDATE.md
@@ -216,6 +216,21 @@ under https://updates.suse.com/
 - /SUSE/Updates/SLE-SERVER-INSTALLER/12-SP2/s390x/update
 - /SUSE/Updates/SLE-SERVER-INSTALLER/12-SP2/x86_64/update
 
+## Update Repository Validation
+
+Using an incompatible self-update repository might result in a crash or
+unexpected behavior. Since SLE15-SP2/Leap15.2 the installer compares the
+versions of some specific packages in the self-update repository and in the inst-sys.
+If any unexpected version is found then an error is displayed and the self-update
+step is completely skipped.
+
+Is is possible to skip this validation is special cases and force applying the
+updates despite the version mismatch by setting environment variable
+`Y2_FORCE_SELF_UPDATE=1`.
+
+See the [Installation::SelfupdateVerifier](../src/lib/installation/selfupdate_verifier.rb)
+class for more details.
+
 ## Security
 
 ### Package Integrity
--- a/doc/SELF_UPDATE.md
+++ b/doc/SELF_UPDATE.md
@@ -226,7 +226,8 @@ step is completely skipped.
 
 Is is possible to skip this validation is special cases and force applying the
 updates despite the version mismatch by setting environment variable
-`Y2_FORCE_SELF_UPDATE=1`.
+`Y2_FORCE_SELF_UPDATE=1`. *This should be done only in special cases (testing)
+when you know what you are doing!*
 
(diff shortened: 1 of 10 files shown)


doc/SELF_UPDATE.md
```diff
@@ -216,6 +216,21 @@ under https://updates.suse.com/
...
+
+Using an incompatible self-update repository might result in a crash or
+unexpected behavior. Since SLE15-SP2/Leap15.2 the installer compares the
+versions of some specific packages in the self-update repository and in the inst-sys.
+If any unexpected version is found then an error is displayed and the self-update
+step is completely skipped.
+
+Is is possible to skip this validation is special cases and force applying the
+updates despite the version mismatch by setting environment variable
+`Y2_FORCE_SELF_UPDATE=1`.
```
dgdavid: NP: Maybe we can include here the sentence you wrote in the PR description, 

```
But this should be done only in special cases (testing) when you know what you are doing.
```

src/lib/installation/selfupdate_verifier.rb
```diff
@@ -0,0 +1,112 @@
...
+  class SelfupdateVerifier
+    include Yast::Logger
+
+    # We only check the version of these packages, the reason for a fixed
+    # list is that these packages are maintained by us and their version
+    # is under our control.
+    #
+    # We do not care about the other packages, in theory there might be
+    # valid reasons for downgrading them. With these YaST packages we can
+    # go back to the previous state but but still increase the version number.
```
dgdavid: typo: duplicated `but`

src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -113,6 +116,7 @@ def update_installer
       updated = update_repositories.map { |u| add_repository(u) }.any?
 
       if updated
+        return false unless valid_repositories?
```
dgdavid: Can we add a unit test for this?

teclator:
**Note:** Just in case we wanted to get rid of some code, it seems that the `custom_self_update_url` method is not used anymore

https://github.com/yast/yast-installation/blob/bd12c11247cd38f4b126f8ebd2f8ba93d59cfdf5/src/lib/installation/clients/inst_update_installer.rb#L203

PR #843, opened 2020-03-13
Abort confirm

jreidinger:
trello: https://trello.com/c/5h7u5BGH/1697-sles15-sp2-p3-1163347-firstboot-does-not-warn-the-user-before-abort-in-welcome

bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1163347

Issue: no confirmation when aborting

solution: use standard AbortConfirm method

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Mar 13 09:26:49 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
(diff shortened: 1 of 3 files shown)


dgdavid:
Maybe its a good opportunity to completely get rid off those `Builtins` and split the long `#main` method in small pieces. Isn't it?

jreidinger:
@dgdavid as it is not covered by tests I do not feel brave enough to do big changes and do not plan to invest too much into this client. Just small improvements.

PR #845, opened 2020-03-17
Do not differ on opensuse

jreidinger:
trello: https://trello.com/c/X6NTif5Z/1714-2-isopensuse-jira-cards
jira: https://jira.suse.com/browse/SLE-11851


jreidinger:
@lslezak well, it is just recommends and it will be installed due to skelcd requirement.

lslezak:
Ah, right, the packages are installed by a different mechanism...

PR #850, opened 2020-04-17
use LANG=UTF-8 in case of RC_LANG=POSIX

teclator:
## Problem

- During firstboot, when RC_LANG is 'POSIX' we try to export LANG as POSIX.UTF-8 which is not valid.

- https://trello.com/c/eMiBCnv0/1813-sle12-sp4-l3-1169017-yast2-firstboot-cannot-configure-network-if-disabled-confused-by-posixutf-8

## Solution

- Fallback to en_US.UTF-8 when RC_LANG is 'POSIX' or 'C'

### DIFF
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.3.0.4
+Version:        3.3.1
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.3.0.3
+Version:        3.3.0.4
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/startup/common/language.sh
+++ b/startup/common/language.sh
@@ -73,8 +73,13 @@ function set_language_cont () {
                 export RC_LANG=en_US
         fi
 
-        # get rid of encoding and/or modifier
-        export LANG=${RC_LANG%%[.@]*}.UTF-8
+        if [ "$RC_LANG" == "POSIX" ] ; then
+                log "\tRC_LANG is POSIX, using LANG en_US.UTF-8 as default..."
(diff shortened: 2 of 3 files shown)


startup/common/language.sh
```diff
@@ -73,8 +73,13 @@ function set_language_cont () {
                 export RC_LANG=en_US
```
teclator, line 73: I was not sure about modifying the `RC_LANG` when it is explicitly set to `POSIX`

startup/common/language.sh
```diff
@@ -73,8 +73,13 @@ function set_language_cont () {
...
         fi
 
-        # get rid of encoding and/or modifier
-        export LANG=${RC_LANG%%[.@]*}.UTF-8
+        if [ "$RC_LANG" == "POSIX" ] ; then
```
teclator, line 76: It is ok to fallback to **en_US.UTF-8** in this case? At least we are doing so when no RC_LANG is set

jreidinger: and what happen when RC_LANG is set to "C"?

jreidinger: BTW what about more generic check, that verify if locale is correct with `localectl list-locales` and grep its output? so do that .UTF-8 and if it is not valid, log and fall back to en_US ?

teclator: probably the same, I was wondering if we should check that condition too

package/yast2-installation.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-installation
-Version:        3.3.0.3
+Version:        3.3.0.4
```
imobachgs, line 20: The version looks good to me. However, it looks like we already jumped to `3.4.0` in `SLE-12-SP4`, so I would say it is safe to use `3.3.1` if you prefer. You can see the [jump](https://github.com/yast/yast-installation/blob/SLE-12-SP5/package/yast2-installation.changes#L2-L11) in the changes file.

teclator: We jumped to `3.4.0` in `SLE-12-SP5` mm it is there some real difference between using 3.3.0.4 or 3.3.1 at all?

jreidinger: no real difference, just consistency between packages. Only rule is never have same version with different content.

PR #863, opened 2020-06-11
Replace some labels in the Previously Used Repositories dialog

dgdavid:
## Problem

While reviewing changes done in https://github.com/yast/yast-registration/pull/466, we decided to improve the UX (although I'd say _cognitive improvement_) by labeling slightly different the _Current Status_ column.

## Solution

To use _Action_ instead _Current Status_ and re-labeling accordingly.

| Before | After | 
|-|-|
| Current Status | Action |
| Toggle Status | Toggle Action |
| Change... | Change URL... |
| Removed  | Remove |
| Enabled | Enable |
| Disabled | Disable |
| Unknown | Keep |

## Notes

No version bump required

## Tests

Only tested manually

## Screenshots

<details>
<summary>Click to show/hide screenshots when running in graphical mode</summary>

---

| Dialog | Help |
|-|-|
| Screenshot_opensusetumbleweed_2020-06-12_15:43:48 | Screenshot_opensusetumbleweed_2020-06-12_15:43:57 |

</details>

<details>
<summary>Click to show/hide screenshots when running in text mode</summary>

---

| Dialog | Help |
|-|-|
| Screenshot_opensusetumbleweed_2020-06-12_14:29:06 | Screenshot_opensusetumbleweed_2020-06-12_14:29:23 |

</details>

## Related links

* https://trello.com/c/BdAVUwtK/1614-1-change-previously-used-repositories-labels

### DIFF
--- a/src/lib/installation/clients/inst_upgrade_urls.rb
+++ b/src/lib/installation/clients/inst_upgrade_urls.rb
@@ -31,6 +31,7 @@ Yast.import "Popup"
 Yast.import "Stage"
 Yast.import "UI"
 Yast.import "Wizard"
+Yast.import "String"
 
 module Yast
   # This client allows reusing the old repositories during system upgrade.
@@ -126,10 +127,10 @@ module Yast
           Left(
             HBox(
               # TRANSLATORS: Push button
-              PushButton(Id(:edit), _("&Change URL...")),
+              PushButton(Id(:edit), edit_button_label),
               HSpacing(1),
               # TRANSLATORS: Push button
-              PushButton(Id(:toggle), _("&Toggle Action")),
+              PushButton(Id(:toggle), toggle_button_label),
               HStretch()
             )
           )
@@ -161,6 +162,22 @@ module Yast
         # TRANSLATORS: The action to perform with a repository
         _("Keep")
       end
+    end
+
+    # Returns the label for the toggle button
+    #
+    # @return [String]
+    def toggle_button_label
+      # TRANSLATORS: the "Toggle Action" button label
+      _("&Toggle Action")
+    end
+
+    # Returns the label for the change url button
+    #
+    # @return [String]
+    def edit_button_label
+      # TRANSLATORS: the "Change URL" button label
+      _("&Change URL...")
     end
 
     def refresh_dialog
@@ -300,18 +317,16 @@ module Yast
         # TRANSLATORS: help text 1/3
         _("<p>Here you see all software repositories found on the system you are upgrading. " \
           "Enable the ones you want to include in the upgrade process.</p>"),
-        # TRANSLATORS: help text 2/3, %{change_action} is replaced by the button label
+        # TRANSLATORS: help text 2/3, %{toggle_action} is replaced by the button label
         format(
-          _("<p>To enable, remove or disable an URL, click the <b>%{change_action}</b> button " \
+          _("<p>To enable, remove or disable an URL, click the <b>%{toggle_action}</b> button " \
             "or double-click the respective table item.</p>"),
-          # TRANSLATORS: the "Toggle Action" button label
-          change_action: _("Toggle Action")
+          toggle_action: Yast::String.RemoveShortcut(toggle_button_label)
         ),
         # TRANSLATORS: help text 3/3, %{change_url} is replaced by the button label
         format(
           _("<p>To change the URL, click the <b>%{change_url}</b> button.</p>"),
-          # TRANSLATORS: the "Change URL" button label
-          change_url: _("Change URL...")
+          change_url: Yast::String.RemoveShortcut(edit_button_label)
         )
       ].join
     end
--- a/src/lib/installation/clients/inst_upgrade_urls.rb
+++ b/src/lib/installation/clients/inst_upgrade_urls.rb
@@ -126,7 +126,7 @@ module Yast
           Left(
             HBox(
               # TRANSLATORS: Push button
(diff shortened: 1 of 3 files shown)


src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -140,7 +140,7 @@ def display_dialog
         ) +
           # TRANSLATORS: help text 2/3
           _(
-            "<p>To enable, remove or disable an URL, click the\n<b>Toggle Status</b> button or double-click the respective table item.</p>"
```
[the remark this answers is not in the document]
dgdavid: Something similar happens with the dialog title

https://github.com/yast/yast-installation/pull/863/files#diff-0fa16fa91089733bf40e8f4324d0b777L109

which does not need it because it fits in a single line (at least in English).

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -151,20 +151,24 @@ def display_dialog
...
+        # TRANSLATORS: The action to perform with a repository
+        _("Disable")
+      when :enable
+        # TRANSLATORS: The action to perform with a repository
+        _("Enable")
       else
-        # TRANSLATORS: The repository status is unknown
-        _("Unknown")
+        # TRANSLATORS: The repository status is unknown, none action will be taken
+        _("None")
```
dgdavid, line 171: Since now the concept is _an action to perform_, I decided to use "None" when the repo status is unknown. IMHO,  having _Action: Unknown_ is quite confusing. However, I'm not sure if this can be problematic. During the manual test, I could check that all repos were proposed to be "Removed" but using the "Toggle Action" button for one of them turns the action into "None" (i.e., the repo gets an unknown status)

Screenshot_opensusetumbleweed_2020-06-11_17:06:21

@lslezak, what do you think?

lslezak: Action "None" looks a bit strange, it's not clear what will actually happen, what will be the resul. I'd prefer "Keep" or "Use" so the result is more clear.

Also the "Change..."  is a bit misleading. What it will change? I think "Change URL..." would be much better.

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -140,7 +140,7 @@ def display_dialog
         ) +
           # TRANSLATORS: help text 2/3
           _(
-            "<p>To enable, remove or disable an URL, click the\n<b>Toggle Status</b> button or double-click the respective table item.</p>"
+            "<p>To enable, remove or disable an URL, click the\n<b>Toggle Action</b> button or double-click the respective table item.</p>"
```
lslezak, line 143: Maybe we could improve it a bit more when we are breaking the translations anyway...

What about using a `%{change_action}` and `%{change_url}` placeholders so we do not break the translations next time?

src/lib/installation/clients/inst_upgrade_urls.rb
dgdavid: Cool! I like "Keep". And I completely agree using "Change URL..." instead of just "Change".

Thanks a lot.

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -300,5 +291,29 @@ def save_pkg_mgr
...
+      [
+        # TRANSLATORS: help text 1/3
+        _("<p>Here you see all software repositories found on the system you are upgrading. " \
+          "Enable the ones you want to include in the upgrade process.</p>"),
+        # TRANSLATORS: help text 2/3, %{change_action} is replaced by the button label
+        format(
+          _("<p>To enable, remove or disable an URL, click the <b>%{change_action}</b> button " \
+            "or double-click the respective table item.</p>"),
+          # TRANSLATORS: the "Toggle Action" button label
+          change_action: _("Toggle Action")
```
dgdavid, line 308: The bad thing is that now we have `_("&Toggle Action")` and `_("Toggle Action")`.

Any ideas to avoid such _duplication_?

src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -296,5 +291,29 @@ def save_pkg_mgr
...
+        ),
+        # TRANSLATORS: help text 3/3, %{change_url} is replaced by the button label
+        format(
+          _("<p>To change the URL, click the <b>%{change_url}</b> button.</p>"),
+          # TRANSLATORS: the "Change URL" button label
+          change_url: _("Change URL...")
```
dgdavid, line 314: Similar to https://github.com/yast/yast-installation/pull/863#pullrequestreview-429736730

src/lib/installation/clients/inst_upgrade_urls.rb
lslezak: Use [String.RemoveShortcut](https://github.com/yast/yast-yast2/blob/66bc64d08c6ff017b88e3cff63fd02d78bbcf7f1/library/types/src/modules/String.rb#L745-L759) to remove `&` from the button label.

dgdavid: Perfect! [It](https://github.com/yast/yast-installation/pull/863/commits/9f93a4f6d5f8f980d896f8d72b285484ab418211) works (see the updated screenshots)

dgdavid:
I solve the conflicts, fixed some RuboCop offenses (not related with this PR) and now I am going to merge it.

PR #864, opened 2020-06-11
Collect memory usage data (bsc#1172139)

mvidner:
- https://trello.com/c/2RY7fqpT/1891-8-research-study-and-maybe-reduce-memory-consumption
- https://bugzilla.suse.com/show_bug.cgi?id=1172139

### Docs

Read the included documentation first: [memsample.md](../blob/3a4bc7e4fb9de1eb61e1cc43d07870d35746849d/doc/memsample.md) (its sections: Problem, Approach, Implementation, Usage, Overhead)

### TL;DR

memusage-y2start-rss

### Usage until merged

Until this PR makes its way into the installation medium you need to apply it manually:

1. boot the installation with network and shell access, for example by the   boot parameters `ifcfg=*=dhcp ssh=1 sshpassword=linux`
2. apply this with `yupdate patch yast-installation memusage` (the branch   name)
3. run the installation with `yast.ssh`

Afterwards proceed like described in the Docs:

### Identified Memory Savings

1. ag_uid: -11 MiB, https://github.com/yast/yast-users/pull/231

2. ag_udev_persistent: -31 MiB, https://github.com/yast/yast-network/pull/1080 and https://github.com/yast/yast-core/pull/147 (both merged)

3. dash, https://github.com/openSUSE/installation-images/pull/386 (for memsample itself, merged), https://github.com/openSUSE/installation-images/pull/388 (for inst_setup, -1.5 MiB)

Dash is a POSIX shell with a small disk+memory footprint.

With bash, memusage and its pipe-created subshell take up 3452+2032=5484 KiB RSS.
With dash, it's 1768+160=1928 KiB :)
The entire unpacked dash.rpm has 159KiB.


bin/memsample-csv-plot
```diff
@@ -0,0 +1,30 @@
...
+echo >&2 "Plotting from $CSV to $PNG"
+cat <<EOS | gnuplot > "$PNG"
+set title "YaST instaler memory usage"
+set datafile separator comma
+set timefmt "%Y-%m-%dT%H:%M:%S+00:00"
+set xdata time
+set xtics rotate
+set terminal png size 800, 600
+set xlabel "wall clock (UTC)"
+set ylabel "used [KiB]"
```
lslezak, line 15: Could we rather use MiB units? Numbers like `500000` without any delimiters are difficult to read, is it 50MiB? 500MiB?

startup/First-Stage/F08-logging
```diff
@@ -51,3 +51,10 @@ export Y2DEBUGONCRASH=1
 log "\tSet YaST2 LOG parameters:"
 log "\tMaximum log size:  $Y2MAXLOGSIZE kB"
 log "\tMaximum log count: $Y2MAXLOGNUM"
+
+#=============================================
+# 8.3) Start sampling memory usage data
+#---------------------------------------------
+log "\tStart sampling memory usage data..."
+memsample --archive=/var/log/YaST2/memsample.zcat &
```
lslezak, line 59: The question is how big is the overhead for measuring the memory usage. How much memory does the process take and what's roughly the size of the created log (which is stored in RAM-disk)?

I mean, is it a good idea to run it *always*? Maybe we could allow disabling it (via e.g. `$YAST_MEMSAMPLE=0` or something like that).

doc/memsample.md
```diff
@@ -0,0 +1,59 @@
...
+Previously we have inserted logging of memory measurements at critical points
+of the main installer process (y2start or y2base).
+
+**memsample** aims to get a full* picture of the install-time system
+(aka *inst-sys*) by
+1. taking a sample every few seconds
+2. measuring the memory used by all the processes, and the RAM-disk (tmpfs)
+3. having low overhead as to allow being always enabled in production
+
+*: A significant problem with time-sampling memory or disk usage is that we may
```
lslezak, line 29: BTW the `/proc/$$/status` file contains a `VmPeak` value, but I'm not sure how much useful is that...

doc/memsample.md
```diff
@@ -0,0 +1,59 @@
...
+![YaST instaler memory usage, SLED15-SP2 GMC](memsample.png)
+
+To reproduce this, see Usage below.
+
+### Approach
+
+Previously we have inserted logging of memory measurements at critical points
+of the main installer process (y2start or y2base).
+
+**memsample** aims to get a full* picture of the install-time system
```
lslezak, line 23: Could we use a different character for the `*` mark, like `†`? The asterisk symbol usually confuses the Markdown syntax highlighting...

bin/memsample
```diff
@@ -0,0 +1,53 @@
...
+            shift
+            ;;
+        --sleep=*)
+            SLEEP=${1#--sleep=}
+            shift
+    esac
+done
+# default values
+: "${ARCHIVE=./memsample.zcat}"
+: "${SLEEP=5}"
```
lslezak, line 23: The names are quite generic, I'd rather add some prefix like `MEMSAMPLE_` to avoid possible collisions with other tools when you set it directly on the boot command line.

bin/memsample
```diff
@@ -0,0 +1,53 @@
...
+            ARCHIVE=${1#--archive=}
+            shift
+            ;;
+        --sleep=*)
+            SLEEP=${1#--sleep=}
+            shift
+    esac
+done
+# default values
+: "${ARCHIVE=./memsample.zcat}"
```
lslezak, line 22: Why not the usual `.gz` suffix?

bin/memsample
```diff
@@ -0,0 +1,53 @@
...
+: "${SLEEP=5}"
+
+echo >&2 "Sampling every $SLEEP s to $ARCHIVE"
+
+{
+    echo "### comment-0000-$(date -Iseconds)"
```
lslezak, line 28: Maybe we could use the `--utc` option here and avoid normalizing later...??

bin/memsample-csv-plot
mvidner: Right, will do.

bin/memsample
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
+# Collect data about memory usage of all processes
+DOCS="https://github.com/yast/yast-installation/blob/master/doc/memsample.md"
+
+# save memory by switching to a smaller shell
+if [ -n "$BASH_VERSION" ] && [ -x /bin/dash ]; then
```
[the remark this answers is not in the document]
mvidner: With bash, memusage and its pipe-created subshell take up 3452+2032 KiB RSS.
An interactive dash takes up about 1000 KiB.
(WIP: dud=dash.rpm to measure dash in inst-sys)

startup/First-Stage/F08-logging
mvidner: Process: with bash it's 5.5 MiB of memory (dash: WIP)
Logs: A quick next-next installation of SLES took about 13 minutes, that was 150 samples totaling 392 KiB of memsample.zcat

Being turned on by default is IMHO a major feature, allowing to diagnose unexpected memory problems. I will proceed with the ag_udev_persistent PR which will save 25 MiB and pay for this :)

How about: 

> The environment variable `MEMSAMPLE` (set on the boot command line) means the interval in seconds between samples. Zero means no sampling at all. The default is 5.

doc/memsample.md
mvidner: I see, thanks. I expect to learn more about memory values as we look deeper.

bin/memsample
mvidner: No, these are local. Implementing a boot parameter is useful but in this case we can use process arguments and they are better.

bin/memsample-archive-to-csv
```diff
@@ -0,0 +1,153 @@
...
+require "date"
+require "shellwords"
+
+class ZcatArchive
+  def initialize(filename)
+    @filename = filename
+  end
+
+  def each(&block)
+    IO.popen("zcat #{@filename.shellescape}") do |gz|
```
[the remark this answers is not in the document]
mvidner: That's what I thought too but it turns out not to work! GzipReader will ignore the appended gz  data. See https://github.com/ruby/zlib/pull/13
Will add a comment here.

bin/memsample
mvidner: It is really a leftover of my earlier idea to store the headers in the gzip filename header, without the "### " marker. I would like to keep the suffix in case I get to implement that.

bin/memsample
mvidner: Makes sense, but maybe also not: y2log uses the local time without a TZ marker (which is stupid) and logging local time here may make correlating the logs easier.

bin/memsample-archive-to-csv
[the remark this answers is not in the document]
lslezak: Oh, interesting...

startup/First-Stage/F08-logging
lslezak: I like the "zero = no sampling" idea! :+1:

bin/memsample
lslezak: I see, using the same time as in `y2log` makes sense...

bin/memsample
lslezak: Linuxrc by default exports unknown variables into environment
- I have started installation with `startshell=1 SLEEP=42`
- Created this simple `foo` script:
```shell
: "${SLEEP=5}"
echo "$SLEEP"
```
- Running `bash foo` prints `42`...

mvidner:
Followup ideas: https://gist.github.com/mvidner/9d959eb91190bce35e0d190324a80fb8 "installer memory consumption: where to look next"

PR #871, opened 2020-07-09
remove access to autoyast profile unless it is necessary

jreidinger:
trello: https://trello.com/c/kCfh7Ylo/1941-5-drop-references-to-profilecurrent

### DIFF
--- a/src/lib/installation/clients/inst_update_installer.rb
+++ b/src/lib/installation/clients/inst_update_installer.rb
@@ -172,7 +172,7 @@ module Yast
     #
     # @return [Boolean] true if self update has been disabled by AutoYaST profile
     def disabled_in_profile?
-      !(AutoinstGeneral.self_update?.nil? || AutoinstGeneral.self_update?)
+      AutoinstGeneral.self_update == false
     end
 
     # Return the self-update URLs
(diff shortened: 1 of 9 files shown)


src/lib/installation/clients/inst_update_installer.rb
```diff
@@ -172,8 +172,7 @@ def disabled_in_linuxrc?
     #
     # @return [Boolean] true if self update has been disabled by AutoYaST profile
     def disabled_in_profile?
-      profile = Yast::Profile.current
-      !profile.fetch("general", {}).fetch("self_update", true)
+      !(AutoinstGeneral.self_update?.nil? || AutoinstGeneral.self_update?)
```
imobachgs, line 175: Would not `AutoinstGeneral.self_update? == false` be equivalent? However, allowing a method called `self_update?` to return true, false or nil looks strange to me, although I understand the rationale behind that.

jreidinger: @imobachgs yes, suggestion with false looks good. Originally it was just boolean, but there are two places that checks it and both have different default :(

imobachgs: Then I would consider renaming it to `self_update` just to avoid confusion. But it is just a suggestion, feel free to keep it as it is (as long as it is documented).

PR #872, opened 2020-07-16
Moving <files> section handling from second installation stage to fir…

schubi2:
…st installation stage.

### DIFF
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-installation
-Version:        4.3.7
+Version:        4.3.8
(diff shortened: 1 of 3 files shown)


package/yast2-installation.spec
```diff
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-installation
-Version:        4.3.8
+Version:        4.3.9
```
jreidinger, line 19: change looks ok, just please update dependency, so it ensure that client exists. Somewhere at https://github.com/yast/yast-installation/pull/872/files#diff-e8dd2765acbfec5b084347f6e1248121R113 just increase version.

PR #873, opened 2020-07-17
handle device autoconfig setting in summary screen (bsc#1168036)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1168036
- https://trello.com/c/if9ZgetC

yast should offer the option to turn off device autoconfig on s390. Similar to the cio settings.

## Related

This needs https://github.com/yast/yast-bootloader/pull/612 to make it possible to adjust the `rd.zdev` kernel option.

## Screenshots
x1

### DIFF
--- a/src/lib/installation/cio_ignore.rb
+++ b/src/lib/installation/cio_ignore.rb
@@ -2,28 +2,24 @@ require "yast"
 
 module Installation
   class CIOIgnore
+    # FIXME
+    # The class name is a bit outdated now that it handles both cio_ignore
+    # and rd.zdev kernel parameters.
     include Singleton
+    include Yast::Logger
     Yast.import "Mode"
     Yast.import "AutoinstConfig"
 
-    attr_accessor :enabled
+    attr_accessor :cio_enabled
+    attr_accessor :autoconf_enabled
 
     def initialize
       reset
     end
 
     def reset
-      @enabled = if Yast::Mode.autoinst
-        Yast::AutoinstConfig.cio_ignore
-      else
-        @enabled = if kvm? || zvm?
-          # cio_ignore does not make sense for KVM or z/VM (fate#317861)
-          false
-        else
-          # default value requested in FATE#315586
-          true
-        end
-      end
+      @autoconf_enabled = autoconf_setting
+      @cio_enabled = cio_setting
     end
 
   private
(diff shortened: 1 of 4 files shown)


wfeldt:
Note that Travis fails in an unrfelated area - possibly due to the latest pkg-bindings changes.

src/lib/installation/cio_ignore.rb
```diff
@@ -37,6 +33,33 @@ def zvm?
...
+    def autoconf_setting
+      Yast.import "Bootloader"
+
+      rd_zdev = Yast::Bootloader.kernel_param(:common, "rd.zdev")
+      log.info "current rd.zdev setting: rd.zdev=#{rd_zdev.inspect}"
+
+      rd_zdev != "no-auto"
+    end
+
+    # Get current device blacklist setting (cio_ignore kernel option)
```
aschnell, line 49: Is that political correctness?

aschnell:
The build failure looks like something Imo just fixed a few days ago.

src/lib/installation/cio_ignore.rb
wfeldt: As long as we develop in a `master` branch I see no need to force matters.

lslezak: :smiley:

src/lib/installation/cio_ignore.rb
```diff
@@ -63,44 +88,57 @@ def run(*args)
...
+    #
+    # @return [String] HTML fragment
+    #
+    # @example
+    #   msg = toggle_text("Foobar state", true, "foobar_enable", "foobar_disable")
+    #
+    def toggle_text(what, state, link_on, link_off)
+      format "%s: %s (<a href=\"%s\">%s</a>).",
+        what,
+        (state ? _("enabled") : _("disabled")),
```
lslezak, line 120: NP: in the long term we should somehow share these texts, they are already used at several places...

src/lib/installation/cio_ignore.rb
```diff
@@ -63,44 +88,57 @@ def run(*args)
...
+    # @param link_on [String] HTML ref for turning state to 'on'
+    # @param link_off [String] HTML ref for turning state to 'off'
+    #
+    # @return [String] HTML fragment
+    #
+    # @example
+    #   msg = toggle_text("Foobar state", true, "foobar_enable", "foobar_disable")
+    #
+    def toggle_text(what, state, link_on, link_off)
+      format "%s: %s (<a href=\"%s\">%s</a>).",
```
lslezak, line 118: I think we should make this string translatable to allow some modifications in some languages. E.g. in RTL languages they might need to change the order of the status text and the toggling link.

wfeldt: I don't think it matters. When you compare two examples: bootloader vs. firewall where bootloader constructs the string (similar as I do here) and firewall translates the entire string, the output looks the same. Compare the screenshots:

- bootloader (see the trusted boot entry)
b
b1

- firewall (see the firewall enable entry)
f
f1

The respective code blobs are here:

- https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/grub2base.rb#L395-L402
- https://github.com/yast/yast-firewall/blob/master/src/lib/y2firewall/clients/proposal.rb#L194-L206

What this does highlight, however, is that giving people the chance to translate html fragments, easily leads to desaster.

lslezak: Interesting... Thanks for analyzing it and for the screenshots! :+1:

PR #874, opened 2020-07-17
Fix UpgradeRepoManager tests

imobachgs:
Fix UpgradeRepoManager tests. It looks like a side effect of https://github.com/yast/yast-yast2/pull/1071.

Open question: do we need to adapt the `UpgradeRepoManager` API?


imobachgs:
> Since it is only in the unit tests, it LGTM. But remember that now we can/should use `ZyppUrl` instead of `URI` See [yast/yast-yast2#1071](https://github.com/yast/yast-yast2/pull/1071)

Yes, I just wanted to fix a problem when building the package. Anyway, according to the API documentation, both classes (URI and ZyppUrl) are supported.

PR #876, opened 2020-07-27
AY: cleanup image creation/deployment

schubi2:
Creating Images will not be available anymore in AY. It is an undocumented feature which has so many bugs.
We have never gotten any bug report about it. ---> we drop it.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Jul 27 17:04:26 CEST 2020 - schubi@suse.de
+
+-  AY: Removed "image" section from "software" section
+   (bsc#1140711).
+- 4.3.12
+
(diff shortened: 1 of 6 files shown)


package/yast2-installation.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Jul 27 17:04:26 CEST 2020 - schubi@suse.de
+
+-  AY: Removed "image" section from "software" section
```
imobachgs, line 4: np: wrong indentation (two whitespaces).

src/lib/installation/clients/deploy_image_auto.rb
```diff
@@ -1,426 +1,172 @@
-# encoding: utf-8
+require "yast"
```
imobachgs, line 1: Please, add the license (the SUSE one).

PR #878, opened 2020-07-28
[WIP] Warn when an old package is selected to install

lslezak:
## Notes

:warning: *Do not merge yet! We need to add the real data about the old packages!* :warning: 

This is just a draft implementation for getting some feedback early.

## Problem

We need to display a warning when user is going to install some known old packages which contain some serious bugs or do not work properly in some situations.

## The Solution

We need to somehow get the data about the obsolete packages, hardcoding the package names and versions directly in the YaST code does not look like a good idea.

My proposal is to read the data about the old packages from YAML files stored in `data/old_packages/*.y{,a}ml`. Each file would contain a list of known old packages (with name, version and architecture) and a message which is displayed to the user when a package older or equal to the specified version is selected to install.

See the example `src/data/old_packages/sle15_sp1.yml` file below in the diff.

### Advantages

- The YAML file can be provided by any package and can be added to the inst-sys by any mechanism (DUD, self-update, 3rd party add-on).
- The packages with the same message are grouped together, this means it's possible to easily add a new package to the existing list.
- The installer honors the `Y2DIR` setting and prefers the `/y2update/` files so the existing files can be overridden by a 3rd party add-on.

### AutoYaST Support

The warning message is also displayed in the AutoYaST mode. By default warnings are automatically closed after 10 seconds time out so this change does not block the AutoYaST installation.

### Missing Feature

We do not support translatable strings in YAML files. Moreover patching the translations is not easy with the current translations architecture. The problem is that the translations are stored per language (`yast2-trans-<LANG>` packages), we would need to release a self-update for many packages.

And getting the actual translations would be difficult, the problematic packages are found after the release and then it's hard to get new translations. And we would need them ASAP to react quickly when a new serious problem is found.

Therefore the translations are not supported. But I think having just English messages is acceptable in this case, given how difficult it would be to get the translations and release them...

## Screenshots

Tested in SP1 with a testing YAML file.

old_packages_popup_sp1

## TODO

- [ ] Check it also in the upgrade mode (and AutoUpgrade as well...)
- [ ] Adapt the `src/data/old_packages/sle15_sp1.yml` file to contain the *real* data

### DIFF
--- a/src/lib/installation/old_package.rb
+++ b/src/lib/installation/old_package.rb
@@ -23,6 +23,11 @@ module Installation
 
     attr_reader :name, :version, :arch, :message
 
+    # @param name [String] name of the package
+    # @param version [String] version of the package
+    # @param arch [String] architecture, e.g. "x86_64"
+    # @param message [String] the error message displayed to the user
+    #  when an old package is selected
     def initialize(name:, version:, arch:, message:)
       @name = name
       @version = version
@@ -30,6 +35,8 @@ module Installation
       @message = message
     end
 
+    # Finds the currently selected old package, if none or newer is selected then
+    # it returns `nil`.
     # @return [Hash,nil] The selected old package or nil.
     def selected_old
       packages = Yast::Pkg.ResolvableProperties(name, :package, "")
@@ -43,6 +50,8 @@ module Installation
 
     # Reads the old package configuration files and creates the respective
     # OldPackage objects.
+    # @return [Array] Configured old packages, empty list if no configuration
+    #  is specified
     def self.read(paths = nil)
       # unfortunately we cannot use Yast::Directory.find_data_file
       # here because it needs an exact file name, it does not accept a glob,
--- a/src/lib/installation/old_package.rb
+++ b/src/lib/installation/old_package.rb
@@ -41,17 +41,21 @@ module Installation
     def selected_old
       packages = Yast::Pkg.ResolvableProperties(name, :package, "")
 
-      # 1 = the selected is newer, the opposite is older or the same
+      # -1 = the second version (the selected package) is newer
       packages.find do |p|
         p["status"] == :selected && p["arch"] == arch &&
-          Yast::Pkg.CompareVersions(version, p["version"]) == 1
+          Yast::Pkg.CompareVersions(version, p["version"]) != -1
       end
     end
 
     # Reads the old package configuration files and creates the respective
-    # OldPackage objects.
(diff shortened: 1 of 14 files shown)


src/lib/installation/old_package.rb
```diff
@@ -0,0 +1,96 @@
...
+      @message = message
+    end
+
+    # Finds the currently selected old package, if none or newer is selected then
+    # it returns `nil`.
+    # @return [Hash,nil] The selected old package or nil.
+    def selected_old
+      packages = Yast::Pkg.ResolvableProperties(name, :package, "")
+
+      # 1 = the selected is newer, the opposite is older or the same
```
jreidinger, line 44: this is a bit confusing as later you compare with 1

src/lib/installation/old_package.rb
```diff
@@ -0,0 +1,96 @@
...
+      packages.find do |p|
+        p["status"] == :selected && p["arch"] == arch &&
+          Yast::Pkg.CompareVersions(version, p["version"]) == 1
+      end
+    end
+
+    # Reads the old package configuration files and creates the respective
+    # OldPackage objects.
```
jreidinger, line 52: document param please

src/lib/installation/old_package.rb
lslezak: Ah, I orignally used `!= 1` in the code. Later I changed the order of the parameters and used `== 1` to make it better readable. And obviously I forgot to adapt the comment...

src/lib/installation/old_package_report.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+
+        str << "<p>"
+        str << Yast::HTML.List(package_names)
+        str << msg
+        str << "</p><br>"
+      end
+
+      message = format(_("The installer detected old package versions selected " \
+        "for installation: \n\n%{list}"), list: pkg_summary)
```
jreidinger, line 58: does `\n\n` has effect when richtext is used?

lslezak: Ah, good point, originally I used a plain text message...

lslezak:
This is waiting for the final approval by @jsrain and the maintenance team. Also the exact message and and the list of the affected packages should be provided.

mvidner:
@lslezak I've just run across this while checking another old PR of mine... I wanted to check its status myself but there is no context, like a bug, PBI or similar.

Is this still relevant?

lslezak:
Not relevant anymore.

PR #880, opened 2020-08-11
AutoYaST: Added supplements: autoyast(deploy_image,ssh_import) into t…

schubi2:
…he spec file in order to install this packages if the section has been defined in the AY configuration file (bsc#1146494).


schubi2:
We have come to the conclusion that this entry does not depend on the first/second stage.

PR #892, opened 2020-10-27
yupdate - make also /usr/share/icons writable

lslezak:
- It is needed for updating yast2-theme
- No package version bump, let's release it with other "real" fix


dgdavid (approved):
LGTM

joseivanlopez (approved):

PR #893, opened 2020-10-29
AY: write linuxrc proxy and hostname to inst-sys

teclator:
## Problem

During an autoinstallation, if the networking section only contains the [keep_install_network](https://documentation.suse.com/sles/15-SP2/single-html/SLES-autoyast/#CreateProfile-Network) option it should copy the network configuration from linuxrc to the target system. This is not the case of the hostname option which is lost.

- https://bugzilla.suse.com/show_bug.cgi?id=1177768

## Solution

Write the linuxrc hostname and proxy configuration also during an autoinstallation.

It needs https://github.com/yast/skelcd-control-leanos/pull/67 in order to be run.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Oct 27 22:30:43 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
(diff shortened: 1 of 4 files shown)


mchf (approved):
lgtm

jreidinger (approved):

PR #895, opened 2020-11-24
fix full media product selection (bsc#1179094, bsc#1176424)

wfeldt:
## Problem

- https://trello.com/c/3NmLtDUr
- https://bugzilla.suse.com/show_bug.cgi?id=1179094
- https://bugzilla.suse.com/show_bug.cgi?id=1176424

When the full media contain only a single product, yast crashes trying to display the license.

## Analysis

The problem starts here:

- https://github.com/yast/yast-installation/blob/master/src/lib/installation/clients/inst_complex_welcome.rb#L172-L198

`available_base_products` returns the list of products but for the offline media ('full media') the returned type is a list of `Y2Packager::ProductLocation` classes which are not compatible with `Y2Packager::Product` which might also be returned and are in fact implicitly assumed in (part of) the code.

The issue arises when there's only a single product as then yast tries to skip the product selection and to show a license confirmation instead on the first screen.

## Solution

For offline media, enforce the product selection, even if there's only a single product. The license will be available later when the product dir is accessed.

## Tests

Tested with SLE15-SP3, SLE15-SP2, Leap 15.2, and Tumbleweed.

## Comment

Not sure why `Y2Packager::ProductLocation` doesn't also have the `ProductLicenseMixin` (and would provide the license).

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Nov 26 14:06:30 UTC 2020 - Steffen Winterfeldt <snwint@suse.com>
+
+- fix full media product selection (bsc#1179094, bsc#1176424)
+- 4.3.21
+
+-------------------------------------------------------------------
 Tue Oct 27 22:30:43 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
 
 - Write hostname and proxy configuration to the inst-sys when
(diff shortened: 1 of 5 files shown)


lslezak:
Please also check how it behaves in openSUSE Leap or Tumbleweed, there is just a single product to install...

lslezak (approved):
Thanks for debugging and fixing!

PR #901, opened 2020-12-22
bsc#1180142: do not use 'installation-helper' to create snapshots

imobachgs:
Adapt to `FsSnapshot` API changes introduced in yast/yast-yast2#1125.

post-update-snapshot-error

### DIFF
--- a/src/lib/installation/snapshots_finish.rb
+++ b/src/lib/installation/snapshots_finish.rb
@@ -53,7 +53,7 @@ module Installation
       Yast2::FsSnapshot.create_post("after update", pre_number, cleanup: :number, important: true)
       Yast2::FsSnapshotStore.clean("update")
       true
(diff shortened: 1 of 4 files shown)


src/lib/installation/snapshots_finish.rb
```diff
@@ -52,12 +53,18 @@ def create_post_snapshot
       Yast2::FsSnapshot.create_post("after update", pre_number, cleanup: :number, important: true)
       Yast2::FsSnapshotStore.clean("update")
       true
+    rescue Yast2::SnapshotCreationFailed, Yast2::FsSnapshotsStore::IOError
+      Yast::Report.Error(_("Could not create a post-update snapshot."))
```
jreidinger, line 57: I would probably also add a logging line that this happen, so it can be easily found.

jreidinger (approved):

jreidinger (approved):

PR #904, opened 2021-01-19
Replace old splash commands with plymouth ones

hellcp:
This is the second part to openSUSE/linuxrc#245


wfeldt:
That's all well and nice but plymouth is turned off before yast is started:

- https://github.com/openSUSE/installation-images/blob/master/data/root/etc/inst_setup#L244

So, maybe the correct approach would be to remove the old splash code without replacing it?

Alternatively, recheck the entire plymouth workflow and see if we'd rather keep it running.

dgdavid:
@hellcp, friendly ping :) 

What do you think after @wfeldt review?

PR #905, opened 2021-01-27
The Installer Console (SLE-16263, SLE-16358)

lslezak:
## The Goal

The goal is to enable configuring some devices (network, proxy) or setting some installer options in special cases. Usually it should not be needed but in some rare cases it would nice to offer some hidden configuration options.

## Links

- [Internal Trello card](https://trello.com/c/TOdZIh98/2209-5-continue-with-an-easy-to-way-to-configure-and-in-installer)
- Jira: [SLE-16263](https://jira.suse.com/browse/SLE-16263), [SLE-16358](https://jira.suse.com/browse/SLE-16358)

## Notes

Because it's too late to change the dialogs in the SLE15-SP3 (and Leap 15.3) the push button is displayed only in a Tumbleweed installation. In SLE15-SP3 we will use a keyboard shortcut (needs https://github.com/yast/yast-ruby-bindings/pull/270), that will also work in Tumbleweed.

#### Graphical Installation

installation_console_gui

#### Text Mode Installation

installation_console_tui

#### Plain VGA Text Console

In a plain VGA console (without KMS) the unicode characters cannot be displayed. Fortunately it displays a full block in that case which is quite similar to the [hamburger icon](https://en.wikipedia.org/wiki/Hamburger_button) used. So lets use this fallback.

console_plain_text_console

#### SSH Installation

When installing via SSH the icon is displayed properly (if the terminal properly supports it).

##### xterm
installation_console_ssh_xterm

##### KDE konsole
installation_console_ssh_konsole

##### GNOME terminal

installation_console_ssh_gnome_term

## Usage

The extra configuration menu allow starting several YaST configuration modules.

## The Expert Commandline Console

The console supports only few commands, additional commands can be defined in each YaST module

- `quit` - closes the console and resumes the installation
- `commands` - prints the available commands  (`help` is a built-in IRB command)

As it runs in an IRB session you can enter any valid Ruby command. The session is running in the YaST installer context so you can inspect the installer settings (e.g. `Yast::WFM.GetLanguage`), change variables, run methods... just like in a debugger.

### Hints

- The Tab completion is enabled, for writing the `repositories` command you can just press `repos<Tab>`, the commands are  autocompleted.
- In the prompt you use the usual "readline" features like history (Up/Down arrows) or history completion (PgUp).

### In Installation

- Boot the installer (latest build of SLE15-SP3, Leap 15.3 or Tumbleweed) with `startshell=1 netsetup=dhcp` (or configure static IP if you need that....)
- When the shell is started run the `yupdate patch yast-installation config_console` command - this will patch the installer with the code from GitHub
- Then run the `yast` command to start the installer as usually
- In Tumbleweed it adds a new button in the language selection screen
- You can also start the console at any point during installation in (graphical installation only, in text mode this does not work :worried: ):
  - Press the `Ctrl+Alt+Shift+D` combination start a debugger session
  - Enter commands `require "installation/console/menu"` to load the code and `::Installation::Console::Menu.run` to actually start the console

### In Running System

You can start the configuration dialog or the console manually. 

```
# command line console
Y2DIR=./src ruby -I src/lib -r installation/console.rb -e ::Installation::Console.run
# configuration dialog
Y2DIR=./src ruby -I src/lib -r installation/console/menu.rb -e 'Yast.ui_component="qt";Yast.import("Wizard");Yast::Wizard.CreateDialog;::Installation::Console::Menu.run'
```

As the installer is not running and there is no UI the most console commands do not make sense. But this can be used for testing the IRB setup, opening xterminal, etc...

## Commandline Console Implementation Details

I took advantage of already existing IRB support in Ruby which allows to easily run a command prompt and execute Ruby commands, inspect variables, etc.

The IRB shell runs in a context of an `Installation::Console::Commands` object, the commands (like `repositories`, `configure_network` are just methods defined in that object.

The commands are loaded from plugins, each YaST module can easily define its own console commands.

In GUI YaST opens a new xterm window and redirects the YaST stdout/stdin from the first linux console to that new xterm window. (Otherwise you would need to switch via `Ctrl+Alt+F1` to the IRB session and then via `Ctrl+F7` back to the YaST installer.) Additionally a small helper is needed to change the focus between the console (xterm) and the YaST window.

In text mode it uses `UI.CloseUI` and `UI.OpenUI` to shut down and restore the ncurses UI. There are some small tweaks to make sure the terminal is properly re-initialized.

## TODO

- [x] Text mode (ncurses) support (needs this additional small fix:  https://github.com/libyui/libyui-ncurses/pull/111)
- [x] Implement a shortcut for directly starting the console (`Ctrl+Alt+Shift+C`, using `C` as console...) - https://github.com/libyui/libyui/pull/179, https://github.com/libyui/libyui-ncurses/pull/113,  https://github.com/libyui/libyui-qt/pull/145, https://github.com/yast/yast-ycp-ui-bindings/pull/66, https://github.com/yast/yast-ruby-bindings/pull/268
- [x] After finishing an YaST module in GUI focus back automatically to the console window (see https://github.com/yast/yast-x11/pull/23)
- [x] Make the commands easily extensible, allow each YaST module to add own commands (implement a plugin mechanism)

To be done later:

- [ ] Documentation - for end-user (how to use it) and developers (how to write plugins)
- [ ] Unit tests

### DIFF
--- a/src/lib/installation/console.rb
+++ b/src/lib/installation/console.rb
@@ -1,0 +1,87 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2021 SUSE LLC, All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "installation/console/commands"
+require "installation/console/gui"
+require "installation/console/tui"
+
+Yast.import "UI"
+
+# Override the IRB to log all executed commands into the y2log so we know
+# what exactly happened there (in case user did something wrong or strange...)
+module IrbLogger
+  # wrap the original "evaluate" method, do some logging around
+  def evaluate(*args)
+    statements = args[1]
+    # do not log the internal IRB command for setting the last value variable
+    if statements.is_a?(::String) && statements.start_with?("_ = ")
+      super
+    else
+      Yast::Y2Logger.instance.info "Executing console command: #{statements.inspect}"
+      ret = super
+      Yast::Y2Logger.instance.info "Console command result: #{ret.inspect}"
+      ret
+    end
+  end
+end
+
+# inject the code
+require "irb/workspace"
+module IRB # :nodoc:
+  class WorkSpace
+    prepend IrbLogger
+  end
+end
+
+module Installation
+  module Console
+    class << self
+      # open a console and run an interactive IRB session in it
+      # testing in installed system:
+      # ruby -I src/lib -r installation/console.rb -e ::Installation::Console.run
+      def run
+        console = Yast::UI.TextMode ? Console::Tui.new : Console::Gui.new
+        console.run do
+          commands = Console::Commands.new(console)
+          # print the basic help text
+          commands.welcome
+
+          # start an IRB session in the context of the "commands" object
+          irb(commands)
+        end
+      end
+
+    private
+
+      # configure IRB and start an interactive session
+      # @param context [Object] context in which the IRB session runs
+      def irb(context)
+        # lazy loading
+        require "irb"
+        # enable TAB completion
+        require "irb/completion"
+
+        # see the Binding::irb method in irb.rb in the Ruby stdlib
+        IRB.setup(eval("__FILE__"), argv: [])
+        # use a simple prompt with some customizations
+        IRB.conf[:PROMPT][:YAST] = IRB.conf[:PROMPT][:SIMPLE].dup
+        IRB.conf[:PROMPT][:YAST][:RETURN] = ""
+        IRB.conf[:PROMPT][:YAST][:PROMPT_I] = "YaST >> "
+        IRB.conf[:PROMPT_MODE] = :YAST
+        workspace = IRB::WorkSpace.new(context)
+        IRB::Irb.new(workspace).run(IRB.conf)
+      end
+    end
+  end
+end
--- a/src/lib/installation/console/commands.rb
+++ b/src/lib/installation/console/commands.rb
@@ -43,8 +43,12 @@ module Installation
         puts
         # this is the most important message so make it translatable,
         # the console is for experts so it is OK have the rest untranslated
-        # TRANSLATORS: help text displayed in the installer command line console
-        puts _("Type 'quit' or press Ctrl+D to close the console and go back to the installer")
+        # TRANSLATORS: help text displayed in the installer command line console,
+        # do not change these texts they are replaced:
+        # %{cmd} is replaced by a command name
+        # %{keys} is replaced by a keyboard shortcut
+        puts _("Type '%{cmd}' or press %{keys} to close the console and go back " \
+          "to the installer") % {cmd: "quit", keys: "Ctrl+D"}
         puts
         puts "Type 'commands' to see the available special commands"
         puts
--- a/src/lib/installation/console/commands.rb
+++ b/src/lib/installation/console/commands.rb
@@ -1,0 +1,96 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2021 SUSE LLC, All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+
+require "pp"
+require "shellwords"
+
+require "yast"
+require "installation/console/plugins"
+
+module Installation
+  module Console
+    class Commands
+      include Yast::I18n
+
+      # This class implements the commands in the installer console,
+      # the actual commands are implemented as plugins loaded from
+      # lib/installation/console/plugins/*.rb files
+      #
+      # All public methods in this class are the commands available in the console,
+      # that means we cannot include Yast::Logger here because it would define
+      # "log" command. For logging we have to use the full form here, e.g.
+      # Yast::Y2Logger.instance.info
+      def initialize(console)
+        textdomain "installation"
+
+        @console = console
+        Plugins.load_plugins
+      end
+
+      # print the "welcome" message with a basic help text
+      def welcome
+        puts "---- This is the YaST Installation Console ----"
+        puts
+        # this is the most important message so make it translatable,
+        # the console is for experts so it is OK have the rest untranslated
+        # TRANSLATORS: help text displayed in the installer command line console
+        puts _("Type 'quit' or press Ctrl+D to close the console and go back to the installer")
+        puts
+        puts "Type 'commands' to see the available special commands"
+        puts
+        puts "This is a Ruby shell, you can also type any Ruby command here"
+        puts "and inspect or change the YaST installer"
+        puts
+        puts "Hints: <Tab> completion is enabled, the command history is kept,"
+        puts "you can use the usual \"readline\" features..."
+        puts
+      end
+
+      # print the available commands
+      def commands
+        puts "Available commands:"
+        puts
+        print_command("quit", "Close the console and return back to the installer")
+
+        private_methods.grep(/_description$/).sort.each do |method|
+          print_command(method.to_s.sub(/_description$/, ""), send(method))
+        end
+
+        puts
+      end
+
+      # all unknown commands are handled via this "method_missing" callback
+      def method_missing(method_name, *_args)
+        Yast::Y2Logger.instance.info "Entered unknown command: #{method_name.inspect}"
+        puts "Error: Unknown command \"#{method_name}\""
+        puts
+        commands
+      end
+
+      # helper for running an YaST module
+      def run_yast_module(*args)
+        @console.run_yast_module(*args)
+      end
+
+    private
+
+      # print help text for a command
+      def print_command(cmd, descr)
+        # indent multiline descriptions
+        description = descr.gsub("\n", "\n    ")
+        puts "  #{cmd} - #{description}"
+        puts
+      end
+    end
+  end
+end
--- a/src/lib/installation/console/commands.rb
+++ b/src/lib/installation/console/commands.rb
@@ -48,7 +48,7 @@ module Installation
         # %{cmd} is replaced by a command name
         # %{keys} is replaced by a keyboard shortcut
         puts _("Type '%{cmd}' or press %{keys} to close the console and go back " \
-          "to the installer") % {cmd: "quit", keys: "Ctrl+D"}
+          "to the installer") % { cmd: "quit", keys: "Ctrl+D" }
         puts
         puts "Type 'commands' to see the available special commands"
         puts
--- a/src/lib/installation/console/gui.rb
+++ b/src/lib/installation/console/gui.rb
@@ -1,0 +1,136 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2021 SUSE LLC, All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+
+require "yast"
+
+module Installation
+  module Console
+    # the installer console implementation for the graphical (Qt) UI
+    class Gui
+      Yast.import "Wizard"
+
+      # open a console and run a block in it
+      def run(&block)
+        start
+        redirect(&block)
+      ensure
+        stop
+      end
+
+      # helper for running an YaST module in console
+      def run_yast_module(*args)
+        # we cannot run any YaST module if there is a popup displayed, the module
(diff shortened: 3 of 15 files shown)


src/lib/installation/console_commands.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+
+    def configure(what = nil)
+      case what
+      when nil
+        commands
+      when :network
+        # we cannot run the module if there is a popup displayed, the module
+        # would be displayed *below* the popup making it inaccessible :-(
+        return unless wizard_dialog?
```
kobliha, line 58: What about creating another Wizard dialog instead? Imagine that someone opens the debugger / console just because there is this pop-up dialog?

src/lib/installation/console_commands.rb
```diff
@@ -0,0 +1,104 @@
...
+
+        puts "Starting the network configuration module..."
+        puts
+        puts "After it is finished (by pressing [Next]/[Back]/[Abort])"
+        puts "press Alt+Tab to get back to this console."
```
kobliha, line 63: What about doing that automatically? https://stackoverflow.com/questions/6908662/sending-ctrlc-using-sendkeys-in-ruby 

OK, it's hack, but this is also a "bug" so ... IMO works well as a temporary solution

src/lib/installation/console_commands.rb
```diff
@@ -0,0 +1,104 @@
...
+
+        # wait a bit so the user can read the message above
+        sleep(5)
+
+        Yast::WFM.call("inst_lan", [{ "skip_detection" => true }])
```
kobliha, line 68: This is potentially problematic in the future. You are passing an argument skip_detection which means you need to know the internal behavior/implementation of that client in advance. And that may change in time. This is potential maintenance nightmare and suggests that the client should instead know where it runs and behave accordingly. @teclator what do you think?

Obviously, it's good enough for the POC, but it may costs us some L3s and P1s long-term if not changed.

lslezak: This exactly the case command as which is executed after poressing "Configure Network" button in the registration step. https://github.com/yast/yast-registration/blob/b021cdc4dd1891c25afc42bf87f99a4b28bb614a/src/lib/registration/helpers.rb#L98 

Some network expert told me to use this. :smiley:

src/lib/installation/console_commands.rb
lslezak: Ah, that's hack is for MS Windows... For Linux we need to use the xdotool as mentioned in the "TODO" section in the PR description.

src/lib/installation/console_commands.rb
lslezak: That does not help, I tried that. The new wizard dialog is still opened *below* the current popup.

src/lib/installation/console_commands.rb
kobliha: Yes, but now you have to change two places in installer if that API changes...

src/lib/installation/console_commands.rb
kobliha: Well, my question was rather generic, but the xdotool will work, thx :)

src/lib/installation/console_commands.rb
lslezak: I see, you are right...

teclator: I do not expect it to change it in short as it is there since.. [2014](https://github.com/yast/yast-network/commit/f0d75323fd5282a22a55a60c697f1d55c70a7b1f) but maybe we could rethink that parameter or have a separate client instead of modify the behavior based in some argument.

src/lib/installation/console.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    private
+
+      # configure IRB and start an interactive session
+      # @param context [Object] context in which the IRB session runs
+      def irb(context)
+        # lazy loading
+        require "irb"
+        # enable TAB completion
+        require "irb/completion"
```
jreidinger, line 73: what about exception handling here?

src/lib/installation/console.rb
```diff
@@ -0,0 +1,87 @@
...
+
+        # see the Binding::irb method in irb.rb in the Ruby stdlib
+        IRB.setup(eval("__FILE__"), argv: [])
```
jreidinger, line 76: how stable are those methods? Did it change often in past?

src/lib/installation/console/commands.rb
```diff
@@ -0,0 +1,96 @@
...
+module Installation
+  module Console
+    class Commands
+      include Yast::I18n
+
+      # This class implements the commands in the installer console,
+      # the actual commands are implemented as plugins loaded from
+      # lib/installation/console/plugins/*.rb files
+      #
+      # All public methods in this class are the commands available in the console,
```
jreidinger, line 29: cannot you overwrite visibility here for log commands? to make it protected or private only?

src/lib/installation/console/commands.rb
```diff
@@ -0,0 +1,96 @@
...
+        Plugins.load_plugins
+      end
+
+      # print the "welcome" message with a basic help text
+      def welcome
+        puts "---- This is the YaST Installation Console ----"
+        puts
+        # this is the most important message so make it translatable,
+        # the console is for experts so it is OK have the rest untranslated
+        # TRANSLATORS: help text displayed in the installer command line console
```
jreidinger, line 46: mention that quit annd also ctrl+D should be untranslated as I think that IRB is not localized. Maybe even move it out of translatable string.

src/lib/installation/console/commands.rb
```diff
@@ -0,0 +1,96 @@
...
+        puts
+      end
+
+      # print the available commands
+      def commands
+        puts "Available commands:"
+        puts
+        print_command("quit", "Close the console and return back to the installer")
+
+        private_methods.grep(/_description$/).sort.each do |method|
```
jreidinger, line 65: I am a bit lost here. Why private methods? Also I expect here a bit description how commands should look like. Now from code it looks like it needs methods + method with description suffix.

src/lib/installation/console/gui.rb
```diff
@@ -0,0 +1,136 @@
...
+      ensure
+        stop
+      end
+
+      # helper for running an YaST module in console
+      def run_yast_module(*args)
+        # we cannot run any YaST module if there is a popup displayed, the module
+        # would be displayed *below* the popup making it inaccessible :-(
+        # make sure a wizard dialog is at the top
+        return unless wizard_dialog?
```
jreidinger, line 35: Well, what about dialogs that are not wizards, but not popup? Ones that are open with OpenDialog?

src/lib/installation/console/gui.rb
```diff
@@ -0,0 +1,136 @@
...
+
+        Yast::WFM.call(*args)
+
+        # automatically switch the window focus from YaST back to the xterm window
+        system("#{SWITCHER} #{Shellwords.escape(window)}") if window
+      end
+
+    private
+
+      # path to the window switching helper tool (from yast2-x11)
```
jreidinger, line 59: will it work on wayland?

src/lib/installation/console/gui.rb
```diff
@@ -0,0 +1,136 @@
...
+        @pid, @close_fd, @tty = @read.readline.split
+      end
+
+      # stop the console, close the xterm window
+      def stop
+        # send an empty string to the waiting "read" process
+        File.write(fd_path(@pid, @close_fd), "\n") if @pid && @close_fd
+        # close the pipes
+        @read.close if @read
+        @write.close if @write
```
jreidinger, line 103: which ruby we target? If SP3 and newer, then you can use `@read&.close` and same for write.

src/lib/installation/console/gui.rb
```diff
@@ -0,0 +1,136 @@
...
+        stderr_orig = $stderr.dup
+        stdin_orig = $stdin.dup
+
+        # redirect all IO to the xterm window (its tty device)
+        $stdout.reopen(@tty)
+        $stderr.reopen(@tty)
+        $stdin.reopen(@tty)
+
+        begin
+          block.call if block_given?
```
jreidinger, line 119: block should be always given, not?

src/lib/installation/console/menu.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      def run
+        return nil unless can_start?
+
+        loop do
+          ret = super
+          break if [:next, :back, :abort, :close].include?(ret)
+        end
+
+        nil
```
jreidinger, line 54: is explicit return needed?

src/lib/installation/console/menu_plugin.rb
```diff
@@ -0,0 +1,53 @@
...
+require "yast"
+
+module Installation
+  module Console
+    # base class for the configuration menu plugins
+    class MenuPlugin
+      extend Yast::Logger
+
+      # when a new subclass is defined this code gets executed
+      def self.inherited(subclass)
```
jreidinger, line 23: nice hook, but how we get ruby load that plugin?

src/lib/installation/console/menu_plugin.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+
+      # return a widget do display
+      # @return [CWM::Widget] the widget
+      def widget
+        nil
+      end
+
+      # define the display order (in ascending order),
+      # if the order value is same than it sorts by the class name
```
jreidinger, line 47: document return value

src/lib/installation/console/plugins.rb
```diff
@@ -0,0 +1,48 @@
...
+  module Console
+    # helper module for loading the console plugins
+    module Plugins
+      extend Yast::Logger
+
+      # load all console plugins from lib/installation/console/plugins
+      # subdirectory
+      def self.load_plugins
+        # use Yast.y2paths to honor the Y2DIR setting
+        plugin_paths = Yast.y2paths.map { |p| File.join(p, "lib/installation/console/plugins") }
```
jreidinger, line 26: well, if plugin is in ruby load path, then this won't work. So I am fine with required namespace, but we should not ignore ruby load path for searching plugins

src/lib/installation/console/plugins.rb
```diff
@@ -0,0 +1,48 @@
...
+        log.debug "All found plugins: #{plugins.inspect}"
+
+        # remove the duplicates, this ensures the Y2DIR precedence
+        plugins.uniq! do |f|
+          File.basename(f)
+        end
+
+        plugins.each do |p|
+          log.info "Loading plugin #{p}"
+          require p
```
jreidinger, line 43: exception handling if plugin is broken?

src/lib/installation/console/plugins/console_button.rb
```diff
@@ -0,0 +1,58 @@
...
+        def initialize
+          textdomain "installation"
+        end
+
+        def label
+          _("Expert Console...")
+        end
+
+        def handle
+          require "installation/console"
```
jreidinger, line 33: why lazy loading?

src/lib/installation/console/plugins/console_button.rb
```diff
@@ -0,0 +1,58 @@
...
+          _("<p>The <b>Expert Console</b> button starts a command line interface " \
+            "to the installer. It is intended for special purposes, wrong usage " \
+            "might result in crash or unexpected behavior.</p>")
+        end
+      end
+
+      # define the plugin
+      class ConsoleButtonPlugin < MenuPlugin
+        def widget
+          ConsoleButton.new
```
jreidinger, line 48: how often it is called? Maybe not creating always new instance would be better?

src/lib/installation/console/plugins/shell_command.rb
```diff
@@ -0,0 +1,30 @@
...
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+
+module Installation
+  module Console
+    # define the "shell" command in the expert console
+    class Commands
```
jreidinger, line 17: this is kind of strange. So plugin in fact inject code into Commands class? This is quite ugly I expect it will be own object that has defined API like command and its description and potentionally more in future.

src/lib/installation/console/gui.rb
lslezak: I do not know, Wayland has some support for X applications, https://wayland.freedesktop.org/docs/html/ch05.html, maybe it will work.

Anyway, this is irrelevant until we switch the installer to Wayland, lets solve that later...

src/lib/installation/console/gui.rb
lslezak: I already tried to use `&.` but Rubocop then complained, we would need to switch to the newer Rubocop. So out of scope of this PR...

src/lib/installation/console/gui.rb
lslezak: Um, yes, we could remove the check...

src/lib/installation/console.rb
lslezak: That looks pretty stable, the latest change there is 3 years old, some parts of the code are there for 20 years. :smiley: 
https://github.com/ruby/ruby/blame/master/lib/irb/init.rb#L16-L26

src/lib/installation/console/commands.rb
lslezak: Yes, that might be possible, I'll check later...

src/lib/installation/console/gui.rb
lslezak: I haven't tested that, that dialogs are used rarely if ever... I'll check later...

src/lib/installation/console/plugins/console_button.rb
lslezak: I expect the console will be used rarely so all the code which handles the console does not need to be loaded. Just to save some memory in the installer...

src/lib/installation/console/plugins/console_button.rb
lslezak: This is called only when you enter the configuration dialog, which means almost never in a typical installation.

src/lib/installation/console/plugins/shell_command.rb
lslezak: OK, I was thinking a lot about the plugins API and I could not find any example our use case.
Lets improve it later...

src/lib/installation/console/commands.rb
lslezak: The documentation is still missing, see the TODO section, will be added later.

jreidinger (approved):

PR #909, opened 2021-02-03
Update documentation to reflect yast-storage-ng (bsc#1175098)

ancorgs:
## Problem

The documents about the storage proposal were messy because they were left in a state where they documented the transformation process from yast-storage into yast-storage-ng instead of focusing on the status quo.

1) The reference documentation of the control file (in this repo) was still describing the old yast-storage format
2) The more detailed documentation at yast2-storage-ng was written as some kind of list of differences with the obsolete old format, instead of simply describing the present.

## Solution

The solution comes in two pull requests: this one and https://github.com/yast/yast-storage-ng/pull/1205

This one updates the information in the control file documentation, ditching the yast-storage content and replacing it with reference information that was originally found at the yast-storage-ng document.

## Review

Better see the rendered result here: https://github.com/yast/yast-installation/blob/3346fda5afeea35cd9669f6f7260a50d76bcd260/doc/control-file.md

### DIFF
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -869,7 +869,9 @@ Example:
         <id>virtualization_host_kvm</id>
         <!-- partitioning override -->
         <partitioning>
-          <proposal_lvm config:type="boolean">true</proposal_lvm>
+          <proposal>
+            <lvm config:type="boolean">true</lvm>
+          </proposal>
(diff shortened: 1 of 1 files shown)


dgdavid (approved):
LGTM, although additional review would be nice :)

PR #926, opened 2021-03-09
Reduce self-update memory consumption

imobachgs:
This PR is a first attempt to reduce the amount of memory used by the self-update mechanism. It introduces the following changes:

* Download each package only once. The installer update repositories can contain several versions of the same package. It causes YaST2 to download each package several times. For instance, if there are two packages `some-package-1.0` and `some-package-2.0`, YaST2 will download `some-package-2.0` twice.
* Put all files from the same update repository in a single squashfs filesystem (instead of using one per package).
* Do not include unchanged files and directories like `/usr/share/{man,info,...}`.

Trello: https://trello.com/c/6RmucVFu/
Bug: [bsc#1182928](https://bugzilla.suse.com/show_bug.cgi?id=1182928)

## Some numbers

* Original: 47MB used by the squashed files; 133MB uncompressed = 180MB.
* Patches: 9MB used by the squashed files; 27MB uncompressed = 35MB.

## And now what?

We need to do a lot of testing. It is a critical feature and it should work just in SP3. The solution for old versions would be different.


wfeldt:
Looks ok so far.

src/lib/installation/update_repository.rb
```diff
@@ -140,17 +140,24 @@ def repo_id
...
     # @return [Array<Y2Packager::Resolvable>] List of packages to install
     #
     # @see Y2Packager::Resolvable
     def packages
       return @packages unless @packages.nil?
+
       add_repo
-      @packages = Y2Packager::Resolvable.find(kind: :package, source: repo_id).sort_by!(&:name)
+      all_packages = Y2Packager::Resolvable.find(kind: :package, source: repo_id)
+      pkg_names = all_packages.map(&:name).uniq
```
lslezak, line 155: I'd add a short comment here like

```ruby
# find the highest version of each package
```

src/lib/installation/update_repository.rb
```diff
@@ -373,6 +352,36 @@ def repo_status
...
+    # * they are identical to the original files (no real changes)
+    #
+    # @param dir [String,Pathname] Directory to clean-up
+    def clean_unneeded_files(dir)
+      top = Pathname.new(dir)
+
+      ignored_dirs = IGNORED_DIRS.map { |d| top.join(d) }.select(&:directory?)
+      ignored_dirs.each { |d| FileUtils.rm_r(d) }
+
+      top.glob("**/*").select(&:file?).each do |path|
```
lslezak, line 377: NP: How long is the complete file list? Building a list with many items could require a lot of memory, maybe we should process the files one-by-one by using the [Find.find](https://ruby-doc.org/stdlib-2.5.0/libdoc/find/rdoc/Find.html) call...

lslezak (commented):
LGTM so far, just some minor comments....

src/lib/installation/update_repository.rb
```diff
@@ -377,6 +361,44 @@ def repo_status
...
+      top.find do |path|
+        next if !path.file? || path.symlink?
+
+        instsys_path = Pathname.new("/").join(path.relative_path_from(top))
+        next unless instsys_path.exist?
+
+        path.unlink if FileUtils.identical?(path, instsys_path)
+      end
+
+      top.find.to_a[1..-1].reverse.each do |path|
```
wfeldt, line 395: Probably a strange case but `[1..-1]` would give nil for an empty package (without file content). Or?

wfeldt: Uhm, just noticed that's after deleting stuff - so it's actually a bit more likely to happen...

imobachgs: Good catch. It is a strange case, but better be prepared.

wfeldt (approved):
lgtm

src/lib/installation/update_repository.rb
wfeldt: I see; ok then.

PR #932, opened 2021-03-11
Remove the libzypp cache symlink (bsc#1182928)

lslezak:
- Remove the libzypp cache symlink when YaST is restarted
- Related to https://github.com/yast/yast-packager/pull/559
- Improved `memsample` script handling
    - Do not start it again if it is already running (might happen if YaST is started again after crash)
   - Stop it when YaST finishes
- 4.3.34

### DIFF
--- a/startup/First-Stage/F09-start
+++ b/startup/First-Stage/F09-start
@@ -17,9 +17,16 @@ callHooks preFirstCall
 # Using /dev/null - If there is nothing to do, let it fail silently
 SUSEConnect --cleanup > /dev/null 2>&1
 
-# delete the repositories accidentally saved into inst-sys
+# delete the repositories saved into inst-sys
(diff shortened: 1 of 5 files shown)


startup/First-Stage/F09-start
```diff
@@ -17,10 +17,17 @@ callHooks preFirstCall
 # Using /dev/null - If there is nothing to do, let it fail silently
 SUSEConnect --cleanup > /dev/null 2>&1
 
-# delete the repositories accidentally saved into inst-sys
+# delete the repositories saved into inst-sys
```
lslezak, line 20: JFYI: now we deliberately save the repositories during the installation workflow to inst-sys (to not loose them in some situations where we need to restart libzypp)

mvidner (approved):

PR #935, opened 2021-03-18
Optional in-process memory profiling: MEMORY_PROFILER=1 and MASSIF=1

mvidner:
- https://bugzilla.suse.com/show_bug.cgi?id=1182649
- https://trello.com/c/oj8K4EPu

Allow memory profiling of the main installer process, via  boot parameters: 
- MASSIF=1 enables Valgrind/Massif (C/C++ level)
- MEMORY_PROFILER=1 enables Ruby level

see [doc/memory-profile-valgrind-massif.md](https://github.com/yast/yast-installation/blob/44835c83560a9820df424ca88cc0f32c90e5abbf/doc/memory-profile-valgrind-massif.md)

The necessary packages need to be present:
- Massif is pulled in with `extend=gdb`, done by https://github.com/openSUSE/installation-images/pull/506 (merged)
  (to test this before i-i is updated, extract /usr/bin/debuginfod-find from a TW debuginfod-client.rpm and copy it to a live inst-sys)
- memory_profiler.gem is TODO

### DIFF
--- a/doc/memory-profile-valgrind-massif.md
+++ b/doc/memory-profile-valgrind-massif.md
@@ -4,7 +4,7 @@ Valgrind is an instrumentation framework that runs a program by simulating
 every instruction.  It has several tools. Memcheck is the original one, used
 for detecting memory errors. Massif is a memory profiler.
 
-(The output it produces, example: https://gist.github.com/mvidner/4e8ed01c7dabb648a50e0dd5f0fdcc62 )
+(The output it produces, example: https://github.com/yast/yast-installation/wiki/Massif-Sample-Output )
 
 The basic invocation of Massif is simple,
 `valgrind --tool=massif my_program its_arguments`, so for YaST it is
--- a/doc/memory-profile-valgrind-massif.md
+++ b/doc/memory-profile-valgrind-massif.md
@@ -1,0 +1,50 @@
+## Massif Basics
+
+Valgrind is an instrumentation framework that runs a program by simulating
+every instruction.  It has several tools. Memcheck is the original one, used
+for detecting memory errors. Massif is a memory profiler.
+
+(The output it produces, example: https://gist.github.com/mvidner/4e8ed01c7dabb648a50e0dd5f0fdcc62 )
+
+The basic invocation of Massif is simple,
+`valgrind --tool=massif my_program its_arguments`, so for YaST it is
+
+```console
+# valgrind --tool=massif /usr/lib/YaST2/bin/y2start sw_single qt
+```
+
+Or, if we don't want to dig in the startup scripts, we use a bigger gun:
+
+```console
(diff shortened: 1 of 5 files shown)


lnussel:
Is there a way to stream that memory stuff to syslog instead of files? that way it would be possible to stream it live to a different system using loghost linuxrc parameter? Might even be possible to visualize live then :)

mvidner:
@lnussel that sounds interesting but I have to admit I am not familiar with remote logging. Do you have in mind an existing setup that I could learn from?

That said, valgrind seems to have its own remote logging mechanism via `valgrind-listener`.

reference: https://en.opensuse.org/SDB:Linuxrc#p_loghost

lnussel:
In the instsys the loghost parameter activates this script: https://github.com/openSUSE/installation-images/blob/master/data/initrd/scripts/remote_log_setup

On the server side I used a simple listener based on https://gist.githubusercontent.com/alejandroliu/3ddd8433f00d221eac76/raw/0dc491a51475039049d3b0f9b4e171ee437e2aa5/syslogd.pl

Since it's possible to specify a port for loghost now you don't need to be root on the listener side.

doc/memory-profile-valgrind-massif.md
```diff
@@ -0,0 +1,50 @@
+## Massif Basics
+
+Valgrind is an instrumentation framework that runs a program by simulating
+every instruction.  It has several tools. Memcheck is the original one, used
+for detecting memory errors. Massif is a memory profiler.
+
+(The output it produces, example: https://gist.github.com/mvidner/4e8ed01c7dabb648a50e0dd5f0fdcc62 )
```
lslezak, line 7: Um, I do like much links to some external gists which might disappear at any time...

What about copying that here into https://github.com/yast/yast-installation/wiki ?

doc/memory-profile-valgrind-massif.md
```diff
@@ -0,0 +1,50 @@
...
+
+The backtraces will contain some names, at the boundaries of shared libraries
+But to see names inside libraries Massif needs debuginfo files.
+
+## Massif at Installation Time
+
+tl;dr: Boot with `extend=gdb MASSIF=1`.
+
+1. Install Massif. It is part of the [**gdb** extension][gdb-ext], so use
+   `extend=gdb` at the boot prompt, or in an inst-sys shell.
```
lslezak, line 40:
```suggestion
   `extend=gdb` at the boot prompt, or run `extend gdb` in an inst-sys shell.
```

doc/memory-profile-valgrind-massif.md
mvidner: Good point, will do.

lslezak (approved):

PR #936, opened 2021-03-29
Expert console: fixed "shell" command

lslezak:
- Fixes frozen `vim` when running from shell in the expert console in text mode
  - https://bugzilla.suse.com/show_bug.cgi?id=1183652
  - It turned out to be a `fbiterm` issue, it can be reproduced with a simple `fbiterm vim` command
  - The workaround is to set `TERM=vt100`
- Improved `shell` in GUI (Qt)
  - The previous fix to avoid "cannot set terminal process group" error messages printed by `bash`
     was to use `dash` instead (see PR #934)
  - But it turned out that `dash` is very limited, for example it does not suport <Tab> auto completion so it is quite
    difficult to use it as an interactive shell
  - The fix is to run `bash` but in a separate `xterm` window so it can activate the job control correctly

### DIFF
--- a/src/lib/installation/console/plugins/shell_command.rb
+++ b/src/lib/installation/console/plugins/shell_command.rb
@@ -28,8 +28,9 @@ module Installation
       def tui_shell
         puts quit_hint
 
-        # some interactive tools like "vim" get stuck when running in "fbiterm",
-        # the workaround is to override TERM to "vt100" (bsc#1183652)
+        # some interactive tools like "vim" get stuck when running in "fbiterm"
+        # "fbiterm" sets TERM to "iterm", the workaround is to override it
+        # to "vt100" (bsc#1183652)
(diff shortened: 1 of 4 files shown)


src/lib/installation/console/plugins/shell_command.rb
```diff
@@ -16,24 +16,47 @@ module Console
...
       end
 
     private
 
+      def tui_shell
+        puts quit_hint
+
+        # some interactive tools like "vim" get stuck when running in "fbiterm",
+        # the workaround is to override TERM to "vt100" (bsc#1183652)
+        term = ENV["TERM"] == "iterm" ? "TERM=vt100" : ""
```
jreidinger, line 33: so fbiterm use TERM=iterm? it is not so straighforward.

jreidinger (approved):
beside that need to document that fbiterm use iterm looks good.

PR #937, opened 2021-04-07
Hide the abort button when calling the network client

dgdavid:
### Problem

> When user presses Abort button on Network settings page, it works same way as "Back" button and returns to the previous dialog. — https://bugzilla.suse.com/show_bug.cgi?id=1183586

As stated in the [second comment](https://bugzilla.suse.com/show_bug.cgi?id=1183586#c2), this happens

>  because the network **subworkflow** should not use [Abort] (...) It should be possible to **abort only in the main workflow**.

### Solution

To use an installation argument (`hide_abort_button`) when calling the _inst\_lan_ client to force it hiding the abort button during the installation. **See https://github.com/yast/yast-network/pull/1195** for more details.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Apr  7 10:54:58 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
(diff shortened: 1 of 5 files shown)


teclator (approved):
LGTM

PR #939, opened 2021-04-07
Stop owning bindir in the filelist - this is owned by filesystem package

dirkmueller:


ancorgs:
TL;DR: closed in favor of #944 

I know this has been opened for 7 days without a review. Sorry for that. I suspect this pull request is not so appealing to review since it's hard to spot the real meaningful changes with so many lines shuffled up and down. For the time being, I have already found that apart from the fully expected changes in the `%files` section, the following line was also removed with no clear explanation:

```
BuildRequires:  yast2-bootloader
```

Since the commits in #943 also fixed the main problem but were way more to the point, we have reorganized all the pending changes at #944. I'm closing all the other related pull requests. Feel free to open a separate pull request for the spec cleaning alone if you feel is important enough or to incorporate any other change I may have overlooked in your PR (ie. I failed to include it in #944).

Thanks and sorry for the noise.

PR #940, opened 2021-04-08
check for usr/lib/modules to cover usrmerge (boo#1029961)

lnussel:


lnussel:
have to wait for #939 before adding package files as it also modifies the spec

dirkmueller:
I've merged this PR into #939

lnussel:
merged into #939

PR #942, opened 2021-04-13
Use empty string instead of 'default' as the 'Default' id

teclator:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1184277

## Solution

- Use empty string instead of 'default' as the 'Default' id being shown correctly in the proposal when there is no polkit_default_privs selected in the profile. 

## Screenshot

PolicyKitDefault
ComboBox
PolicyKitComboBoxHelp


jreidinger (approved):

jreidinger (approved):

teclator:
:heavy_check_mark: Created IBS submit manually [request #239420](https://build.suse.de/request/show/239420)

PR #943, opened 2021-04-14
package: Do not own system directories

DimStar77:


DimStar77:
close in favor of more complete https://github.com/yast/yast-installation/pull/939

PR #946, opened 2021-04-16
Split #1 out of five from PR 939

dirkmueller:


package/yast2-installation.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Apr 16 08:39:17 UTC 2021 - Dirk Müller <dmueller@suse.com>
+
+- remove check for non-systemd distros
+- first out of 5 splits of "spec-cleaner -m -i *ec" run (for easier review)
+- 4.4.4
```
ancorgs, line 6: If you plan to create por PRs, you don't need to increase the version in each of them. Increasing the version is what triggers a new submit request in OBS. It's fine to accumulate changes in Github (with their own changelog entries if you wish) and then increase the version only in the last to trigger the continuous integration.

dirkmueller: removed version bump again

ancorgs (approved):
LGTM

PR #949, opened 2021-04-27
Drop Yast::LanItems dependency

teclator:
## Problem

We have been preparing the drop of **Yast::LanItems** adapting the some code in network which is still used by this module so we need to adapt it to.

- see https://github.com/yast/yast-network/pull/1216
- https://bugzilla.suse.com/show_bug.cgi?id=1185338

## Solution

- Drop Yast::LanItems dependency completely.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Apr 27 09:23:17 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Remove Yast::LanItems dependency (bsc#1185338)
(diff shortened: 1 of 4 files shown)


teclator:
Probably would be a good idea to increase the yast2-network dependency using the one used by https://github.com/yast/yast-network/pull/1217 but as the code is already in Factory we probably could merge it as it is.

imobachgs (approved):

mchf (approved):
It turned into nice code de-duplication ;-)

LGTM

teclator:
Updated dependency, it needs https://github.com/yast/yast-network/pull/1217

PR #957, opened 2021-05-18
Do not try to open a X display forever (bsc#1185095)

teclator:
## Problem

It could take some time until the X Server is ready and it could be even open to an infinite loop. While we are waiting a blank screen is shown without any hint about what is going on.

- https://trello.com/c/4DpZZook/2441-3-osdistribution-p3-1185095-auto-installation-stalls-at-start-of-stage-2-icewm-package-not-installed

## Solution

Fallback to **ncurses** in case that open a X display fails after 15 attempts / seconds. In case of **VNC** the loop will maintained until success by now.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue May 18 12:39:25 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Fallback to ncurses when a X display is not opened after 15
+  seconds (bsc#1185095)
+- 4.4.7
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


dgdavid (approved):
LGTM

package/yast2-installation.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue May 18 12:39:25 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Fallback to ncurses when a X display is not opened after 15
+  seconds (bsc#1185095)
```
dgdavid, line 5: seconds=? Do you mean, _attempts_?

teclator: Currently we sleep one second per attempt, so by now we are really waiting 15 seconds, I preferred to write in that way in the changelog.

PR #958, opened 2021-05-19
Avoid conflicts with filesystem and hicolor-icon-theme (bsc#1184786)

dirkmueller:


jreidinger (approved):

jreidinger:
lets try it as rpm build passed, I think it should be safe enough to give it a try.

PR #960, opened 2021-05-27
yupdate: Do not create overlays for non-existing directories

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1183795#c21
- The default overlay list includes the `/usr/lib64/YaST2` directory but that does not exist on 32-bit systems and it fails.
- If the target directory does not exist do not try creating an overlay for it, just skip it.
- Technical note: `File.writable?` returns `false` for non-existing paths, you have to explicitly check that the directory exists to catch this case.

### DIFF
--- a/bin/yupdate
+++ b/bin/yupdate
@@ -191,7 +191,11 @@ HELP
       "/usr/share/icons"
     ].freeze
 
-    attr_reader :dir, :orig_dir
+    # @return [String] the original requested directory
+    attr_reader :orig_dir
+
+    # @return [String] expanded requested directory (i.e. symlink target)
+    attr_reader :dir
 
     # manage the OverlayFS for this directory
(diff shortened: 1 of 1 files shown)


bin/yupdate
```diff
@@ -204,7 +204,8 @@ HELP
 
     # create an OverlayFS overlay for this directory if it is not writable
     def create
-      return if File.writable?(dir)
+      # skip non-existing or already writable directories
+      return if !File.directory?(f) || File.writable?(dir)
```
mvidner, line 208:
```suggestion
      return if !File.directory?(dir) || File.writable?(dir)
```

BTW, having both `orig_dir` and `origdir` without comments sounds like future trouble.

mvidner (approved):

PR #961, opened 2021-05-31
Allow to set IP forwarding defaults through the control file (bsc#1186280)

teclator:
## Problem

There is no way to influence IP forwarding defaults settings during installation, something that would be nice to be have (see https://github.com/yast/yast-network/pull/1229)

- https://bugzilla.suse.com/show_bug.cgi?id=1186280
- https://trello.com/c/Jao9dDKQ/2477-5-make-forwarding-configurable-for-installer

## Solution

- Modify IP forwarding network configuration using the defaults defined in the control file when a new role is selected.

It depends on yast/yast-network#1229

### DIFF
--- a/src/lib/installation/select_system_role.rb
+++ b/src/lib/installation/select_system_role.rb
@@ -22,6 +22,7 @@ require "ui/installation_dialog"
 require "ui/text_helpers"
 require "installation/services"
 require "installation/system_role"
+require "y2network/proposal_settings"
 
 Yast.import "GetInstArgs"
 Yast.import "Packages"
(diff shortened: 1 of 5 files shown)


src/lib/installation/select_system_role.rb
```diff
@@ -259,6 +261,8 @@ def select_packages
 
     # for given role sets in {::Installation::Services} list of services to enable
     # according to its config. Do not use alone and use apply_role instead.
+    #
+    # FIXME: duplicate code?
```
imobachgs, line 265: From a reviewer's perspective, it is not clear why (or what) is duplicated. Could you clarify, please?

imobachgs (approved):

src/lib/installation/select_system_role.rb
teclator: In fact wanted to discuss it with you :) as it seems the code could be a duplicate of SystemRole#adapt_services

imobachgs (approved):

PR #962, opened 2021-06-01
Document network section

teclator:
Document the `network` section

- https://trello.com/c/Jao9dDKQ/2477-5-make-forwarding-configurable-for-installer
- https://bugzilla.suse.com/show_bug.cgi?id=1186280

### DIFF
--- a/doc/control-file.md
+++ b/doc/control-file.md
@@ -1236,16 +1236,15 @@ This is the full list of SLE-15 SP2:
 
 ## Network
 
-The network sections contains some options that allows to modify network 
-configuration defaults.
+Some options to modify network configuration defaults.
 
   * `force_static_ip` *(boolean, default: `false`)*
-    New connections will be initialized using an static bootproto.
+    Initialize new connections using a static bootproto.
   * `startmode` *(string)*
     When to bring up an interface. Possible values are `hotplug`, `auto`, `ifplugd`, 
     `manual`, `nfsroot` or `off`.
   * `network_manager`
-    Whether NetworkManager should be the default backend to be used or not.
+    Whether NetworkManager should be the default backend.
     Possible values are  `always`, `never` or `laptop` which will be true in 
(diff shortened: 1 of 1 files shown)


doc/control-file.md
```diff
@@ -1224,6 +1234,25 @@ This is the full list of SLE-15 SP2:
 </subvolumes>
 ```
 
+## Network
+
+The network sections contains some options that allows to modify network
```
imobachgs, line 1239: "some options that allow" or simply "some options to modify"

doc/control-file.md
```diff
@@ -1224,6 +1234,25 @@ This is the full list of SLE-15 SP2:
 </subvolumes>
 ```
 
+## Network
+
+The network sections contains some options that allows to modify network 
+configuration defaults.
+
+  * `force_static_ip` *(boolean, default: `false`)*
+    New connections will be initialized using an static bootproto.
```
imobachgs, line 1243: What about "Initialize new connections using a static bootproto"?

doc/control-file.md
```diff
@@ -1224,6 +1234,25 @@ This is the full list of SLE-15 SP2:
...
+  * `startmode` *(string)*
+    When to bring up an interface. Possible values are `hotplug`, `auto`, `ifplugd`, 
+    `manual`, `nfsroot` or `off`.
+  * `network_manager`
+    Whether NetworkManager should be the default backend to be used or not.
```
imobachgs, line 1248: "Whether NetworkManager should be the default backend."

imobachgs (commented):
Some minor comments. Otherwise, it looks good.

doc/control-file.md
teclator: I will adapt with the suggestion but was event tempted to remove it completely as currently the real behavior is only about "showing an extra" paragraph as part of the help text. But the option is still there.

imobachgs (approved):

PR #963, opened 2021-06-07
[SP3] Improve evaluating the old repositories (bsc#1185822)

lslezak:
## The Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1185822
- During migration usually the new repositories have a different URL so they point to new package versions (e.g. the Basesystem Module for SLE15-SP2 is replaced with SLE15-SP3 repository with different name and URL)
- During migration YaST offers to remove the old repositories because the old repositories/packages could potentially cause conflicts or other dependency problems
- But the detection of the old repositories does not work properly if the new repository is the same as the old one (X replaced by X), in that case the repository looks like as if it was old and should not be used anymore

## The Solution

- The improved code loads the current (SP3) registered repositories and these repositories are excluded from the old ones

### Testing

- Tested manually, see below

### Screenshots

- The original behavior: the "new" SP3 repositories with the NVidia module are preselected for removal
migration_old_repos_orig

- The improved behavior: only the old SP2 repository is listed
migration_old_repos_fixed

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -64,3 +64,8 @@ Style/VariableName:
     - src/lib/**/*.rb
   Exclude:
     - src/lib/installation/clients/*.rb
+
+Style/PredicateName:
+  Exclude:
+    # mocked Registration.is_registered?
+    - test/lib/upgrade_repo_manager_test.rb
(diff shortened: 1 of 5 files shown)


imobachgs (approved):

lslezak:
Submitted to SP3 manually in https://build.suse.de/request/show/242399

PR #965, opened 2021-06-08
Write proxy settings at the end of the First Stage and export https_proxy

teclator:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1185016
- https://trello.com/c/s9gfK64m/2499-sles15-sp2-p2-1185016-l3-autoyast-does-not-set-proxy-properly-ref00d1iglod5001ixjkyhref

During installation, the installation fails if http and https traffic is restricted allowing only connnections through a proxy. The current proxy settings export the **http_proxy** variable but not the **https_proxy** which is used for downloading the online updates.

- https://github.com/yast/yast-packager/blob/master/src/lib/y2packager/clients/inst_productsources.rb#L482

Leap15.3_proxy

Related: 

- https://bugzilla.suse.com/show_bug.cgi?id=305163
- https://bugzilla.suse.com/show_bug.cgi?id=923788
- https://bugzilla.suse.com/show_bug.cgi?id=869640 (remove old inst_network_setup.rb client)

## Solution

Export also the **https_proxy** variable when proxy settings are established through **linuxrc** as it was already done by the old network setup client (removed here https://github.com/yast/yast-installation/pull/189/files#diff-c8deba4573493d51e3e8715788d69a6413de22a753ed38706e851bc261763087L1431)

Screenshot_testing_2021-06-08_11:19:57

## Tests

- Added unit test
- Tested manually with the changes in the other repositories (yast/yast-proxy#40 , yast/yast-network#1231 and yast/yast-autoinstallation#767) with a nomal installation and with AutoYaST (using before_proposal and without before_proposal)

### DIFF
--- a/src/lib/installation/clients/proxy_finish.rb
+++ b/src/lib/installation/clients/proxy_finish.rb
@@ -32,10 +32,13 @@
 #
 module Yast
   class ProxyFinishClient < Client
+    include Yast::Logger
+
     def main
       textdomain "installation"
 
       Yast.import "Stage"
+      Yast.import "Proxy"
 
       @ret = nil
       @func = ""
@@ -64,16 +67,15 @@ module Yast
         }
       elsif @func == "Write"
         if Stage.initial
-          @proxy = Convert.to_string(SCR.Read(path(".etc.install_inf.ProxyURL")))
+          # In case of written by linuxrc or by AutoYaST it should be copied to the target system
+          return if Proxy.GetEnvironment.reject { |_, v| v.to_s.empty? }.empty?
 
-          if !@proxy.nil?
-            Builtins.y2milestone("setting proxy to %1", @proxy)
+          ex = Proxy.Export
+          log.info("Writing proxy settings to the target system: #{ex.inspect}")
 
-            # maybe use Proxy module
-            SCR.Write(path(".sysconfig.proxy.HTTP_PROXY"), @proxy)
-            SCR.Write(path(".sysconfig.proxy.FTP_PROXY"), @proxy)
-            SCR.Write(path(".sysconfig.proxy"), nil)
-          end
+          Proxy.Import(ex)
+          Proxy.WriteSysconfig
+          Proxy.WriteCurlrc
         end
       else
         Builtins.y2error("unknown function: %1", @func)
--- a/src/lib/installation/clients/proxy_finish.rb
+++ b/src/lib/installation/clients/proxy_finish.rb
@@ -66,7 +66,7 @@ module Yast
           "when"  => [:installation, :update, :autoinst]
         }
       elsif @func == "Write"
-        if Stage.initial && Proxy.to_target
+        if write_to_target?
           proxy_settings = Proxy.Export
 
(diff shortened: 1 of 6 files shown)


src/lib/installation/clients/proxy_finish.rb
```diff
@@ -32,10 +32,13 @@
 #
 module Yast
   class ProxyFinishClient < Client
+    include Yast::Logger
+
     def main
       textdomain "installation"
 
       Yast.import "Stage"
+      Yast.import "Proxy"
```
mvidner, line 41: I've checked that yast2-installation.rpm already Requires yast2-proxy.rpm :heavy_check_mark:

mvidner (approved):

src/lib/installation/clients/proxy_finish.rb
teclator: Yep, but will need to update the dependency. 

I tested the patch and there is a problem when we obtain the environment variables, so have modified it. Basically it reads the configuration again in case of not modified and the config is read from the target system instead of the inst-sys.

I was thinking in the best way to determine if we should write the settings to the target system or not taking in account also AutoYast and to be honest didn't like any solution in particular, so by now will introduce a flag in the **Proxy** module.

src/lib/installation/clients/proxy_finish.rb
```diff
@@ -64,16 +67,15 @@ def main
         }
       elsif @func == "Write"
         if Stage.initial
-          @proxy = Convert.to_string(SCR.Read(path(".etc.install_inf.ProxyURL")))
+          # In case of written by linuxrc or by AutoYaST it should be copied to the target system
+          return if Proxy.GetEnvironment.reject { |_, v| v.to_s.empty? }.empty?
 
-          if !@proxy.nil?
-            Builtins.y2milestone("setting proxy to %1", @proxy)
+          ex = Proxy.Export
```
imobachgs: np: `ex` is not the most descriptive name :-)

src/lib/installation/clients/proxy_finish.rb
```diff
@@ -64,16 +67,15 @@ def main
...
+          log.info("Writing proxy settings to the target system: #{ex.inspect}")
 
-            # maybe use Proxy module
-            SCR.Write(path(".sysconfig.proxy.HTTP_PROXY"), @proxy)
-            SCR.Write(path(".sysconfig.proxy.FTP_PROXY"), @proxy)
-            SCR.Write(path(".sysconfig.proxy"), nil)
-          end
+          Proxy.Import(ex)
+          Proxy.WriteSysconfig
```
imobachgs, line 75: Why do we have two separates `Write*` methods?

teclator: I preferred to use both instead of Write because it checks if the config is modified or not and also is prepared to show a dialog and we are only interested in the write itself.

src/lib/installation/clients/proxy_finish.rb
teclator: It was already modified, it is currently **proxy_settings**

imobachgs (approved):

PR #975, opened 2021-08-19
Refactoring of Unmounting during Installation (umount_finish.rb)

shundhammer:
## Trello

https://trello.com/c/XaqRXabj/

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1149980

## Problem

There were quite a number of very weird and confusing messages in the y2log in this final step of executing the installation: Unmounting all the mounts to the newly installed target, i.e. `/mnt`.

In particular, unmounting some btrfs subvolumes like `/mnt/var` _always_ threw errors during this process because they are automatically unmounted along with their parent main btrfs; attempting to unmount them individually as long as that parent main btrfs is still mounted fails with an error "_target is busy_".

## Challenges

This `umount_finish.rb` installation step runs as one of the last steps when the installation is executed: After the user clicked through all the interactive installation steps, after the installation settings / proposal is confirmed, after partitions  and filesystems are created, after all the RPMs are installed to the new system, after log files etc. are copied over to it; after everything is done, and just before the machine is rebooted to boot the newly installed system.

It is hard and painful to test this becaues it takes so long to get there, and getting there also means heavily modifying the target machine. Copying and editing source files to see a change is also painful: Copy the file(s) to the inst-sys and bind-mount them individually so they replace existing files on the RAM disk; and after a reboot everything is gone.

## Making Life Easier

So this pull request introduces a whole new class to handle most of the more advanced things to do during that `umount_finish.rb` step: Collecting information what mounts are active on the target system, determining which ones to actually unmount, and in what order.

This is all not really rocket science, but it needs to be done carefully, and we need to be able to test more advanced scenarios. Doing all that the hard way while an actual installation is being executed is hard, error-prone and painful.

So this introduces a new class that handles the logic for unmounting, that can be fed from existing scenarios in simple files, and that can report what it _would_ do without actually doing it; i.e. return a list of the paths that need to be unmounted, and return it in the correct order.

## The Unmounter

The new `Installation::Unmounter` is that class. By default, it reads `/proc/mounts` go get the information what is currently mounted directly from the currently running kernel. But it can also receive a file exactly like `/proc/mounts` that was previously copied from another system and use that for its calculations.

It has a method `unmount_paths` that returns the paths, a method `ignored_paths` to see which mounts were deliberately ignored (including the btrfs subvolumes and of course those paths that don't start with `/mnt`), and it has a method `execute` that actually calls the external `umount` command for each of the paths from `unmount_paths`.

This is meant as a drop-in replacement in `umount_finish.rb`, but unlike that, it has unit tests to test how it works, and there are common scenarios for its unit tests:

- partition-based btrfs for root with or without a separate `/home`
- a separate `/home` also on btrfs or on xfs
- an encrypted LVM setup with btrfs for root with a separeate `/home`

## Test

Testing is still hard, but we need to do a lot less of it manually now because the unit tests for the unmounter already covers a lot of scenarios.


src/lib/installation/unmounter.rb
```diff
@@ -0,0 +1,182 @@
...
+    # if it meets the criteria (mount prefix)
+    #
+    # @param line [String] one line of /proc/mounts
+    # @return [Mount,nil] parsed mount if relevant
+    #
+    def add_mount(line)
+      mount = parse_mount(line)
+      return nil if mount.nil? # Empty or comment
+
+      if !mount.mount_path.start_with?(@mnt_prefix)
```
mvidner, line 105: What if I have these?

```
/mnt/a
/mnt/b
/mnt2/c
```

A string-wise start_with will be wrong for the last one. Please add a test for this.

I thought [Pathname](https://ruby-doc.org/stdlib-2.7.2/libdoc/pathname/rdoc/Pathname.html) would have a start_with method but it doesn't. Perhaps it would work if our constructor made sure `mnt_prefix` ends with a slash?

src/lib/installation/unmounter.rb
```diff
@@ -0,0 +1,182 @@
...
+  #
+  # Without specifying a file to read as the second parameter in the
+  # constructor, it will default to /proc/mounts which is the right thing for
+  # real life use.
+  #
+  class Unmounter
+    include Yast::Logger
+    # @return [Array<Mount>] Relevant mounts to unmount
+    attr_reader :mounts
+    # @return [Array<Mount>] Ignored mounts
```
mvidner, line 45:
```suggestion
    # @return [Array<Mount>] Ignored mounts (not starting with the given prefix)
```

mvidner:
Makes sense, I like the API (and its docs!)
Just an implementation gotcha with substrings.
Also, `column -t` rules!
> `column -t -o\|` rules even more but is arguably less useful ;-)

src/lib/installation/unmounter.rb
shundhammer: The slash is dangerous: It excludes the root filesystem of the target from the results; that will be mounted to `/mnt`.

But what we could check here would be a regexp with the prefix and at the end either the end of string or a slash. That should work.

src/lib/installation/unmounter.rb
shundhammer: Yes, since a few hours that's even true; it also used to ignore subvolumes. I had been under the impression that unmounting a btrfs main volume also implicitly unmounts it subvolumes, but that turned out to be false. So, it used to be more ignored mounts than just those that don't start with the prefix; but in the meantime it's really only that much simpler condition.

src/lib/installation/unmounter.rb
shundhammer: Duh; that's so much simpler with just an additional one-liner testing if the path is equal to the prefix. Implemented.

src/lib/installation/unmounter.rb
shundhammer: Added.

src/lib/installation/unmounter.rb
```diff
@@ -112,6 +112,17 @@ def add_mount(line)
...
 
+    # Check if a mount should be ignored, i.e. if the path doesn't start with
+    # the mount prefix (usually "/mnt").
+    #
+    # @return [Boolean] ignore
+    #
+    def ignore?(mount)
+      return false if mount.mount_path == @mnt_prefix
+
+      !mount.mount_path.start_with?(@mnt_prefix + "/")
```
mvidner, line 123: What if I pass `/mnt/` as the prefix?

shundhammer: That would be wrong, I guess. ;-)

But yes, a trailing slash would be easy to remove in the constructor. And making sure that it's more than just one character, so `/` would still work; until it brings your system to a screeching halt because it would unmount everything, including the root filesystem. ;-)

shundhammer: Now removing one trailing slash if there is one.

src/lib/installation/unmounter.rb
```diff
@@ -0,0 +1,182 @@
...
+
+    # Unmounter constructor.
+    #
+    # @param mnt_prefix [String] Prefix which paths should be unmounted.
+    #
+    # @param mounts_file_name [String] what to use instead of /proc/mounts.
+    #   Use an empty string (not nil!) to not read anything at all yet
+    #   (but in that case, use read_mounts_file or add_mount later).
+    #
+    def initialize(mnt_prefix, mounts_file_name = nil)
```
ancorgs, line 76: Why not this?

```ruby
def initialize(mnt_prefix = "/mnt", mounts_file_name = "/proc/mounts)
```

Is more explicit (declaring the default values already in the signature of the method) and would produce the same result, isn't it?

shundhammer: Eh - actually, yes you are right. Going to change.

src/lib/installation/unmounter.rb
```diff
@@ -0,0 +1,196 @@
...
+      @mnt_prefix = mnt_prefix.dup || "/mnt"
+      @mnt_prefix.chomp!("/") unless @mnt_prefix == "/"
```
ancorgs, line 78: For this kind of things you could consider the usage of `Pathname` from Ruby's standard library.
https://ruby-doc.org/stdlib-2.7.4/libdoc/pathname/rdoc/Pathname.html

It's useful to make that kind of cleanup (even more extreme ones like the one below) but also to concatenate paths and that kind of stuff.

```ruby
Pathname.new("/crazy_stuff/.././mnt/").cleanpath.to_s #=> "/mnt/"
```

src/lib/installation/unmounter.rb
shundhammer: Done.

src/lib/installation/unmounter.rb
shundhammer: Right; in that case, it will also help even crazier prefixes like `/foo//bar///`; right now it's only fixing one single trailing slash, i.e. `/mnt/` . I hadn't thought of _Pathname_.

shundhammer: Done.

PR #981, opened 2021-09-29
Close the libzypp cache (bsc#1189793)

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1189793
- Close the libzypp cache to allow cleanly unmounting `/mnt/var/cache/zypp` directory

### Debugging

- I have manually tested with the debugger that after calling `Pkg.SourceFinishAll` the `/mnt/var` directory can be cleanly unmounted:
drop_zypp_cache

### Testing

- Tested manually with patched installer (using `yupdate`), the `y2log` did not contain any error and the `/mnt/var` directory was cleanly unmounted.


ancorgs (approved):
LGTM

PR #986, opened 2021-10-20
remove product workflow from welcome client

jreidinger:


src/lib/installation/clients/inst_complex_welcome.rb
```diff
@@ -76,10 +75,7 @@ def main
         result = handle_dialog_result(dialog_result)
         next unless result
 
-        return result if !available_products? || result != :next
-        # in the online installation the workflow is merged after registering the system
-        return result if Y2Packager::MediumType.online?
-        return merge_and_run_workflow
+        return result
```
imobachgs, line 78: You can do `return result if result` and remove the line above (`next unless result`). Isn't it?

jreidinger: yes, you are right

imobachgs (approved):

PR #994, opened 2021-11-24
Remove no longer used extra warning

dgdavid:
## Problem

There is a dead code that has been not used for a long time already. 

It used to append an extra warning paragraph to the confirmation presented to the user just before starting the real installation process when there was any destructive action in the list of commits of the old [_libstorage_](https://github.com/openSUSE/libstorage).

However, it has not been shown in any SLE15 product (in which libstorage has been replaced by [libstorage-ng](https://github.com/openSUSE/libstorage-ng)) and nobody seems to miss it. What is more, the generic 

> If you continue now, partitions on your
> hard disk will be modified according to the installation settings in the
> previous dialogs.

text looks informative enough for reminding the user that after proceeding there is no way back.

## Solution

Remove the code

## Tests

* Added small unit test to check the confirm dialog content
* Tested manually via [driver update](https://en.opensuse.org/SDB:Linuxrc#p_dud)

## Screenshots

| Proposal with some destructive actions | Warning shown before starting the installation |
| - | - |
| Proposal with some destructive actions | Warning shown before starting the installation |


src/include/installation/misc.rb
```diff
@@ -104,36 +99,15 @@ def confirmInstallation
...
-          # Text for confirmation popup before the installation really starts 2/3
-          body = Ops.add(
-            body,
-            _(
-              "<p>If you continue now, partitions on your\n" \
-                "hard disk will be modified according to the installation settings in the\n" \
-                "previous dialogs.</p>"
-            )
+        # Text for confirmation popup before the installation really starts 2/3
+        body = Ops.add(
```
ancorgs, line 103: This looks like an opportunity to kill a YCP zombie (`Ops.add`)

test/installation_misc_test.rb
```diff
@@ -1,13 +0,0 @@
-#! /usr/bin/env rspec
```
ancorgs, line 1: This test was basically emptied here at 631bca32be26ad6ad9d

I wonder what the intention was. @imobachgs Rings any bell?

imobachgs: No, sorry. I think we can get rid of it.

src/include/installation/misc.rb
dgdavid: Sure, let's kill them

src/include/installation/misc.rb
```diff
@@ -203,6 +109,51 @@ def confirmInstallation
...
+      result << _(
+        "<p>If you continue now, partitions on your\n" \
+          "hard disk will be modified according to the installation settings in the\n" \
+          "previous dialogs.</p>"
+      )
+      result << _(
+        "<p>Go back and check the settings if you are unsure.</p>"
+      )
+
+      result.join
```
ancorgs, line 129: NP: any reason to use an array and then call `join` instead of simply concatenating everything into a string?

dgdavid: Actually not. I'll change it.

ancorgs (approved):
LGTM

src/include/installation/misc.rb
```diff
@@ -203,6 +109,47 @@ def confirmInstallation
...
+    # @return [String]
+    def confirm_installation_text
+      result = ""
+
+      result << HTML.Heading(_("Confirm Installation"))
+      result << _("<p>Information required for the base installation is now complete.</p>")
+      result << _(
+        "<p>If you continue now, partitions on your\n" \
+          "hard disk will be modified according to the installation settings in the\n" \
+          "previous dialogs.</p>"
```
dgdavid, line 123: Just for the record, those "\n" should be _leftovers_ (maybe for not breaking translations?) since they are not effective in the `RichText` used to display the message in the confirmation dialog. There, only `<br>` works.

PR #1000, opened 2021-12-09
Self-update: do not scan the installation medium (bsc#1193536)

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1193536
- Do not scan the installation repository for the self update products, it takes very long time on the SLE Full medium
- Updated documentation
- See more details in https://github.com/yast/yast-registration/pull/552

### DIFF
--- a/doc/SELF_UPDATE.md
+++ b/doc/SELF_UPDATE.md
@@ -164,10 +164,10 @@ The URL of the update repository is evaluated in this order:
    - `$arch` - is replaced by the RPM package architecture used by the machine
      (like `x86_64` or `ppc64le`)
    - These variables are replaced by the value from the `/etc/os-release` file:
-     - `$os_release_name`       => `NAME`
-     - `$os_release_id`         => `ID`
-     - `$os_release_version`    => `VERSION`
-     - `$os_release_version_id` => `VERSION_ID`
+     - `$os_release_name`       => `NAME`       (e.g. `SLE`)
+     - `$os_release_id`         => `ID`         (e.g. `sle`)
+     - `$os_release_version`    => `VERSION`    (e.g. `15-SP4`)
+     - `$os_release_version_id` => `VERSION_ID` (e.g. `15.4`)
 
 The first suitable URL will be used. There are two exceptions:
 
--- a/doc/SELF_UPDATE.md
+++ b/doc/SELF_UPDATE.md
@@ -8,9 +8,9 @@ installation even after the media has been released. Check
 ### :information_source: Note
 
 The self update feature is removed from the updated SLE Service Pack media
-(the updated media released after the initial SP release), these media already
-contain the updated installer and the included updates could conflict with
-the self updates.
+(the updated media released after the initial SP release, so called "quarterly
+updates"), these media already contain the updated installer and the included
+updates could conflict with the self updates.
 
 If you need to patch the installer on these updated media you have to use
 a driver update disk (DUD).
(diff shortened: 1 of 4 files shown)


doc/SELF_UPDATE.md
```diff
@@ -157,6 +160,12 @@ The URL of the update repository is evaluated in this order:
    </globals>
    ```
 
+   This URL can contain these variables which are replaced by YaST:
+   - `$os_release_version` - is replaced by the `VERSION` value from the
```
imobachgs, line 164: Is the support for this variable already implemented?

lslezak: Yes, I'm just documenting the current status. The variable is [already used](https://github.com/yast/skelcd-control-leanos/blob/66c31f8dd531f2c626a61be503920db192dabed6/control/control.leanos.xml#L80).

doc/SELF_UPDATE.md
```diff
@@ -136,18 +136,21 @@ The URL of the update repository is evaluated in this order:
...
+   `control.xml` file in the `/globals/self_update_id` and
+   `/globals/self_update_version` XML nodes.
 
    ```xml
    <globals>
      <self_update_id>SLES</self_update_id>
    </globals>
    ```
 
+   If the `self_update_version` value is missing then the `VERSION_ID` value
```
mvidner, line 151: Are we really using both `VERSION` and `VERSION_ID`?

On Leap 15.1, they are both "15.1" but `man os-release` says

> VERSION=
> 
> A string identifying the operating system version, excluding any
> OS name information, possibly including a release code name, and
> suitable for presentation to the user. This field is optional.
> Example: "VERSION=17" or "VERSION="17 (Beefy Miracle)"".
> 
> VERSION_ID=
> 
> A lower-case string (mostly numeric, no spaces or other
> characters outside of 0–9, a–z, ".", "_" and "-") identifying the
> operating system version, excluding any OS name information or
> release code name, and suitable for processing by scripts or
> usage in generated filenames. This field is
> optional. Example: "VERSION_ID=17" or "VERSION_ID=11.04".

lslezak: Yes, we use both. :worried:  

The problem is that the default SLE self-update repository is `https://updates.suse.com/SUSE/Updates/SLE-INSTALLER/15-SP4/x86_64/update`, and the `/etc/os-release` contains this:

```
VERSION="15-SP4"
VERSION_ID="15.4"
```

So we need to use the `VERSION` here.

But the SCC expects the usual version in form `15.4`, se wee need to send the `VERSION_ID` value...

So we use them just because they fit those places, of course in ideal case we should use just one value everywhere. :thinking:

mvidner: Thanks, I thought this would be the reason but I did not have a SLE machine at hand.

doc/SELF_UPDATE.md
```diff
@@ -157,6 +160,12 @@ The URL of the update repository is evaluated in this order:
    </globals>
    ```
 
+   This URL can contain these variables which are replaced by YaST:
+   - `$os_release_version` - is replaced by the `VERSION` value from the
+     `/etc/os-release` file
```
mvidner, line 165:
```suggestion
     `/etc/os-release` file (e.g. 15-SP4).
```

doc/SELF_UPDATE.md
```diff
@@ -136,18 +136,21 @@ The URL of the update repository is evaluated in this order:
...
+   `/globals/self_update_version` XML nodes.
 
    ```xml
    <globals>
      <self_update_id>SLES</self_update_id>
    </globals>
    ```
 
+   If the `self_update_version` value is missing then the `VERSION_ID` value
+   from the `/etc/os-release` file is used as a fallback.
```
mvidner, line 152:
```suggestion
   from the `/etc/os-release` file is used as a fallback (e.g. 15.4).
```

lslezak:
I have added some more `/etc/os-release` replacements and added some missing mocks in the tests (to avoid running SLP discovery...).

imobachgs (approved):

mvidner (approved):

PR #1002, opened 2021-12-10
Document mount_options attribute

joseivanlopez:
See https://github.com/yast/yast-installation-control/pull/115.


dgdavid (approved):
LGTM

PR #1010, opened 2022-01-05
Do not propose Linux Security Module default config when declared as not configurable

teclator:
Do not propose **Linux Security Module** default configuration when it is declared as **not configurable** avoiding the write of the **kernel boot parameters** when are not needed / wanted. For more details see https://github.com/yast/yast-security/pull/117

### DIFF
--- a/src/lib/installation/security_settings.rb
+++ b/src/lib/installation/security_settings.rb
@@ -70,8 +70,8 @@ module Installation
     # When Linux Security Module is declared as configurable and there is no Module selected yet
     # it will select the desired LSM and the needed patterns for it accordingly
     def propose_lsm_config
-      return false unless lsm_config.configurable?
-      return false if lsm_config.selected
+      return unless lsm_config.configurable?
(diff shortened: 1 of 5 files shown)


src/lib/installation/security_settings.rb
```diff
@@ -67,6 +67,17 @@ def load_features
       load_feature(:polkit_default_privs, :polkit_default_privileges)
     end
 
+    # When Linux Security Module is declared as configurable and there is no Module selected yet
+    # it will select the desired LSM and the needed patterns for it accordingly
+    def propose_lsm_config
+      return false unless lsm_config.configurable?
+      return false if lsm_config.selected
```
dgdavid, line 74: NP: just `return`... or return `true` at the end of method too

dgdavid (approved):
LGTM, although I'd just `return` instead of `return false` when early exiting from the `propose_lsm_config` method.

dgdavid (approved):

PR #1011, opened 2022-01-07
Update yast2 dependency

teclator:
It needs https://github.com/yast/yast-yast2/pull/1227


dgdavid (approved):
LGTM, but let's wait until all checks are green.

PR #1012, opened 2022-01-12
New rubocop2

jreidinger:
Adapt to new rubocop version

### DIFF
--- a/bin/yupdate
+++ b/bin/yupdate
@@ -223,8 +223,8 @@ module YUpdate
       # through the bind mount and we would see the changed content here
       system("mount", "--make-private", origdir)
 
-      system("mount", "-t", "overlay", "overlay", "-o", "lowerdir=#{dir},"\
-                                                        "upperdir=#{upperdir},workdir=#{workdir}", dir)
+      system("mount", "-t", "overlay", "overlay", "-o",
+        "lowerdir=#{dir},upperdir=#{upperdir},workdir=#{workdir}", dir)
     end
 
     # delete the OverlayFS for this directory, all changes will be reverted back
--- a/bin/yupdate
+++ b/bin/yupdate
@@ -335,9 +335,9 @@ module YUpdate
       msg "Downloading #{url}"
 
       if block_given?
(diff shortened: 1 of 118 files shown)


bin/yupdate
```diff
@@ -224,7 +224,7 @@ module YUpdate
       system("mount", "--make-private", origdir)
 
       system("mount", "-t", "overlay", "overlay", "-o", "lowerdir=#{dir},"\
-        "upperdir=#{upperdir},workdir=#{workdir}", dir)
+                                                        "upperdir=#{upperdir},workdir=#{workdir}", dir)
```
dgdavid, line 227: Is this really needed? Looks weird.

src/lib/installation/widgets/online_repos.rb
```diff
@@ -1,3 +1,22 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2021 SUSE LLC
```
dgdavid, line 2: NP: It's 2022 already :) You can even use [2017-2022]

dgdavid (approved):
LGTM

bin/yupdate
jreidinger: well, it was caused by that string split in previous line. I think original code is strange. I will fix it

src/lib/installation/widgets/online_repos.rb
jreidinger: we should use century there, it is more stable :)

src/lib/installation/clients/inst_prepareprogress.rb
```diff
@@ -76,8 +77,8 @@ def main
           {
             "name"        => "images",
             "description" => _("Deploying Images..."),
-            "value"       => 300_000, # just make it longer than inst_finish, TODO: better value later
-            "units"       => :kb
+            "value"       => 400, # just make it longer than inst_finish
```
dgdavid, line 80: What a difference :)

dgdavid (approved):

PR #1019, opened 2022-02-10
Set the Xft.dpi resource to fix small fonts in hidpi screens

antlarr:
Qt 5.15 (actually, 5.14) introduces a behavior change with respect to hidpi screens and it now needs Xft.dpi to be set in order to use it. This change uses xrandr to find out edid information and xrdb to set Xft.dpi accordingly when starting the X server in YaST2.call. It always sets dpi to multiples of 48 with a minimum of 96 and a maximum of 288 (being effectively the same as having zoom factors of 1x, 1.5x, 2x, 2.5x amd 3x). Higher (or lower) dpi values are not used in order to try to filter out monitors that return wrong edid values.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Feb  9 15:05:32 UTC 2022 - Antonio Larrosa <alarrosa@suse.com>
+
+- Set the Xft.dpi resource after running the X server,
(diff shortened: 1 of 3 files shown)


antlarr:
BTW, this fixes https://bugzilla.opensuse.org/show_bug.cgi?id=1173451 and should go into SUSE:SLE-15-SP4:GA as well as TW

lslezak (commented):
Looks good :+1:
I'd just add some more logging.

PR #1020, opened 2022-02-18
SLE#25560: do not create a snapshot after install/upgrade if root filesystem is RO

imobachgs:
Feature [SLE#25560](https://jira.suse.com/browse/SLE-22560).
Trello: https://trello.com/c/RovP5dk3/

On systems using the transactional-update mechanism, the filesystem is supposed to be mounted as read-only at the end of the installation. In those cases, we should not create a snapshot.

In this attempt, we check the mount options in the storage staging devicegraph.

## Tests

- [x] SLE 15 SP3
- [x] MicroOS (Snapshot20220215)
- [x] SLE 15 SP3 with AutoYaST, as the snapshots_finish client is executed during the 2nd stage too.
- [x] SLE 15 SP3 upgrade
- [x] SLE 15 SP3 upgrade with AutoYaST

microos-snapshots

### DIFF
--- a/src/lib/installation/snapshots_finish.rb
+++ b/src/lib/installation/snapshots_finish.rb
@@ -2,6 +2,7 @@ require "yast"
 require "yast2/fs_snapshot"
 require "yast2/fs_snapshot_store"
 require "installation/finish_client"
+require "y2storage/storage_manager"
 
 module Installation
   class SnapshotsFinish < ::Installation::FinishClient
@@ -28,7 +29,7 @@ module Installation
     def write
       snapper_config
 
-      if InstFunctions.second_stage_required? || !Yast2::FsSnapshot.configured?
+      if InstFunctions.second_stage_required? || !Yast2::FsSnapshot.configured? || ro_root_fs?
         log.info("Skipping root filesystem snapshot creation")
         return false
       end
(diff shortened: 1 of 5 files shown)


src/lib/installation/snapshots_finish.rb
```diff
@@ -28,16 +29,16 @@ def initialize
...
-      if !InstFunctions.second_stage_required? && Yast2::FsSnapshot.configured?
-        log.info("Creating root filesystem snapshot")
-        if Mode.update
-          create_post_snapshot
-        else
-          create_single_snapshot
-        end
-      else
+      if InstFunctions.second_stage_required? || !Yast2::FsSnapshot.configured? || ro_root_fs?
         log.info("Skipping root filesystem snapshot creation")
```
imobachgs, line 33: Does it worth logging a different message for each condition?

src/lib/installation/snapshots_finish.rb
```diff
@@ -77,5 +78,17 @@ def snapper_config
         log.info("There is no need to configure Snapper")
       end
     end
+
+    # Determines whether the root filesystem is mounted as read-only
+    #
+    # @return [Boolean] true if it is mounted as read-only; false if it is not
+    #   mounted as read-only or if it is not found
+    def ro_root_fs?
```
ancorgs, line 86: This method implementation is perfectly ok. Unfortunately we don't have a way to check the condition that is significantly more straightforward in the Y2Storage API.

Although this may be slightly better:

```ruby
staging = Y2Storage::StorageManager.instance.staging
root = Y2Storage::MountPoint.find_by_path(staging, "/")
return false unless root

root.mount_options.include?("ro")
```

Recovering the `MountPoint` object directly with `find_by_path` is slightly more direct than looking up the filesystems. But there is no difference in practice. Both implementations should work exactly ok in all cases I can think.

imobachgs: Thanks. I will adapt the code.

ancorgs (approved):
LGTM

PR #1023, opened 2022-02-20
 SLE#25560: do not create a snapshot after install/upgrade if root filesystem is RO (master)

imobachgs:
Merge #1020 (and #1022) into `master`.


doc/snapshots.md
```diff
@@ -0,0 +1,78 @@
+# Installation and upgrade snapshots
+
+This document tries to clarify how YaST, as an installer, creates snapshots during installation and
+offline upgrade when Snapper is enabled. In a nutshell, YaST creates these snapshots:
+
+* A *single* snapshot at the end of the installation (except for transactional systems, like
+  MicroOS).
```
jreidinger, line 7: what about transactional server role? does it count as transactional system?

imobachgs: Just, it is mentioned in the section about transactional systems.

doc/snapshots.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+According to the table above, YaST created snapshot 5 (*pre*) before starting the upgrade process
+and snapshot 6 (*post*) at the end. However, snapshot 1 is still the root file system.
+
+## Transactional systems
+
+Transactional systems, like MicroOS or the *Transactional Server* role of openSUSE Tumbleweed, take
+care of their own snapshots. In those cases, YaST does not perform any snapshot at the end of the
+installation. About the offline upgrade, using YaST to upgrade those systems is not supported.
```
jreidinger, line 71: is this somewhere documented more officially? If so, I would add link to it here.

doc/snapshots.md
```diff
@@ -0,0 +1,78 @@
...
+
+Starting in yast2-installation 4.3.45, YaST detects if it is installing a transactional system by
+checking whether the root filesystem is mounted as read-only.
```
jreidinger, line 74: So if I get it basically now for transactional systems, yast does not create any snapshots, right?

doc/snapshots.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Related features
+
+[FATE#317932](https://w3.suse.de/~lpechacek/fate-archive/317973.html) and SLE-22560.
```
jreidinger, line 78: why not link also to jsc?

src/lib/installation/snapshots_finish.rb
```diff
@@ -29,16 +30,16 @@ def initialize
...
-        if Mode.update
-          create_post_snapshot
-        else
-          create_single_snapshot
-        end
-      else
+      if InstFunctions.second_stage_required? || !Yast2::FsSnapshot.configured? || ro_root_fs?
         log.info("Skipping root filesystem snapshot creation")
-        false
+        return false
```
jreidinger, line 35: do we see in logs reason why root fs snapshot is skipped? It can help to debug if someone complain that he does not have snapshot.

doc/snapshots.md
imobachgs: No, at least I do not know about any official documentation. I asked Jiri for confirmation in Jira, that's all.

doc/snapshots.md
imobachgs: Because it is an internal link.

jreidinger: yeap, but w3.suse.de is also internal. So either have both of it or none of it.

doc/snapshots.md
jreidinger: Not sure if customers will be happy about it. It should be somewhere documented for users.

doc/snapshots.md
imobachgs: Yes, that's it.

src/lib/installation/snapshots_finish.rb
imobachgs: OK, I have improved the PR to log the problem.

src/lib/installation/snapshots_finish.rb
```diff
@@ -78,5 +79,17 @@ def snapper_config
...
+    # Determines whether the root filesystem is mounted as read-only
+    #
+    # @return [Boolean] true if it is mounted as read-only; false if it is not
+    #   mounted as read-only or if it is not found
+    def ro_root_fs?
+      staging = Y2Storage::StorageManager.instance.staging
+      root_fs = Y2Storage::MountPoint.find_by_path(staging, "/").first
+      return false unless root_fs
+
+      root_fs.mount_options.include?("ro")
```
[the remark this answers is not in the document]
imobachgs: It looks like it is the only way by now.

doc/snapshots.md
imobachgs: Good point. I will link the JIRA entry too.

doc/snapshots.md
imobachgs: I would expect the documentation to contain some information about that. I will ask in the JIRA entry.

jreidinger (approved):

PR #1024, opened 2022-02-21
Fix Second Stage timeout when using ssh (bsc#1195059)

teclator:
## Problem

**AutoYaST init-scripts** are run before **systemd-user-sessions.service** which prevent users including privileged ones to login the system which in case of **ssh** installation does not allow the user to connect until pam_systemd gives a timeout.

- https://bugzilla.suse.com/show_bug.cgi?id=1195059

## Solution

Adapted the service file as suggested being run just after systemd-user-sessions.service.


joseivanlopez (approved):
LGTM

PR #1026, opened 2022-02-22
Drop -rfbwait parameter on Xvnc

jotolo2:
With the update to tigervnc 1.12.0.
Xvnc no longer supports -rfbwait parameter.
This should fix bsc#1196201.

With older Xvnc versions, -rfbwait default value is 30 seconds. 
That shouldn't be a problem.

### DIFF
--- a/startup/common/vnc.sh
+++ b/startup/common/vnc.sh
@@ -90,7 +90,6 @@ startVNCServer () {
 		-geometry "$VNCSize" \
 		-depth 16 \
                 -dpi 96 \
-		-rfbwait 120000 \
 		-rfbport 5901 \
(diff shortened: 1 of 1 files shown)


ancorgs:
Just for the record. Both @jotolo2 and @sndirsch confirmed this change is correct and safe to be also released as part of SLE-15-SP4. And @suntorytimed confirmed it would be acceptable for that release.

So I'm merging this right away and I will create a separate PR with the version bump needed to trigger the continuous integration that submits this to openSUSE Tumbleweed and SLE-15-SP4 (and thus Leap 15.4).

ancorgs (approved):
LGTM, as per previous comment.

PR #1017, opened 2022-02-01
Memsample: redirect the STDERR output (bsc#1195116)

lslezak:
### The Problem

screenshot

- Redirect the STDERR output in the memsample script to not break the YaST UI
- See https://bugzilla.suse.com/show_bug.cgi?id=1195116#c8
- 4.4.38

PR #1027, opened 2022-02-22
SSH installation theme fix

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1196287
- The `installation_slim` theme does not exist anymore, after merging https://github.com/yast/yast-yast2/pull/1242 it is not changed to the default theme. If theme does not exist then the default (light) Qt theme is used.
- Additionally fixed crash is the `memsample-archive-to-csv` script, related to https://github.com/yast/yast-installation/pull/1017

## Testing

Tested manually, with the fix it uses the usual default installation style:

ssh_installation_dark


shundhammer (approved):
LGTM

PR #1028, opened 2022-02-23
Remake some icons

hellcp:


ancorgs:
I guess this should only target Tumbleweed and not 15.4. Correct?

hellcp:
Yeah, keep in mind there is a parallel PR in packager as well

dgdavid:
> Yeah, keep in mind there is a parallel PR in packager as well

This one: https://github.com/yast/yast-packager/pull/603

jreidinger (approved):

PR #1033, opened 2022-03-08
Prevent getty auto-generation

joseivanlopez:
- [bsc#1196614](https://bugzilla.suse.com/show_bug.cgi?id=1196614)
- https://trello.com/c/yLz1DrJp

## Problem

Dependencies in *YaST2-Second-Stage.service* have been adjusted. First, it was necessary to force the second stage service to run after *systemd-user-sessions.service* in order to avoid issues when connecting with ssh, see https://github.com/yast/yast-installation/pull/1024. Then, that patch was improved to avoid display manager and getty prompt to appear too early, see https://github.com/yast/yast-installation/pull/1030.

With the latest changes, it is still possible to switch to another virtual console (ctrl+alt+f2-6), making *systemd-getty-generator* to instantiate a getty service on demand. Note that getty auto-generation is prevented if *systemd-user-sessions.service* runs after YaST second stage, as it was done before the lastest changes in the service dependencies.

And, for some reason, if a getty is instantiate on demand before *YaST2-Second-Stage.service* is started, then xvnc fails to start. Note that switching to a virtual console before second stage usually does not happen, but openQA tests could do it to execute something.

## Solution

*YaST2-Second-Stage.service* was already preventing *getty@tty1.service* to be instantiate before finishing the second stage. This was avoiding an early prompt login in tty1. Now, *getty@tty2-6.service* are also explicitly prevented, so *systemd-getty-generator* will not generate getty sessions on demand. For openQA tests in s390, the service *serial-getty@ttysclp0.service* needs to be prevented, too. 

## Testing

Manually tested with both ssh and vnc installations.

### DIFF
--- a/package/YaST2-Second-Stage.service
+++ b/package/YaST2-Second-Stage.service
@@ -1,9 +1,14 @@
 [Unit]
 Description=YaST2 Second Stage
-After=apparmor.service local-fs.target plymouth-start.service systemd-user-sessions.service
+After=apparmor.service local-fs.target plymouth-start.service
+# systemd-user-sessions is needed to allow ssh connection (bsc#1195059)
+After=systemd-user-sessions.service
 Conflicts=plymouth-start.service
-Before=getty@tty1.service serial-getty@ttyS0.service serial-getty@ttyS1.service serial-getty@ttyS2.service
-Before=serial-getty@hvc0.service serial-getty@ttyAMA0.service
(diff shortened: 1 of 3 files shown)


teclator (approved):
LGTM

teclator (changes requested):
The PR needs to be adapted according to openQA test which failed because it uses ttysclp0 virtual console.

See https://bugzilla.suse.com/show_bug.cgi?id=1196614#c24

PR #1036, opened 2022-03-17
Run YaST2-Second-Stage after purge-kernels service

teclator:
## Problem

If **purge-kernels** does not finish before **YaST2-Second-Stage** is run it locks **zypper** and a nasty message is shown. For preventing this situation YaST2-Second-Stage should be run after the **purge-kernels.service**.

- https://bugzilla.suse.com/show_bug.cgi?id=1196431 (https://trello.com/c/tl491wU2)

## Solution

Adapt **YaST2-Second-Stage.service** to be fun after **purge-kernels.service**

### DIFF
--- a/package/YaST2-Second-Stage.service
+++ b/package/YaST2-Second-Stage.service
@@ -3,6 +3,8 @@ Description=YaST2 Second Stage
 After=apparmor.service local-fs.target plymouth-start.service
 # systemd-user-sessions is needed to allow ssh connection (bsc#1195059)
 After=systemd-user-sessions.service
+# Run after kernels are purged to prevent a zypper block (bsc#1196431) 
+After=purge-kernels.service
 Conflicts=plymouth-start.service
 # Prevent getty auto-generation (bsc#1196614, bsc#1196594)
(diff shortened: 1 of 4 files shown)


package/YaST2-Second-Stage.service
```diff
@@ -3,6 +3,8 @@ Description=YaST2 Second Stage
 After=apparmor.service local-fs.target plymouth-start.service
 # systemd-user-sessions is needed to allow ssh connection (bsc#1195059)
 After=systemd-user-sessions.service
+# Run after kernels are purged to prevent a zypper block (bsc#1196431)
```
joseivanlopez, line 6: block or lock?

joseivanlopez (approved):
LGTM

joseivanlopez (changes requested):
BTW, shouldn't it be added to firstboot service too?

package/YaST2-Second-Stage.service
teclator: lock

teclator:
> BTW, shouldn't it be added to firstboot service too?

Yes, adapted accordingly

joseivanlopez (approved):

PR #1038, opened 2022-03-18
Add Before=xvnc.socket dependency to Firstboot and Second Stage services (bsc#1197265)

teclator:
## Problem

- During the Second Stage of an installation, the **xvnc.socket** can starts before the **YaST2-Second-Stage** service runs its own Xvnc server. That could produce a black screen vnc connection as reported by openQA.

- In the past (see [yast/yast-installation: Pull Request 316](https://github.com/yast/yast-installation/pull/316)) we used to have a **systemd dependency** to be run before **xinetd** service and we removed that because some cycle dependency but once we removed **xinetd** it looks safer to add a before **xvnc.socket** dependency instead of stop the service and run it again which looks strange.

- https://bugzilla.suse.com/show_bug.cgi?id=1197265 (https://trello.com/c/cl2XRuwf/)

## Solution

Add **Before=xvnc.socket** to **YaST2-Second-Stage** and **YaST2-Firstboot** services. This will also prevent agetty to get the 5901 port (related to https://bugzilla.suse.com/show_bug.cgi?id=1196614).

### DIFF
--- a/package/YaST2-Firstboot.service
+++ b/package/YaST2-Firstboot.service
@@ -6,7 +6,7 @@ After=purge-kernels.service
 Conflicts=plymouth-start.service
 Before=getty@tty1.service serial-getty@ttyS0.service serial-getty@ttyS1.service serial-getty@ttyS2.service
 Before=serial-getty@hvc0.service serial-getty@ttyAMA0.service
-Before=display-manager.service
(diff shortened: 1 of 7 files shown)


joseivanlopez (approved):
LGTM

PR #1039, opened 2022-03-18
Add Before=xvnc.socket dependency to Firstboot and Second Stage services (bsc#1197265) in SLE-15-SP4 / TW

teclator:
It merges #1038


joseivanlopez (approved):
LGTM

PR #1042, opened 2022-04-27
Revert "AY: Do not stop xvnc.socket but rely on systemd" in SLE-15-SP4

teclator:
It merges #1041 into SLE-15-SP4


dgdavid (approved):

dgdavid:
Still looking good, @teclator 

Thanks for making CI happy :)

PR #1045, opened 2022-05-24
Added Before=serial-getty@sclp_line0.service to YaST2-Second-Stage.service (SLE-15-SP4)

teclator:
It merges #1044 which contains the description and explanation


joseivanlopez (approved):
LGTM

PR #1046, opened 2022-05-24
Added Before=serial-getty@sclp_line0.service to YaST2-Second-Stage.service (master)

teclator:
It merges #1045 in master


joseivanlopez (approved):
LGTM

PR #1049, opened 2022-07-08
another try at adressing the hidpi issue (bsc#1199020)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1199020
- https://trello.com/c/ST2Xm5iz

The previous attempt https://github.com/yast/yast-installation/pull/1019 doesn't work and runs into issues described in  the above mentioned bug.

## Solution (to be discussed)

Here is a proposal received via email from Karsten Nordsiek.

The patch will need some polishing but essentially the decision is now based on the screen width instead of the dpi number.

### DIFF
--- a/startup/YaST2.call
+++ b/startup/YaST2.call
@@ -84,7 +84,7 @@ function calculate_x11_dpi () {
 	log "Monitor width mm: $MON_WIDTH_MM"
 	log "Monitor width px: $MON_WIDTH_PX"
 	log "Monitor dpi: $DPI"
-	echo "$DPI"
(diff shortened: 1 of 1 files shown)


shundhammer:
I am very sceptical about this approach. It's certainly not the sheer pixel width that matters; if that screen is very wide in reality (like those new curved screens), just assuming HiDPI because of that pixel width will certainly introduce new problems. That would mean just moving the problem around and making another group of users unhappy instead.

shundhammer:
See also https://github.com/yast/yast-installation/pull/1057

PR #1052, opened 2022-07-19
Do not use "xrdb" for setting the "Xft.dpi" value (bsc#1201532)

lslezak:
## Problem

- `yast2-installation` depends on `xrdb`, that has two negative consequences
  - `xrdb` depends on some X libraries and because `yast2-installation` is included in the installed system then the X libraries are installed even in a minimal text mod installation
  - it also depends on the C pre-processor which has ~22MB (!)

## Solution

- Use an YaST specific tool instead of `xrdb` (see https://github.com/yast/yast-x11/pull/25), it does not depend on `cpp`
- Install the `yast2-x11` package only when `libyui-qt` is installed, otherwise it does not make sense

## Notes

The "new" expressions in RPM dependencies are supported since rpm 4.13. This is used in SUSE since 2017, SLE15-GA contains rpm 4.14.1 so this is supported even in the first SLE15 release. See more details in https://rpm-software-management.github.io/rpm/manual/boolean_dependencies.html

## TODO

- [ ] Check that installation-images is still correctly built
- [ ] Check the real saved space in the YaST containers
- [ ] Check if we could backport the fix to SP4 and how much space we would save in textmode installation

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,16 @@
 -------------------------------------------------------------------
+Mon Jul 18 16:05:26 UTC 2022 - Ladislav Slezák <lslezak@suse.cz>
+
+- Do not use "xrdb" for setting the "Xft.dpi" value, use a specific
+  YaST tool from the yast2-x11 package (bsc#1201532)
+  (xrdb depends on the C pre-processor increasing the dependencies
+   about of 22MB)
+- Install yast2-x11 only when GUI (libyui-qt) is installed,
+  avoid installing the dependent X libraries in minimal (text mode)
+  installation
+- 4.5.4
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


lslezak:
Note to the SP4 backport:
- The affected YaST packages are not part of the minimal installation so they do not harm in this case
- If you select the text mode installation then the X server + libraries are included and so the dependencies do not harm, the X packages are installed anyway
- The only case where it could make troubles is minimal system + manually installed affected packages. But I think that is probably a very rare case, if you install the very minimal system you probably would not use YaST there...

The conclusion: The backport is not required, the problem should happen only in rare cases.

lslezak:
This saved about 13MB in the downloaded image size and about 30MB when the image is uncompressed on the disk.

lslezak:
Both SLE and openSUSE skelcd-control-* packages require `yast2-x11` so it is guaranteed that the package is included in the inst-sys.

PR #1053, opened 2022-07-22
Before user sessions

teclator:
## Problem

In order to prevent a ssh timeout produced because of systemd logind we [added](https://github.com/yast/yast-network/pull/1024) a condition to run the **SecondStage.service** after  **systemd-user-sessions** but once syste systemd logind has been patched (see https://github.com/systemd/systemd/pull/23936 & https://build.suse.de/request/show/276157) our fix can be reverted removing 
 the Before condition.

- https://bugzilla.suse.com/show_bug.cgi?id=1200780 & https://bugzilla.suse.com/show_bug.cgi?id=1195059

## Solution

Remove the **After=systemd-user-sessions** condition from the YaST2-SecondStage.service once systemd logind has been patched to prevent the ssh timeout.

### DIFF
--- a/package/YaST2-Second-Stage.service
+++ b/package/YaST2-Second-Stage.service
@@ -1,8 +1,6 @@
 [Unit]
 Description=YaST2 Second Stage
 After=apparmor.service local-fs.target plymouth-start.service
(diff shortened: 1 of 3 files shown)


joseivanlopez (approved):
LGTM

PR #1057, opened 2022-09-12
Directly Calculate the Qt Scale Factor, don't rely on the Xft.dpi X Resource

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1199020

## Trello

https://trello.com/c/ST2Xm5iz/

## Problem

Qt has a life of its own when it comes to determining the scaling of YaST's geometry and what font sizes to use during installation in HiDPI cases. Sometimes the fonts are too large, sometimes they are too small.

## Cause

While we used to check the DPI based on EDID and the `xrandr` output and set the `Xft.dpi` X resource accordingly (making sure to always use values that can be divided by 48, at least 96 dpi, at most 288 dpi), Qt rounded the resulting values to an integer number, based on a new "scale factor rounding policy".

We found that rounding and the resulting integer scaling factors are too coarse; factors like 1.5 or 2.5 work just fine (even 1.25, 1.75, 2.25, 2.75 would work just fine) as the major desktops prove who use those factors for magnifying GUIs all the time.

## Fix

Don't rely on setting the DPI to the `Xft.dp` X resource and hope for the best that Qt will do the right thing; we are now explicitly setting the scale factor with two environment variables:

- `QT_SCALE_FACTOR`
- `QT_SCALE_FACTOR_ROUNDING_POLICY="PassThrough"` (to avoid rounding)

Since we already make sure that the DPI value we use is rounded so it can always be divided by 48, and we limit to min 96 and max 288, we will always get one of those scale factors:

- 1
- 1.5
- 2
- 2.5
- 3

## To Discuss

In a previous PR, @antlarr introduced a maximum of 288 dpi to guard against bogus EDID values. But modern 13" laptops with a 4k display already have 336 dpi; should we use a higher maximum?

## Related Gist

https://gist.github.com/shundhammer/79d534b8452554b16fb2f667f58290f9

### DIFF
--- a/startup/YaST2.call
+++ b/startup/YaST2.call
@@ -75,13 +75,31 @@ function calculate_x11_dpi () {
 #------------------------------------------------------
 # Calculate the monitor's dpi from xrandr's output
 # ---
-	local MON_SIZE=`xrandr |grep mm |head -n 1`
-	local MON_WIDTH_MM=`echo $MON_SIZE | sed -e "s/.* \([0-9]\+\)mm x \([0-9]\+\)mm.*/\1/"`
+        local MON_SIZE=`xrandr |grep mm |head -n 1`
 	local MON_WIDTH_PX=`echo $MON_SIZE | sed -e "s/.* \([0-9]\+\)x[0-9]\++.*/\1/"`
-	local DPI=`ruby -e "puts (($MON_WIDTH_PX/($MON_WIDTH_MM*0.0393701)/48).round)*48"`
+	local MON_WIDTH_MM=`echo $MON_SIZE | sed -e "s/.* \([0-9]\+\)mm x \([0-9]\+\)mm.*/\1/"`
+
 	log "Monitor size: $MON_SIZE"
-	log "Monitor width mm: $MON_WIDTH_MM"
 	log "Monitor width px: $MON_WIDTH_PX"
+	log "Monitor width mm: $MON_WIDTH_MM"
+
+        if [ ${MON_WIDTH_MM:-0} -le 0 ]; then
+                log "Monitor width in mm unknown, using 96 dpi as default"
+                DPI=96
+        else
+                local DPI=`ruby -e "puts (($MON_WIDTH_PX/($MON_WIDTH_MM*0.0393701)/48).round)*48"`
+        fi
+
+        if [ ${DPI:-96} -gt 288 ]; then
+                log "Monitor dpi is suspiciously high (${DPI}), using 288 dpi"
+                DPI=288
+                # Not sure if capping this at 288 dpi is really appropriate;
+                # 13" 4k laptops easily get up to 336 dpi.
+                # The original intent was to protect against bogus EDID values.
+                #
+                # 2022-09-12 shundhammer
+        fi
+
 	log "Monitor dpi: $DPI"
 	echo "$DPI"
 }
@@ -92,6 +110,37 @@ function set_xft_dpi () {
 # Set Xft.dpi resource using a helper tool
 # ---
 	/usr/lib/YaST2/bin/xftdpi "$1" && log "Xft.dpi set to: $1"
+        # If the Xft.dpi X resource is set, it overrides the value
+        # that Qt otherwise calculates.
+}
+
+
+function set_qt_scale_factor () {
+#------------------------------------------------------
+# Set Qt environment variables for the scale factor, i.e. the factor by which
+# to increase the size of the UI in HiDPI cases.
+#
+# We assume the UI is designed for 96 dpi and calculate the scale factor based
+# on that. We use multiples of 0.5 here (1.0, 1.5, 2.0, 2.5, 3.0) which is
+# already guaranteed by enforcing a DPI value that is a multiple of 48
+# (see function calculate_x11_dpi()).
+# ---
+        local DPI=${1:-96}
+
+        if [ $DPI -le 96 ]; then
+                export QT_SCALE_FACTOR=1
+                # Also in non-HiDPI cases to avoid surprises if Qt
+                # should decide to do its own scaling
+        else
+                export QT_SCALE_FACTOR=`ruby -e "puts $DPI/96.0"`
+        fi
+
+        # Override the Qt default of rounding to the next integer.
+        # https://doc-snapshots.qt.io/qt5-5.15/highdpi.html
+        export QT_SCALE_FACTOR_ROUNDING_POLICY="PassThrough"
+
+        log "QT_SCALE_FACTOR: $QT_SCALE_FACTOR"
+        log "QT_SCALE_FACTOR_ROUNDING_POLICY: $QT_SCALE_FACTOR_ROUNDING_POLICY"
 }
 
 
@@ -115,17 +164,8 @@ function prepare_for_x11 () {
 			log "\tX-Server is ready: $xserver_pid"
 
 			local DPI=`calculate_x11_dpi`
-			if [ "$DPI" -ge 144 -a "$DPI" -le 288 ]; then
-				set_xft_dpi $DPI
-			elif [ "$DPI" -gt 288 ]; then
-				# This is a workaround against monitors
-				# that provide bad edid values
-				log "WARNING: Detected monitor DPI is too high ($DPI), using value 288"
-				set_xft_dpi 288
-			else
-				log "WARNING: Detected monitor DPI is too low ($DPI), using value 96"
-				set_xft_dpi 96
-			fi
+			# set_xft_dpi $DPI
+                        set_qt_scale_factor $DPI
 		fi
 	fi
 
@@ -506,6 +546,13 @@ function start_yast_again () {
 	fi
 }
 
+
+
+#----------------------------------------------------------------------
+# MAIN
+#----------------------------------------------------------------------
+
+
 #=============================================
 # Start the Magic :-)
 #=============================================
@@ -514,6 +561,7 @@ function start_yast_again () {
 . /usr/lib/YaST2/startup/common/functions.sh
 . /usr/lib/YaST2/startup/common/network.sh
 . /usr/lib/YaST2/startup/requires
+
 
 #=============================================
 # 1.1) set splash progress bar to 90%
--- a/startup/YaST2.call
+++ b/startup/YaST2.call
@@ -563,6 +563,18 @@ function start_yast_again () {
 . /usr/lib/YaST2/startup/requires
 
 
+# Debugging helper: Only calculate and show the DPI, then exit.
+if [ -n "$DEBUG_YAST_DPI" ]; then
+        echo "*** DPI Debug mode ***"
+        DPI=`calculate_x11_dpi`
+        echo "DPI: $DPI"
+        set_qt_scale_factor $DPI
+        env | grep "^QT_SCALE"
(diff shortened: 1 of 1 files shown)


shundhammer:
@antlarr, @Vogtinator : FYI

startup/YaST2.call
```diff
@@ -71,31 +70,80 @@ function wait_for_x11() {
...
+	local MON_WIDTH_MM=`echo $MON_SIZE | sed -e "s/.* \([0-9]\+\)mm x \([0-9]\+\)mm.*/\1/"`
+
 	log "Monitor size: $MON_SIZE"
-	log "Monitor width mm: $MON_WIDTH_MM"
 	log "Monitor width px: $MON_WIDTH_PX"
+	log "Monitor width mm: $MON_WIDTH_MM"
+
+        if [ ${MON_WIDTH_MM:-0} -le 0 ]; then
+                log "Monitor width in mm unknown, using 96 dpi as default"
+                DPI=96
```
dgdavid, line 88: NP: missing `local`?

shundhammer: Yes, good point.

shundhammer: Done.

startup/YaST2.call
```diff
@@ -563,6 +563,18 @@ function start_yast_again () {
 . /usr/lib/YaST2/startup/requires
 
 
+# Debugging helper: Only calculate and show the DPI, then exit.
+if [ -n "$DEBUG_YAST_DPI" ]; then
```
wfeldt: np: not `YAST_DEBUG_DPI`? Just trying to esatblish a consistent yast env vars naming scheme...

shundhammer: Good point. I'll change it.

shundhammer: Also updated the documentation here.

shundhammer:
@joseivanlopez tested this on real HiDPI hardware: A Dell XPS 13 9310 laptop with 4k on a 13" display:

y2start.log:

```
Stage [call]: Monitor size: eDP-1 connected primary 3840x2400+0+0 (normal left inverted right x axis y axis) 288mm x 180mm
Stage [call]: Monitor width px: 3840
Stage [call]: Monitor width mm: 288
Stage [call]: Monitor dpi: 336
Stage [call]: QT_SCALE_FACTOR: 3.5
Stage [call]: QT_SCALE_FACTOR_ROUNDING_POLICY: PassThrough
```

yast2-000

yast2-002

yast2-004

So the fonts and dimensions look alright; correctly scaled up, as expected.

He also wrote that the X11 mouse cursor was tiny, as well as window manager titles. Both is expected since we scale up our Qt application (which has that built-in scaling); the X server which loads a cursor font, or the window manager know nothing about that.

While the WM titles are not important at all during installation, IMHO we can ignore them; but a reasonably sized mouse cursor would be very desirable. But that will be a separate bug / feature / PBI.

wfeldt:
I've tried the patch on my 14" 1920x1080 laptop and the chosen scaling 1.75 is too big there. I would like to keep
a scaling of 1.

wfeldt:
My very personal note on this whole hidpi mess: I'd never scale anything. People buy these things for a reason.

But if we scale I'd just avoid very high dpi values. In my case the display has 158 dpi originally and with the chosen scaling
this comes down to 90 dpi effectively. That's way to coarse and reminds me of monitor resolutions from some 20+ years ago.

How about choosing the scaling so that the effective dpi is at least (random pick) 144 (= 1.5 * 96)?

This would result in a scaling of 1 for my laptop and result in the still acceptable scaling of 1 on my 4k monitor (I'm using 1.25 actually). :-)

shundhammer:
Summarizing a discussion that we just had in Slack:

Steffen would prefer much less drastic scaling.

I argued that one basic problem is that too many of our dialogs are much too cramped: Too much text and widgets, very much defeating the core idea behind YaST to make complicated things simpler.

Also, we always need to support typical VM resolutions like 800x600, and NCurses with 80x24 characters, so we need to design our dialogs with content that also fits into those sizes; i.e. with _much_ less stuff than the dialogs here (license and language selection, role selection, proposal (which is no longer a summary as the headline suggests, but an ever-growing list that scrolls off)).

Of course that means that it will either look tiny and lost in 4k unscaled, or too large if scaled. But a lot of users complained that they couldn't read the content anymore (if unscaled by accident), or that the scaling caused fonts that were much too large, cutting buttons off, or leaving almost no space for list and table contents.

This PR was meant to alleviate those problems, not as a complete redesign of our installation dialogs.

shundhammer:
As a compromise, I just changed the reference from 96 dpi to 144 dpi, so the scale factor is now smaller (i.e. the fonts are not quite as large). We might have to experiment with different values to see what works best.

shundhammer:
## Test on Real HiDPI Hardware 2 (with 144 dpi Reference)

@joseivanlopez tested this again after the change to 144 dpi as the reference on real HiDPI hardware: A Dell XPS 13 9310 laptop with 4k on a 13" display.

Relevant part of y2start.log:

```
Stage [call]: HiDPI Check...
Stage [call]: ==============
	|-- Monitor size: eDP-1 connected primary 3840x2400+0+0 (normal left inverted right x axis y axis) 288mm x 180mm
	|-- Monitor width px: 3840
	|-- Monitor width mm: 288
	|-- Monitor dpi: 336
	|-- Reference DPI: 144
	|-- QT_SCALE_FACTOR: 2.25
	|-- QT_SCALE_FACTOR_ROUNDING_POLICY: PassThrough
```

Complete y2start.log (click to open):
<details>
```
Stage [1]: YaST2 startup level (1)...
Stage [1]: ==========================
	|-- declare -x BOOT_IMAGE="/boot/x86_64/loader/linux"
	|-- declare -x Cmdline="splash=silent"
	|-- declare -x EFI="1"
	|-- declare -x EGL_LOG_LEVEL="fatal"
	|-- declare -x ESCDELAY="10"
	|-- declare -x HOME="/"
	|-- declare -x InitrdID="2022-09-12.1c17066d"
	|-- declare -x InitrdModules="zram loop scsi_dh_alua scsi_dh_emc scsi_dh_rdac rtsx_pci mei mmc_core rtsx_pci_sdmmc iscsi_boot_sysfs cdrom sr_mod st sg thermal"
	|-- declare -x Insecure="0"
	|-- declare -x InstsysID="2022-09-12.1c17066d"
	|-- declare -x Keyboard="1"
	|-- declare -x LD_LIBRARY_PATH=":/y2update/lib64:/y2update/lib:/y2update/plugin"
	|-- declare -x LIBGL_ALWAYS_INDIRECT="1"
	|-- declare -x Locale="en_US"
	|-- declare -x Manual="0"
	|-- declare -x MemFree="15755276"
	|-- declare -x OLDPWD
	|-- declare -x Options="thermal tzp=50"
	|-- declare -x PATH="/bin:/sbin:/usr/bin:/usr/sbin:/usr/X11/bin:/lbin"
	|-- declare -x PWD="/"
	|-- declare -x PYTHONPATH="/usr/lib/python3.10/site-packages:"
	|-- declare -x RepoURL="<RepoURL may contain sensitive information>"
	|-- declare -x SHLVL="2"
	|-- declare -x SSHD="0"
	|-- declare -x SelfUpdate="1"
	|-- declare -x SetHostname="1"
	|-- declare -x SetHostnameUsed="0"
	|-- declare -x Sourcemounted="0"
	|-- declare -x StartShell="1"
	|-- declare -x TERM="linux"
	|-- declare -x Textmode="0"
	|-- declare -x UpdateDir="/linux/suse/x86_64-tw"
	|-- declare -x UseSSH="0"
	|-- declare -x VNC="0"
	|-- declare -x WithFCoE="0"
	|-- declare -x WithiSCSI="0"
	|-- declare -x XCURSOR_THEME="DMZ"
	|-- declare -x Y2GDB="0"
	|-- declare -x YAST2_SSH="false"
	|-- declare -x YAST_DEBUG="/debug/yast.debug"
	|-- declare -x YaST2update="1"
	|-- declare -x ZyppRepoURL="<ZyppRepoURL may contain sensitive information>"
	|-- declare -x dud="disk:/y2call.dud"
	|-- declare -x kexec_reboot="1"
	|-- declare -x linuxrc_no_auto_assembly="1"
	|-- declare -x splash="silent"
	|-- declare -x startshell="1"
	|-- declare -x zram_swap="1G"
	|-- Creating hook script list: preFirstStage...
	|-- Level 1 stages:
	|-- F02-hwinfo
	|-- F03-umount
	|-- F04-language
	|-- F05-cmdline
	|-- F06-terminal
	|-- F08-logging
	|-- F09-start
	|-- F10-cleanup
Stage [1]: Starting F02-hwinfo...
Stage [1]: ======================
	|-- Hardware detection for arch: x86_64
Stage [1]: Starting F03-umount...
Stage [1]: ======================
	|-- Trying to umount inst-sys exits with code: 32
Stage [1]: Starting F04-language...
Stage [1]: ========================
	|-- Language set to: en_US.UTF-8
	|-- Starting UTF-8 mode...
Stage [1]: Starting F05-cmdline...
Stage [1]: =======================
Stage [1]: Starting F06-terminal...
Stage [1]: ========================
	|-- Setup virtual console:
	|-- Lines: 150
	|-- Columns: 480
	|-- Type: linux
Stage [1]: Starting F08-logging...
Stage [1]: =======================
	|-- Computed default log size: 3615782 kB
	|-- Set YaST2 LOG parameters:
	|-- Maximum log size: 20000 kB
	|-- Maximum log count: 5
	|-- Start sampling memory usage data:
	|-- PID: 4032
Stage [1]: Starting F09-start...
Stage [1]: =====================
	|-- Creating hook script list: preFirstCall...
	|-- Delegate program call to YaST2.call installation initial
Stage [call]: Starting prequalification checks...
Stage [call]: ===================================
	|-- X-Server is needed
	|-- Copying /etc/X11/xorg.conf.template to /etc/X11/xorg.conf
	|-- No SSH daemon running -> Medium SSH disabled
Stage [call]: Evaluate medium selection...
Stage [call]: ============================
Stage [call]: Check selected medium...
Stage [call]: ========================
	|-- Wished medium is: QT
	|-- Selected medium is: QT
	|-- X-Server is ready: 4495
Stage [call]: HiDPI Check...
Stage [call]: ==============
	|-- Monitor size: eDP-1 connected primary 3840x2400+0+0 (normal left inverted right x axis y axis) 288mm x 180mm
	|-- Monitor width px: 3840
	|-- Monitor width mm: 288
	|-- Monitor dpi: 336
	|-- Reference DPI: 144
	|-- QT_SCALE_FACTOR: 2.25
	|-- QT_SCALE_FACTOR_ROUNDING_POLICY: PassThrough
Stage [call]: ================
Stage [call]: Starting YaST...
Stage [call]: ================
	|-- Allow big memory allocation: overcommit_memory=1
	|-- MODULE_NAME: installation
	|-- MODE_FLAGS:
	|-- MODULE_ARGS: --arg initial
	|-- MODE: qt
	|-- UI_ARGS: --noborder --auto-fonts --fullscreen
	|-- QT_IM_MODULE: xim

```
</details>

yast2-000

yast2-001

yast2-002

yast2-003

wfeldt (approved):
It now goes for a scale factor of 1.25 for my FHD 14" laptop which is fine.

lgtm

PR #1058, opened 2022-09-15
Avoid Recreating all openQA Needles

shundhammer:
## Problem

The last PR https://github.com/yast/yast-installation/pull/1057 changed the font size even in non-HiDPI cases ever so slightly, so the existing openQA needles don't fit anymore.

190419184-e43e5ec2-c564-4817-bad5-4b44a152ece7

## Cause

The two environment variables were always set, even for a scale factor of 1:

```
QT_SCALE_FACTOR=1
QT_SCALE_FACTOR_ROUNDING_POLICY="PassThrough"
```

which obviously does make a difference.

## Fix

Only set them if > 1.


ancorgs (approved):
LGTM

PR #1059, opened 2022-09-20
Write ssg-apply config

joseivanlopez:
## Problem

Security policies and rules can be enabled/disabled during the (auto-)installation. That info has to be transfered to the config file of the script provided by the ssg-apply package.  

For more info, see:

* https://github.com/yast/yast-security/pull/128
* https://github.com/yast/yast-security/pull/137

## Solution

Adapt *security_finish* client to also write the security policies config.

## Testing

* Added new unit tests
* Tested manually

### DIFF
--- a/src/lib/installation/clients/security_finish.rb
+++ b/src/lib/installation/clients/security_finish.rb
@@ -1,5 +1,5 @@
 # ------------------------------------------------------------------------------
-# Copyright (c) 2017 SUSE LLC
+# Copyright (c) [2017-2022] SUSE LLC
 #
 #
 # This program is free software; you can redistribute it and/or modify it under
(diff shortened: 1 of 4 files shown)


src/lib/installation/clients/security_finish.rb
```diff
@@ -182,6 +186,24 @@ def write_firewall
         Yast::Service.Enable("sshd") if @settings.enable_sshd
         configure_firewall if @firewalld.installed?
       end
+
+      # Writes custom configuration for the ssg-apply script
+      def write_ssg_apply_config
+        # write security policies only during a fresh install
+        return if Yast::Mode.update
+
+        policies_manager = Y2Security::SecurityPolicies::Manager.instance
```
imobachgs, line 195: Should we consider placing this logic in the `Manager` class? Then we could reuse it in AutoYaST, for instance, if needed.

imobachgs (approved):

joseivanlopez:
Closed in favor of https://github.com/yast/yast-installation/pull/1061.

PR #1060, opened 2022-10-03
Fix generation of help text in the installation summary

ancorgs:
## Problem

In theory, the different proposal clients displayed in the installation summary have the option to incorporate their own help texts to the general help of the summary screen.

That's documented in the [`Installation::ProposalClient`](https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/installation/proposal_client.rb) class.

Unfortunately, the whole mechanism seems to be rather broken:

- The help texts of the clients are simply being ignored at least since the introduction of https://github.com/yast/yast-installation/pull/279 (at 2015!)
- The mentioned documentation at `ProposalClient` is also wrong. According to the current (broken) code, the help is fetched from the `description` hash, not from the `make_proposal` one (as documented).
- Nobody noticed it because there are exactly ZERO proposals with a help text at the moment of writing.

## Solution

Fix the broken code so the help texts from the proposal clients is added again to the general help text.

That will make again possible to specify per-client help texts, which we plan to use in the context of https://jira.suse.com/browse/PM-3526

## Testing

Tested manually

## Other pull requests

A pull request to `yast-yast2` should likely be created to fix the documentation of `ProposalClient`.

A pull request will be created in `yast-security` to introduce the desired help text in the corresponding proposal client.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,9 +1,15 @@
 -------------------------------------------------------------------
+Mon Oct  3 10:42:07 UTC 2022 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Fixed the help in the installation summary to include the texts
+  from the corresponding proposals (related to jsc#SLE-24764).
+- 4.4.57
+
+-------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ancorgs:
Changed the base branch to point to `security-policy` so everything is merged together.

PR #1061, opened 2022-10-04
Security policy

joseivanlopez:
## Problem

Security policies and rules can be enabled/disabled during the (auto-)installation. That info has to be transfered to the config file of the script provided by the ssg-apply package.  

For more info, see:

* https://github.com/yast/yast-security/pull/128
* https://github.com/yast/yast-security/pull/137

## Solution

Adapt *security_finish* client to also write the security policies config.

## Testing

* Added new unit tests
* Tested manually


ancorgs (approved):
LGTM

PR #1063, opened 2022-10-05
add 'repo', 'cd', and 'hd' schemes to Yast::Transfer::FileFromUrl

wfeldt:
## Task

- https://trello.com/c/CAytlL3W
- https://jira.suse.com/browse/SLE-22578
- https://jira.suse.com/browse/SLE-24584

Add URI schema that works relative to the installation medium. Also, unify the URI schemes so that the Zypp/Addon syntax also works with AutoYaST.

Note that `relurl` cannot be used as it already has the meaning of 'relative to the AutoYaST profile' in AutoYaST context.

So, a new `repo` scheme is introduced.

## See also

- https://github.com/openSUSE/linuxrc/pull/299
- https://github.com/yast/yast-yast2/pull/1276
- https://github.com/yast/yast-autoinstallation/pull/852

### DIFF
--- a/src/lib/transfer/file_from_url.rb
+++ b/src/lib/transfer/file_from_url.rb
@@ -145,6 +145,12 @@ module Yast::Transfer
         end
       end
 
+      # convert 'label' scheme to 'device' scheme
+      if toks["scheme"] == "label"
+        toks["scheme"] = "device"
+        toks["host"] = "disk/by-label/#{toks["host"]}"
+      end
+
       _Scheme = toks["scheme"]
       _Host = toks["host"]
       _Path = toks["path"]
--- a/src/lib/transfer/file_from_url.rb
+++ b/src/lib/transfer/file_from_url.rb
@@ -116,7 +116,7 @@ module Yast::Transfer
         _Localfile
       )
 
-      log.info "toks initial: #{toks.inspect}"
+      log.info "toks initial: #{hide_password(toks).inspect}"
 
       if _Scheme == "repo"
         base_url = InstURL.installInf2Url("")
@@ -125,17 +125,17 @@ module Yast::Transfer
           return false
         end
 
-        log.info("installation path from install.inf: #{base_url}")
+        log.info("installation path from install.inf: #{URL.HidePassword(base_url)}")
 
         toks["scheme"] = "relurl"
         rel_url = URL.Build(toks)
         log.info("relative url: #{rel_url}")
 
         absolute_url = Yast2::RelURL.from_installation_repository(rel_url).absolute_url.to_s
-        log.info("absolute url: #{absolute_url}")
+        log.info("absolute url: #{URL.HidePassword(absolute_url)}")
 
         toks = URL.Parse(absolute_url)
-        log.info "toks absolute: #{toks.inspect}"
+        log.info "toks absolute: #{hide_password(toks).inspect}"
       end
 
       # convert 'cd', "dvd', and 'hd' Zypp schemes to 'device' schema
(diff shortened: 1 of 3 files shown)


imobachgs (approved):

src/lib/transfer/file_from_url.rb
```diff
@@ -97,21 +101,78 @@ def get_file_from_url(scheme:, host:, urlpath:, localfile:,
...
+
+        absolute_url = Yast2::RelURL.from_installation_repository(rel_url).absolute_url.to_s
+        log.info("absolute url: #{absolute_url}")
+
+        toks = URL.Parse(absolute_url)
+        log.info "toks absolute: #{toks.inspect}"
+      end
+
+      # convert 'cd' and 'hd' Zypp schemes to 'device' schema
+      if ["cd", "hd"].include?(toks["scheme"])
```
lslezak, line 142: Um, why we should change the `cd:/` scheme? It is a standard scheme fully supported by libzypp, see "Supported URI formats" section in [man zypper](https://en.opensuse.org/SDB:Zypper_manual_(plain)).

BTW libzypp also supports alternative `dvd:/` scheme...

lslezak (commented):
Changing the CD/DVD scheme?

src/lib/transfer/file_from_url.rb
wfeldt: Yes, but `get_file_from_url` does not support it.

src/lib/transfer/file_from_url.rb
```diff
@@ -97,21 +101,78 @@ def get_file_from_url(scheme:, host:, urlpath:, localfile:,
...
+
       Builtins.y2milestone(
         "Scheme:%1 Host:%2 Path:%3 Localfile:%4",
         _Scheme,
         _Host,
         _Path,
         _Localfile
       )
+
+      log.info "toks initial: #{toks.inspect}"
```
lslezak, line 119: Can the URL possibly contain a sensitive data? Like a password? In that case we should not log all URL parts, the password should be omitted...

src/lib/transfer/file_from_url.rb
```diff
@@ -97,21 +101,78 @@ def get_file_from_url(scheme:, host:, urlpath:, localfile:,
...
+
+      log.info "toks final: #{toks.inspect}"
+
+      # URL.Build does not reconstruct the URL in all cases; notably it has
+      # some ideas about what the host part might look like - which conflicts
+      # with the host part being used for device names in local disk URIs.
+      #
+      # It does not matter much as full_url is only used for ftp/http(s).
       full_url = URL.Build(toks)
+      log.info("full url (host part might be missing): #{full_url}")
```
lslezak, line 175: For logging URLs we should always use the `URL.HidePassword` method.

```suggestion
      log.info("full url (host part might be missing): #{URL.HidePassword(full_url)}")
```

src/lib/transfer/file_from_url.rb
```diff
@@ -97,21 +101,78 @@ def get_file_from_url(scheme:, host:, urlpath:, localfile:,
...
+      _Path = toks["path"]
+
       if Builtins.regexpsub(_Path, "(.*)//(.*)", "\\1/\\2") != nil
         _Path = Builtins.regexpsub(_Path, "(.*)//(.*)", "\\1/\\2")
+        log.info "path changed from #{toks["path"]} to #{_Path}"
       end
-      Ops.set(toks, "path", _Path)
+      toks["path"] = _Path
+
+      log.info "toks final: #{toks.inspect}"
```
lslezak, line 167: Similar problem here...

lslezak (commented):
A password from URL can possibly leak into the y2log... :thinking:

PR #1064, opened 2022-10-21
Fix entropy poolsize when copying with dd

teclator:
## Problem

**YaST** copies the entropy pool during installation. But the pool size is actually provided in bits, not bytes as dd bs parameter expect.

- [bsc#1204559](https://bugzilla.suse.com/show_bug.cgi?id=1204559)

## Solution

Convert the read pool size from bits to bytes in order to be passed to dd


wfeldt (approved):
lgtm

PR #1065, opened 2022-10-21
Fix entropy poolsize when copying with dd (master)

teclator:
Merge #1064 into master


teclator:
Already merged by (https://github.com/yast/yast-installation/pull/1069)

PR #1070, opened 2022-11-25
adjust plymouth and First, Second stage start order.

qiangzhao:
adjust Plymouth and Yast-First-Stage and Yast-Second-stage start order.

### DIFF
--- a/package/YaST2-Firstboot.service
+++ b/package/YaST2-Firstboot.service
@@ -1,9 +1,8 @@
 [Unit]
 Description=YaST2 Firstboot
-After=apparmor.service local-fs.target plymouth-start.service YaST2-Second-Stage.service
+After=apparmor.service local-fs.target plymouth-quit-wait.service YaST2-Second-Stage.service
 # Run after kernels are purged to prevent a zypper lock (bsc#1196431)
 After=purge-kernels.service
-Conflicts=plymouth-start.service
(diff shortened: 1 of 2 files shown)


mvidner (changes requested):
Thanks! But WHY?

- Please summarize the reason in a comment/commit message
- Add a bug/feature reference

qiangzhao:
> Please summarize the reason in a comment/commit message
plymouth-quit-wait is the solution inside plymouth source, the plymouth-quit-wait.service happens on the time plymouth totally exit. It is designed as a *signal* to notice other program which wait for using display device could launch.
From the old method, "Yast2-Firstboot.service" and "Yast2-Second-Stage.service" only call "plymoth quit", it will notice plymouth daemon to quit, but it will not wait for plymouth daemon to finish the quit action, in some scenario, the release device action of plymouth daemon will take a little more time for kernel/DRM to return. In this case, if "Yast2-Firstboot.service" and "Yast2-Second-Stage.service" launch up, they may race with plymouth daemon to grasp the display device handle. and may lock each other. 

> Add a bug/feature reference
This is a problem that I found in my own research on Plymouth's source and systemd service files inside the Tumbleweed.

qiangzhao:
Hi @mvidner :
the reason as the above comment, Please review.

kobliha:
For the reference: https://bugzilla.suse.com/show_bug.cgi?id=1215282#c21

PR #1071, opened 2022-12-05
Use default Depth for Xvnc

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1205585

## Problem

On installations using Xvnc, sometimes the font rendering is broken in the Qt UI to the point of making some fonts illegible.

wizard-welcome

_See the wizard steps on the left side._

## Underlying Cause

@Vogtinator found out that there is a bug in the Qt libs:

https://bugreports.qt.io/browse/QTBUG-109169

> It's a bad conversion between QRgb and QRgba64 types in a gamma correct blending function. 

## Workaround

@Vogtinator also suggested to drop using the `-depth 16` option when starting `Xvnc` to avoid getting into that special case. That's what this PR does.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Dec  5 10:13:28 UTC 2022 - Stefan Hundhammer <shundhammer@suse.com>
+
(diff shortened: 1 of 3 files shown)


joseivanlopez (approved):
LGTM

imobachgs (approved):

PR #1078, opened 2023-01-25
Connect only nbft

teclator:
## Problem

It was requested to use `nvme connect-nbft` when **UseNBFT** was set by **linuxrc** but we call  `nvme connect-all`.

- https://trello.com/c/eFJXEeep/3274-3-urgent-top-priority-connect-to-nbft-disks
- https://jira.suse.com/browse/PED-3138
- https://build.suse.de/request/show/288814

## Solution 

Implement the call as requested (it depends on https://build.suse.de/request/show/288814)

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Jan 25 19:56:12 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
+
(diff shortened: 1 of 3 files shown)


mwilck (approved):
This is correct. Currently`nvme connect-all` does not imply `connect-nbft`.  (This may change in the future).  `connect-all` would only consider locally configured NVMeoF connections (regular or discovery subsystems), and because no such thing exists at this early stage, would basically be a no-op.

`connect-nbft` is correct.

imobachgs (approved):

PR #1085, opened 2023-05-11
Use a larger Font for xterm during Installation via X Resources

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1211267

## Problem

For debugging, you can start an _xterm_ during a YaST Qt installation with the `Ctrl`-`Shift`-`Alt` key combination. But since the X11 environment in the inst-sys is minimalistic, and _xterm_ is a very old (almost 40 years) X application, it falls back to built-in very small pixel fonts.

That makes that _xterm_ really hard to use even on normal-sized modern displays; on HiDPI displays, it's too tiny to be usable.

## Fix

This adds an old-fashioned `~/.Xdefaults` file to set the font size for _xterm_ via X resources:

```
! Fonts for xterm during a YaST Qt UI installation
! (You can start an xterm with Ctrl-Shift-Alt X)
XTerm*faceName:  DejaVu Sans Mono
XTerm*faceSize:  12
```

_Notice that the comment character in X resource files is '!', not '#' as in most file types._

Even though _xterm_ is one of the oldest X11 applications ever, it received support for modern scalable fonts; see `man xterm` for the resource names.

### DIFF
--- a/startup/YaST2.call
+++ b/startup/YaST2.call
@@ -179,6 +179,50 @@ function set_qt_scale_factor () {
 }
 
 
+function calculate_xterm_font_size() {
+#------------------------------------------------------
+# Calculate an appropriate font size for xterm based on the DPI
+# ($1) and set XTERM_FONT_SIZE accordingly.
+# ---
+
+	local DPI=${1:-96}
+
+	if [ ${DPI:-0} -le 96 ]; then
+	    XTERM_FONT_SIZE=10
+	elif [ $DPI -le 192 ]; then
+	    XTERM_FONT_SIZE=12
+	else
+	    XTERM_FONT_SIZE=14
+	fi
+
+	export XTERM_FONT_SIZE
+	log "\tXTERM_FONT_SIZE: $XTERM_FONT_SIZE for $DPI dpi"
+}
+
+
+function add_x11_resources() {
+#------------------------------------------------------
+# Add some X resources for xterm to /root/.Xdefaults
+#
+# Notice that this file is only used if nothing has been
+# loaded into the X server with 'xrdb' yet.
+# ---
+
+	local RESOURCE_FILE=/root/.Xdefaults
+	XTERM_FONT_SIZE=${XTERM_FONT_SIZE:-12}
+	log "\tAdding X resources for xterm to $RESOURCE_FILE"
+
+	cat >>$RESOURCE_FILE <<EOF
+
+! Fonts for xterm during a YaST Qt UI installation
+! (You can start an xterm with Ctrl-Shift-Alt X)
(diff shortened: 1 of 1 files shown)


hramrach:
I was wondering why you invent such baroque scheme when you can take a base value like 12 and multiply it by the Qt scaling factor to get consistent results.

But it looks like you are doing this in shell code and cannot easily multiply by a fractional number.

shundhammer:
> I was wondering why you invent such baroque scheme when you can take a base value like 12 and multiply it by the Qt scaling factor to get consistent results.

I am very sceptical that the Qt scaling factor would be a good multiplier here: The _xterm_ font size doesn't quite fit even with smaller resolutions today, so I really want to start at 10, use 12 for standard resolutions and something larger (right now 14) for HiDPI or resolutions approaching HiDPI.

We also have a lower and an upper limit for the Qt scaling factor.

There is still room for optimization for those font sizes., of course.

> But it looks like you are doing this in shell code and cannot easily multiply by a fractional number.

We already shell out to Ruby to calculate that scaling factor; that would not really be the problem.

hramrach:
There you go https://github.com/yast/yast-installation/pull/1086/files

startup/YaST2.call
```diff
@@ -179,6 +179,50 @@ function set_qt_scale_factor () {
...
+
+	if [ ${DPI:-0} -le 96 ]; then
+	    XTERM_FONT_SIZE=10
+	elif [ $DPI -le 192 ]; then
+	    XTERM_FONT_SIZE=12
+	else
+	    XTERM_FONT_SIZE=14
+	fi
+
+	export XTERM_FONT_SIZE
```
lslezak, line 198: It would be nice to keep the current `XTERM_FONT_SIZE` if it is already defined.

Linuxrc propagates all unknown boot options as environment variables. Then you could use e.g. `XTERM_FONT_SIZE=12` boot option to force your own value if the default calculated value is not correct.

Or you could set it manually when using `startshell=1` boot option before starting the installer.

lslezak (commented):
Allow overriding the calculated `XTERM_FONT_SIZE` value

startup/YaST2.call
shundhammer: Added.

lslezak (approved):
:+1:

lslezak:
JFYI: simplified testing in inst-sys:

- Boot latest Tumbleweed installation with `startshell=1 netsetup=dhcp`
- At the shell prompt run `yupdate patch yast-installation huha-xterm-font` (now use `master` as the last branch parameter, the branch has been deleted)
- Then run `XTERM_FONT_SIZE=12 yast` and test it in YaST

PR #1089, opened 2023-06-13
Don't always Enable SSHD and Open SSH Port [SLE-15-SP4]

shundhammer:
# This was superseded by #1090.

### DIFF
--- a/src/lib/installation/security_settings.rb
+++ b/src/lib/installation/security_settings.rb
@@ -144,7 +144,7 @@ module Installation
     #   authentication and the system is not accesible through ssh
     def access_problem?
       # public key is not the only way
-      return false unless only_public_key_auth
+      return false unless only_public_key_auth?
 
       # without running sshd it is useless
       return true unless @enable_sshd
@@ -181,11 +181,11 @@ module Installation
     end
 
     def wanted_enable_sshd?
-      Yast::Linuxrc.usessh || only_public_key_auth || @enable_sshd
+      Yast::Linuxrc.usessh || only_public_key_auth? || @enable_sshd
     end
 
     def wanted_open_ssh?
-      Yast::Linuxrc.usessh || only_public_key_auth || @open_ssh
+      Yast::Linuxrc.usessh || only_public_key_auth? || @open_ssh
     end
 
     def wanted_open_vnc?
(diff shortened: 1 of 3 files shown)


src/lib/installation/security_settings.rb
```diff
@@ -67,6 +67,30 @@ def load_features
       load_feature(:polkit_default_privs, :polkit_default_privileges)
     end
 
+    # Return whether the sshd service will be enabled.
+    #
+    # @return [Boolean]
+    def enable_sshd
```
teclator, line 73: To be honest I don't like this approach and it is a getter that modify the state of the object which is not expected . For enabling the sshd there is already a bang method. I would maintain the getters as them are.

src/lib/installation/security_settings.rb
```diff
@@ -181,27 +205,32 @@ def global_section
     end
 
     def wanted_enable_sshd?
-      Yast::Linuxrc.usessh || only_public_key_auth || @enable_sshd
```
teclator, line 184: This is the main fix, we should not evaluate if we only have public key access during the initialization but when the proposal is done, therefore, we could do it as part of the proposal and for that we probably will need to extend the state to know if it was already proposed or not.

teclator:
Until now, the **SecuritySettings** instance was created at the proposal summary step or by **AutoYaST** when it was needed and after the users configuration was already done. But since we [added](https://github.com/yast/yast-security/pull/128) security policies support (STIG) this is different and it is evaluated too early mainly because it also create and instance of **SecuritySettings** during the initialization instead of doing it using a lazy load init.

We already have a **propose_lsm_config** as part of the initialization, maybe we should move that proposal and the ssh config proposal to a **propose method** which will be called when it is really needed storing that the propose is already done just to avoid resetting it.

teclator:
The other options is fix the initialization of the SecurityPolicies::TargetConfig moving the @security accessor to a method 

 https://github.com/yast/yast-security/pull/128/files#diff-b6b853b94bce3cd38105d93a59732aa0345ed6ad7e53df5afad8e337d13bdb7cR54

src/lib/installation/security_settings.rb
shundhammer: Then we'd have to rely on somebody from the outside to do the right thing. That is a lot less reliable.

PR #1090, opened 2023-06-15
Don't always Enable SSHD and Open SSH Port (V2.0) [SLE-15-SP4]

shundhammer:
## Target Branch

**This is for _SLE-15-SP4_.** A merge to _SLE-15-SP5_ and _master_ / _Factory_ will follow.

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1211764

## Trello

https://trello.com/c/ofwfoQfm/

## Problem

In the security proposal just before committing the installation, SSHD was always enabled, and the SSH port 22 was always opened by default, no matter if a root password was set or not.

If there is no root password, that makes sense since then it might be a case where only public key authentication was possible after the installation. But if a root password was set, it should not be the default to enable the SSHD service and open the SSH port.

## Cause

This uses a class `Installation::SecuritySettings` which is meant to be used as a singleton. And as the singleton instance is used for the first time, all of its values were initialized, and those values were FINAL.

That included the values for opening the ssh port and the firewall. And as a fallback, if there was no root user yet, or the root password was empty, it assumed that there was only public key authentication, and in that case, it opened the SSH port and enabled SSHD.

The trouble was that all this happened BEFORE the user was prompted for the root password, so at that point, of course the root password was still empty, so it always fell back to public key authentication.

## Fix

This factors out the check if only public key authentication to a new separate method `SecuritySettings.propose` and calls that method when the security proposal is made in the general proposal dialog ("Installation Settings") during the installation.

## Test

TBD

## Related PRs

- _SLE-15-SP5_: TBD
- _master_ / _Factory_: TBD


src/lib/installation/security_settings.rb
```diff
@@ -78,22 +78,49 @@ def propose_lsm_config
...
+
+      log.info("Only public key auth")
+      open_ssh! unless @open_ssh
+      enable_sshd! unless @enable_sshd
+    end
+
+    # Reset the proposal; i.e. the next call to 'propose' will do a fresh
+    # proposal.
+    def reset_proposal
+      @proposal_done = false
```
teclator, line 105: NP: Similar to the comment above, it could lead to a confusion if someone also expect the proposal to make some other security config proposal. In this case only the ssh config will be marked as not proposed in order to propose it again.

teclator (approved):
LGTM

src/lib/installation/security_settings.rb
shundhammer: If we choose to extend this, we'll have to think about that; should there be one flag for all of them, or do we need several? It will depend on many things.

But again, it might also never happen.

PR #1093, opened 2023-07-04
ignore cpu_mitigations for bootloaders which do not support cpu mitig…

schubi2:
…ations e.g. none/systemd-boot

## Problem

systemd-boot bootloader currently does not support cpu mitigation. So no proposal should be shown for this 
setting like "none" bootloader is already doing it. So this is more a generic solution for it.

- https://github.com/yast/yast-bootloader/pull/686
- https://jira.suse.com/browse/PED-1906

## Solution

Checking if the corresponding bootloader has a cpu_mitigations call.

## Testing
- Tested manually

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Tue Sep  5 09:41:29 UTC 2023 - Josef Reidinger <jreidinger@suse.com>
 
-- Adapt code for changes in bootlaoder done for systemd-boot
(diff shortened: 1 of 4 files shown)


jreidinger:
change looks good, just please bump version and add changelog entry. Thanks

jreidinger:
OK, one more change is required at https://github.com/yast/yast-installation/blob/master/src/lib/installation/dialogs/security.rb#L124 which use wrong name

mvidner (changes requested):

mvidner (approved):

PR #1094, opened 2023-07-11
Fixed yupdate script (bsc#1213197)

lslezak:
## Problem

- Patching the Agama installer does not work properly
- It fails because the pre/post scripts are not executed

## Solution

- It turned out that the GitHub tarballs have all content in a `<repo>-<branch>` subdirectory, e.g. `agama-master`
- When downloading from GitHub add this extra path so the files are correctly found

## Testing

- Tested manually

### DIFF
--- a/bin/yupdate
+++ b/bin/yupdate
@@ -751,13 +751,13 @@ module YUpdate
       g.install_required_gems
     end
 
-    def run_pre_script(download_dir, subdir)
-      script = File.join(download_dir, subdir, ".yupdate.pre")
+    def run_pre_script(download_dir)
+      script = File.join(download_dir, ".yupdate.pre")
       run_script(script)
     end
 
-    def run_post_script(download_dir, subdir)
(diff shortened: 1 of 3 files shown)


bin/yupdate
```diff
@@ -766,12 +768,15 @@ module YUpdate
...
+    # @param downloader the downloader to use
+    # @param subdir subdirectory in the source tarball
+    def install_tar(downloader, subdir = "")
       Dir.mktmpdir do |download_dir|
         downloader.extract_to(download_dir)
-        run_pre_script(download_dir)
+        run_pre_script(download_dir, subdir)
         install_sources(download_dir)
-        run_post_script(download_dir)
+        run_post_script(download_dir, subdir)
```
mvidner, line 779: Nitpicking... the API would be nicer if we kept the signature and did the `File.join` here
but it's not worth changing and retesting I guess

mvidner (approved):

mvidner (approved):
Simpler now, yay thanks :)

PR #1095, opened 2023-08-04
Status toggling in the "Previously Used Repositories" dialog didn't work well

mchf:
## Problem

From the bug description:
What happened prior to Leap 15.3 was the Toggle Status button cycled through three options: Delete, Keep, and Disable. For Leap 15.4 and 15.5, The displayed options are only Delete and Keep. However, there are still three states, as repeatedly pressing Toggle Status will show you. First press of Toggle Status changes the display to Keep; second press the display stays Keep; third press the display returns to Delete.

- [bsc1213959](https://bugzilla.suse.com/show_bug.cgi?id=1213959)
- Most probably introduced in this commit 661f8f5d6e450aff5727f4b4c3b5cac9dd874624

## Solution

Fixed symbols used for referencing to particular labels

Opensuse 15.4 and 15.5 affected. However impact is small so I propose it for tumbleweed only

## Testing

- *Tested manually*

## Screenshots

### DIFF
--- a/src/lib/installation/clients/inst_upgrade_urls.rb
+++ b/src/lib/installation/clients/inst_upgrade_urls.rb
@@ -150,10 +150,10 @@ module Yast
       when :removed
         # TRANSLATORS: The action to perform with a repository
         _("Remove")
-      when :disable
+      when :disabled
         # TRANSLATORS: The action to perform with a repository
         _("Disable")
-      when :enable
+      when :enabled
         # TRANSLATORS: The action to perform with a repository
         _("Enable")
       else
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Aug  4 07:09:13 UTC 2023 - Michal Filka <mfilka@suse.com>
+
+- bsc#1213959
+  - Change status label properly when toggling status in the
+    "Previousy used repositories" dialog.
+- 4.6.6 
+
+-------------------------------------------------------------------
 Tue Jul 11 07:20:29 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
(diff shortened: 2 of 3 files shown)


src/lib/installation/clients/inst_upgrade_urls.rb
```diff
@@ -150,10 +150,10 @@ def repo_action(status)
...
         _("Remove")
-      when :disable
+      when :disabled
         # TRANSLATORS: The action to perform with a repository
         _("Disable")
-      when :enable
+      when :enabled
         # TRANSLATORS: The action to perform with a repository
         _("Enable")
       else
```
lslezak, line 159: To avoid this problem in the future I'd add also a `when` section for the `keep` state and raise an exception in the `else` branch to report unhandled cases.

```ruby
when :enabled
  # TRANSLATORS: The action to perform with a repository
  _("Enable")
when :keep # or whatever symbol that state actually uses...
  # TRANSLATORS: The action to perform with a repository
  _("Keep")
else
  raise "Unhandled repository status #{status.inspect}"
end
```

lslezak (commented):
Explicitly handle all known cases, raise exception for unhandled case.

src/lib/installation/clients/inst_upgrade_urls.rb
mchf: after further investigation this is not so easily doable: https://github.com/yast/yast-installation/blob/master/src/lib/installation/upgrade_repo_manager.rb#L101

lslezak: It turned out that we cannot use this as the "keep" state is implicit if not set to any `:enabled`, `:disabled` or `:removed`... :worried:

lslezak (approved):

PR #1096, opened 2023-08-15
Require awk for use in Startup Scripts

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1214277

## Trello

https://trello.com/c/bgfX60e4/

## Problem

If `awk` is not installed, there are error messages on the console during a firstboot workflow.

## Fix

Require `awk` in the spec file.

Explicitly using `awk` and not `gawk` (the GNU version of `awk`) so lighterweight alternatives like _busybox_ can be used if they provide the `awk` symbol (which _busybox_ currently does not).

## What about other Common Linux Tools?

Many are part of the _coreutils_ package which we already require. Others are so common that a Linux system has little chance of working when they are missing, such as `grep` or `sed`: A lot of other packages depend on them.

See also the discussion in the bug.

### DIFF
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-installation
-Version:        4.6.6
+Version:        4.6.7
 Release:        0
 Summary:        YaST2 - Installation Parts
 License:        GPL-2.0-only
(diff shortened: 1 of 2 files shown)


package/yast2-installation.spec
```diff
@@ -58,6 +58,8 @@ Requires:       coreutils
 Requires:       gzip
 # use in startup scripts
 Requires:       initviocons
+# bsc#1214277; require awk, not gawk, to allow for lighterweight alternatives like busybox
+Requires:       awk
```
mvidner, line 62: I thought that `obs-service-format_spec_file` wanted the deps sorted alphabetically but it does not seem to be so... OK

mvidner (approved):

PR #1100, opened 2023-11-01
[master] Refresh repositories with changed URL and reload them again

lslezak:
- Just merging PR #1098  to master
- There is a small improvement, it uses the new `Repository.refresh` method
- See https://github.com/yast/yast-yast2/pull/1296


mvidner (approved):
LGTM

The CI failure looks triggered only by the new refresh method which I dare say is correct

lslezak:
Restarting the CI after merging the other PR fixed the problems.

PR #1101, opened 2023-11-29
bsc#1217637: retrieve an AutoYaST profile on a NFS+IPv6 setup

imobachgs:
## Problem

[bsc#1217637](https://bugzilla.suse.com/show_bug.cgi?id=1217637)

Retrieving an AutoYaST profile using an NFS share with an IPv6 address is impossible.

## Solution

Enclose the IPv6 address within square brackets when calling the `mount` command.

## Test

* Unit-tested only.
* Manual test is pending.


dgdavid (approved):
LGTM

PR #1104, opened 2024-01-24
Write cio_ignore kernel arguments always

teclator:
## Problem

When installing an s390x machine it is not writing the cio_ignore kernel command line parameter which is given through the parmfile.

- https://trello.com/c/LF2woKmG/3484-2-alp-p3-1210525-alpagamas390x-cioignore-parameters-from-the-installation-are-not-transferred-to-the-installed-system

## Solution

In **zVM** machines it does not write the **cio_settings** at all because it was requested **(fate#317861)**, but as the bug already reports, if the cio_ignore parameter is given through the **parmfile** we should kept it as it was given and if not we should call `cio_ignore -k` to obtain the kernel command line arguments to be used.

It depends in a change in the kernel because it was using the /etc/install.inf file for checking if there is some cmdline argument which is wrong in agama because there is no linuxrc at all.

## Testing

- *Tested manually*


src/lib/installation/cio_ignore.rb
```diff
@@ -56,6 +56,7 @@ def cio_setting
         # cio_ignore does not make sense for KVM or z/VM (fate#317861)
         # but for other cases return true as requested FATE#315586
         !(kvm? || zvm?)
+        true
```
jreidinger, line 59: before merge we should clean above lines

src/lib/installation/cio_ignore.rb
```diff
@@ -255,15 +240,22 @@ def write_autoconf_setting
...
+
+      return unless param == :missing
+
+      res = Yast::WFM.Execute(YAST_LOCAL_BASH_PATH, "/sbin/cio_ignore -k")
+
+      raise "cio_ignore -k failed with #{res["stderr"]}" if res["exit"] != 0
+
       # boot code is already proposed and will be written in next step, so just modify
-      Yast::Bootloader.modify_kernel_params("cio_ignore" => "all,!ipldev,!condev")
+      Yast::Bootloader.modify_kernel_params("cio_ignore" => res["stdout"].lines.first)
```
jreidinger, line 252: please strip lines here. I am not sure how bootloader handle newline in string

```suggestion
      Yast::Bootloader.modify_kernel_params("cio_ignore" => res["stdout"].lines.first.strip)
```

src/lib/installation/cio_ignore.rb
```diff
@@ -275,6 +267,7 @@ def store_active_devices
 
       # make sure the file ends with a new line character
       devices << "" unless devices.empty?
+      log.info "active devices to be written: #{devices.join(",")}"
```
jreidinger, line 270: I think it won't be big list, but it is good practice to call it just once. So `devices_txt = devices.join(",")` and use it later.

teclator: the join is different, but ok, we could reuse it, and realized there could be a bug in case of empty

teclator: The join is not exactly the same in both cases but I think is ok reusing it with the new line character instead of comma separated one.

PR #1105, opened 2024-01-26
In zVM || KVM write the cio_kernel_parameter only if given

teclator:
Change #1104 implementation writing the **cio_ignore kernel parameter** only if given according to the bug request. It depends on https://github.com/yast/yast-yast2/pull/1300

### DIFF
--- a/src/lib/installation/cio_ignore.rb
+++ b/src/lib/installation/cio_ignore.rb
@@ -24,16 +24,6 @@ module Installation
 
   private
 
-    def kvm?
-      File.exist?("/proc/sysinfo") &&
-        File.readlines("/proc/sysinfo").grep(/Control Program: KVM\/Linux/).any?
-    end
-
-    def zvm?
-      File.exist?("/proc/sysinfo") &&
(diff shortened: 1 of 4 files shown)


src/lib/installation/cio_ignore.rb
```diff
@@ -24,6 +24,16 @@ def reset
...
 
+    def kvm?
+      File.exist?("/proc/sysinfo") &&
+        File.readlines("/proc/sysinfo").grep(/Control Program: KVM\/Linux/).any?
+    end
+
+    def zvm?
+      File.exist?("/proc/sysinfo") &&
+        File.readlines("/proc/sysinfo").grep(/Control Program: z\/VM/).any?
+    end
```
jreidinger, line 35: Instead of adding back this method I suggest to use shared functionality from Yast::Arch - https://github.com/yast/yast-yast2/blob/d6ae8689e2ea9abf9908c1c9bd76c3dd557f78cf/library/general/src/modules/Arch.rb#L440

jreidinger (approved):

PR #1112, opened 2024-04-03
Fix call for connecting all discovered NVMe-over-Fabrics subsystems (bsc#1222246)

teclator:
## Problem

When **linuxrc** (see [openSUSE/linuxrc#310](https://github.com/openSUSE/linuxrc/pull/310 "‌")) detects that there is some NBFT configured interface it will sets the **UseNBFT** variable and YaST calls `nvme connect-nbft` in order to discover and connect to all discovered **NVMe-over-Fabrics subsystems** as was [requested](https://github.com/yast/yast-installation/pull/1078 "‌") at implementation time.

As the nvme cli has been adapted supporting NBFT table the call should be also adapted calling `nvme connect-all --nbft` as requested in the bug and also as we can see in the [implementation PR description](https://github.com/linux-nvme/nvme-cli/pull/1791).

- https://bugzilla.suse.com/show_bug.cgi?id=1222246

## Solution

Adapted the call as requested.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Apr  3 13:39:37 UTC 2024 - Knut Anderssen <kanderssen@suse.com>
+
+- Adapted call for connecting all discovered NVMe-over-Fabrics 
+  subsystems (bsc#1222246).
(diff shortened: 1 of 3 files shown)


shundhammer (approved):
LGTM

PR #1113, opened 2024-04-03
Fix call for connecting all discovered NVMe-over-Fabrics subsystems (bsc#1222246) (master)

teclator:
It merges #1112 into master


shundhammer (approved):
LGTM

PR #1114, opened 2024-04-15
WIP: Handle autoinst AND autoupgrade (bsc#1222153)

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1222153

## Problem

In AutoYaST upgrade mode, the `halt` instruction doesn't work:

```XML
<general>
  <mode>
    <halt config:type="boolean">false</halt>
  </mode>
  ...
</general>
```

## Cause

The code only checked for `Mode.autoinst`, not also for `Mode.autoupgrade`, or, more generally, for `Mode.auto` which checks for both.

## Fix

Check for `Mode.auto`, not just for `Mode.autoinst`.

## Test

This is intended as a DUD for the customer to test.

### DIFF
--- a/src/lib/installation/clients/yast_inf_finish.rb
+++ b/src/lib/installation/clients/yast_inf_finish.rb
@@ -84,7 +84,7 @@ module Yast
         Ops.set(@linuxrc, "Root", "kexec") if LoadKexec()
 
         # Override linuxrc settings in autoinst mode
-        if Mode.autoinst
+        if Mode.auto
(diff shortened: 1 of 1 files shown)


shundhammer:
No idea why coveralls thinks the coverage decreased. It's just a one-word change.

jreidinger:
BTW for SP7 and master it is handled in #1129 and #1132 where openqa require slight modification in behavior.

PR #1115, opened 2024-04-17
Added encryption_method and encryption_pbkdf in the product description file.

schubi2:


ancorgs (changes requested):
If we take https://github.com/yast/yast-storage-ng/pull/1383 as reference (I think that's what we really plan to merge into master), the descriptions here are not accurate.

schubi2:
I have changed it....

schubi2:
> If we take [yast/yast-storage-ng#1383](https://github.com/yast/yast-storage-ng/pull/1383) as reference (I think that's what we really plan to merge into master), the descriptions here are not accurate.

I have changed it. Thanks !

ancorgs (approved):
LGTM

PR #1117, opened 2024-05-14
Drop fbiterm in Tumbleweed Installation

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1224053

## Trello

https://trello.com/c/haqsYySu

## Problem

We are using _fbiterm_ for much improved support for nontrivial locales like Chinese, Japanese, Korean in NCurses. But _fbiterm_ is part of the _xiterm_ package (it's the framebuffer version of _xiterm_, hence **FB**-**I**term) which has been abandoned upstream many years ago. With the upcoming GCC 14, it has build problems. The maintainer of the _xiterm_ package feels unable to continue supporting it; so we need to drop the _xiterm_ package and with it _fbiterm_.

## Fix

- ~Don't require _xiterm_ in the YaST packages anymore.~
  It turns out we never required _xiterm_ or _fbiterm_ in any YaST package: We always checked at runtime if `/usr/bin/fbiterm` was available and used it only if it was.

- Don't use `fbiterm` in the YaST startup scripts anymore.
  Strictly speaking, we could simply rely on the check if it was available, but this PR cleans up a lot of legacy that goes back a long time, and that obscured the code in the startup scripts.

## Caveat

Installing in certain languages won't work anymore in NCurses (text mode). Those languages will use a fallback to English from now on; but only in NCurses. The Qt (graphical) installation will work as before.

## Target Distributions

For now, only Tumbleweed / Factory / YaST master is affected. SLE-15 SPx and SLE-12 SPx will retain _fbiterm_.

## Test

Manual test in a modified inst-sys with the new scripts

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue May 14 13:55:18 UTC 2024 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Drop fbiterm (bsc#1224053)
+- 5.0.11
+
+-------------------------------------------------------------------
 Wed May  8 07:22:28 UTC 2024 - Stefan Schubert <schubi@suse.com>
 
 - Added encryption_method and encryption_pbkdf in the product
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-installation
-Version:        5.0.10
+Version:        5.0.11
 Release:        0
 Summary:        YaST2 - Installation Parts
 License:        GPL-2.0-only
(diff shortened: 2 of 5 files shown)


shundhammer:
See also https://github.com/openSUSE/installation-images/pull/708

joseivanlopez (approved):
LGTM

shundhammer:
@lslezak : This is the example for CI via GitHub failing because the image doesn't have `/bin/diff`.

shundhammer:
Submitted manually with `rake osc:sr` to OBS openSUSE:Factory:

  https://build.opensuse.org/request/show/1177300

lslezak:
I have added the `diffutils` to the container image. Looks like it was removed from the base container, recently I had to add it elsewhere too...

PR #1118, opened 2024-05-30
Do report an error on `yupdate overlay`.

mvidner:
Before: it would silently do nothing
 (and I would wonder why my intent, `yupdate overlay create`, did not work)
Now: report an error, as intended


mvidner:
Found when testing https://github.com/yast/yast-registration/pull/600

lslezak (approved):
LGTM, thanks for fixing this!

PR #1119, opened 2024-08-27
Don't Block in AutoYaST Upgrade with Reboot Message [SLE-15-SP5]

shundhammer:
## Target Branch

**This is for SLE-15-SP5**. A merge to SLE-15-SP6, -SP7 and master / Factory will follow.

## Bugzilla

## Trello

## Problem

## Cause

## Fix

## Related PRs

- Merge to SLE-15-SP6: _TBD_
- Merge to SLE-15-SP7: _TBD_
- Merge to master / Factory: _TBD_

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Aug 27 09:06:22 UTC 2024 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Don't block in AutoYaST upgrade (bsc#1181625)
+- 4.5.20
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


package/yast2-installation.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Aug 27 09:06:22 UTC 2024 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Don't block in AutoYaST installation or upgrade (bsc#1181625)
```
jreidinger, line 4: this looks misleading, as autoinst already skip it in autoyast installation. So the fix is only for auto upgrade, is not?

shundhammer: changed.

jreidinger (approved):

package/yast2-installation.changes
shundhammer: Changed.

PR #1120, opened 2024-08-27
Don't Block in AutoYaST Upgrade with Reboot Message [SLE-15-SP6]

shundhammer:
## Target Branch

**This merges #1119 to SLE-15-SP6.**  A merge to SLE-15-SP7 and master / Factory will follow.

## Bugzilla

## Trello

## Problem

## Cause

## Fix

## Related PRs

- Original PR for SLE-15-SP5: #1119 
- Merge to SLE-15-SP7: _TBD_
- Merge to master / Factory: _TBD_

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Aug 27 09:06:22 UTC 2024 - Stefan Hundhammer <shundhammer@suse.com>
+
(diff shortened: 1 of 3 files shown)


jreidinger (approved):

shundhammer:
Submitted manually to IBS:

  https://build.suse.de/request/show/342273

Internal Jenkins still doesn't have the new rubygem-yast-rake with the SP6:Update target.

PR #1122, opened 2024-09-02
Fixing the self_update documentation

kobliha:
Reported by L3, ptoptions need to use the "+" sign to "add" something as described in https://en.opensuse.org/SDB:Linuxrc#p_ptoptions

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Mon Sep  2 07:55:45 UTC 2024 - locilka@suse.com
+
+- Fixing ptoptions usage in self_update documentation
+
+-------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


package/yast2-installation.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Mon Sep  2 07:55:45 UTC 2024 - locilka@suse.com
+
+- Fixing ptoptions usage in self_update documentation
```
imobachgs, line 4: Do we have a bug number?

kobliha: No, it's reported at Slack. IMO not needed as this is a well-known thing. See https://en.opensuse.org/SDB:Linuxrc#p_ptoptions

mvidner (changes requested):
Please do mention https://en.opensuse.org/SDB:Linuxrc#p_ptoptions in the commit message

imobachgs (approved):

mvidner (approved):

PR #1123, opened 2024-09-09
Don't Block in AutoYaST Upgrade with Reboot Message [master]

shundhammer:
## Target Branch

**This is the _master_ / _Factory_ version of #1119 / #1120.** It was originally meant to be a merge of the SLE-15-SP7 / SP6 version, but this commit somehow got lost in the merging process; so this is a cherry-pick of that commit.

## Bugzilla

## Trello

## Problem

## Cause

## Fix

## Related PRs

- Original PR for SLE-15-SP5: #1119 
- Merge to SLE-15-SP6: #1120
- Merge to SLE-15-SP7: Not needed since SP7 was branched off SP6 after this was merged.

### DIFF
--- a/package/yast2-installation.changes
+++ b/package/yast2-installation.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Sep  9 12:33:25 UTC 2024 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Don't block in AutoYaST upgrade (bsc#1181625)
+- 5.0.12
+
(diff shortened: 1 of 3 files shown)


joseivanlopez (approved):
LGTM

shundhammer:
See also https://bugzilla.suse.com/show_bug.cgi?id=1231522 which was a follow-up bug complaining about this fix.

shundhammer:
And now also https://bugzilla.suse.com/show_bug.cgi?id=1236434

PR #1129, opened 2024-12-17
Show dialog explicit

jreidinger:
## Problem

openqa needs a window to collect logs before reboot even in auto modes

## Solution

Respect explicit reboot timeout on command line even in automatic modes.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

### DIFF
--- a/src/lib/installation/clients/inst_finish.rb
+++ b/src/lib/installation/clients/inst_finish.rb
@@ -137,7 +137,6 @@ module Yast
 
     def report_messages
       return if Misc.boot_msg.empty?
-      return if Mode.auto
 
       # --------------------------------------------------------------
       # Check if there is a message left to display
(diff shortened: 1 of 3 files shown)


src/lib/installation/clients/inst_finish.rb
```diff
@@ -148,6 +147,9 @@ def report_messages
       if Linuxrc.reboot_timeout
         Report.DisplayMessages(true, Linuxrc.reboot_timeout)
       else
+        # Skip in autoinstallation and autoupgrade mode only if not explicitelly
+        # set on command line (bsc#1231522)
+        return if Mode.auto
```
teclator, line 152:
```suggestion
        return if Mode.auto
        
```

teclator (approved):
LGTM

src/lib/installation/clients/inst_finish.rb
```diff
@@ -148,6 +147,10 @@ def report_messages
       if Linuxrc.reboot_timeout
         Report.DisplayMessages(true, Linuxrc.reboot_timeout)
       else
+        # Skip in autoinstallation and autoupgrade mode only if not explicitelly
```
teclator, line 150:
```suggestion
        # Skip in autoinstallation and autoupgrade mode only if not explicitly
```

jreidinger: yeah, my spell checker catched it in changes, but here it is not used.

PR #1133, opened 2025-05-15
Added encryption_authentication in the control-file description

schubi2:
Preselect authentication (TPM, FIDO2, password) for encrypted devices.

### DIFF
--- a/package/yast2-installation.spec
+++ b/package/yast2-installation.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-installation
-Version:        5.0.16
+Version:        5.0.17
 Release:        0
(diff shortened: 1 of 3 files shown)


package/yast2-installation.spec
```diff
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-installation
-Version:        5.0.16
+Version:        5.0.17
```
joseivanlopez, line 19: Do we need a new version because this change in a documentation file?

schubi2: I do not really know the rules when the version number has to be increased. I have thought that it has to be increased for every change in master. Shall I decrease it again ?

PR #1139, opened 2025-10-28
Use a Longer ZYPP Timeout in Second Stage [master]

shundhammer:
## Merge PR

This merges #1138 from SLE-15-SP7 to _master_.


ancorgs (approved):
LGTM

